PR #589, opened 2021-02-03
default.scss: Move answers.scss after collapsible filters

creotutar:
We move the HH-configurable answers.scss after collapsible filters so
that any changes in answers.scss to collapsible filters classes take
precedence (without the need for !important).

J=None
TEST=manual

Make sure that a style change to (example below) works.

```
  .CollapsibleFilters .Answers-resultsHeader {
    background-color: red;
  }
```


creotutar:
Closing, this was done as part of #613

PR #590, opened 2021-02-04
Add custom command patches

cea2aj:
Add patches which update sites with older theme versions to have the same commands as v1.20

These patches allow developers to use CBD with older theme versions. These patches are not necessary if a site is on theme v1.20.

This PR contains 3 patches:
- A patch for 1.15 and earlier
- A patch for 1.16
- A patch for themes 1.17 through 1.19

After applying the patch, the files must be committed, and then the CBD session need to be restarted

To use these patches, click on the branch name of this PR, open a desired patch file from inside the patches folder, click 'Raw', and copy the URL. In SGS CBD, supply that URL for the diff to apply.

J=SLAP-1013
TEST=manual

Open a site in SGS CBD on theme 1.15 and test these patches using the 'Apply Diff' functionality. Also test on a 1.16 site and on a 1.19 site.


oshi97 (approved):

cea2aj:
The updated versions of these patches are here: https://github.com/yext/answers-hitchhiker-theme/pull/691

PR #607, opened 2021-02-12
Update CFilters helpers

cea2aj:
Update the cfilters helpers to use the new core interface

Since core.persistentStorage and core.globalStorage no longer exist in the same spot in the SDK, we cannot include calls to them in the Theme code. Use core.storage instead.

Also enable facets on the test site's location page for testing.

J=SLAP-1097
TEST=manual

Use the cfilters configured on the locations page to test the 'clear search' button, the 'sorts and filter' page, and the 'reset filters' button


cea2aj:
Closing this because these changes got put into #616

PR #623, opened 2021-02-25
interactive-map: NewMapConfig class to hold config logic

creotutar:
We add a NewMapConfig class to handle the logic of configuration for the
NewMap component. This helps us move the responsibility of configuration
parsing to its own class and cut down the bulk in InteractiveMap. This
is also a good way for Hitchhikers to shadow only this configuration and
not the entire component, for common use cases.

Additionally, we move some (previously) hardcoded values for clustering
to the new config to further let HHs change only this new file for configuration
changes.

Note that some configuration is used by both the InteractiveMap and the
NewMap and thus they exist in both component configuration. Also note
that the InteractveMap must (currently) accept specific passthrough
configuration for the NewMap, as it exists as the parent component for
it. (If you recall, we must pass in specific functions to NewMap, thus
must be the parent).

See the current component documentation (below).

```
InteractiveMap
{
    // required, The container that all functionality should exist inside, not re-rendered
    container: no default,

    // required, The current Answers API vertical key
    verticalKey: no default,

    // required, The vertical configuration
    verticalPages: no default,

    // optional, Map options to be passed directly to the Map Provider
    providerOptions: {}

    // optional, The default center coordinate for the map, an object with {lat, lng}
    providerOptions.center: { lat: 37.0902, lng: -95.7129 }

    // optional, The default zoom level for the map
    providerOptions.zoom: 14

    // optional, Whether the map should search on a map movement action like map drag
    disableSearchOnMapMove: false

    // optional, Contains no results configuration
    noResults: {},

    // optional, Whether the map should display all results on no results
    noResults.displayAllResults: false,

    // The following are accepted as a means of passthrough to NewMap
    mapProvider,
    apiKey,
    clientId,
    locale,
    pin,
    pinCluster,
    enablePinClustering,
    onPinSelect
}
```

```
NewMap
{
    // required
    container: no default,

    // required, The provider for the map, normalized to lowercase
    mapProvider: no default,

    // either apiKey or clientId required, The API key for the map provider (if applicable)
    apiKey: no default,

    // either apiKey or clientId required, The client id for the map provider (if applicable)
    clientId: no default,

    // required, has involved logic for defaulting based on validity of given locale
    locale: no default

    // required, The content wrapper for the floating content above the map
    contentWrapperEl: no default

    // optional, Map options to be passed directly to the Map Provider
    providerOptions: {}

    // optional, The default center coordinate for the map, an object with {lat, lng}
    defaultCenter: { lat: 37.0902, lng: -95.7129 }

    // optional, The default zoom level for the map
    defaultZoom: 14

    // optional, The mobile breakpoint (inclusive max) in px
    mobileBreakpointMax: 991

    // optional, The pin options for the map, with information for each pin state (e.g. default, hovered)
    pinOptions:  the default pin options

    // optional, The cluster pin options for the map, with information for each pin state
    pinClusterOptions:  the default pin cluster options

    // optional, Whether the map should cluster pins that are close to each other
    enablePinClustering: false

    // optional, Contains no results configuration
    noResults: {},

    // optional, Whether the map should display all results on no results
    noResults.displayAllResults: false,

    // optional, Callback for when a non-cluster pin is selected
    onPinSelect: function () {},

    // optional, Callback for when the map is rendered
    onPostMapRender: function () {},

    // optional, Callback for when a non-cluster pin is clicked
    pinClickListener: function () {},

    // optional, Callback for when a cluster pin is clicked
    pinClusterClickListener: function () {},

    // optional, Callback for when a map drag event has finished
    dragEndListener: function () {},

    // optional, Callback for when a map zoom event has fired
    zoomChangedListener: function () {},

    // optional, Callback for when a map zoom event has finished
    zoomEndListener: function () {},
}
```

J=SLAP-1117
TEST=manual

Smoke test all configuration to see that changes made to the JSON
reflect on the map experience. For example:

* Changing the map pin colors changes the pins in the map
* Enabling/disabling clustering works
* Default (landing state) map center can be changed
* Provider options are passed through
* Zooming still calls search this area
* Padding on the map is correct (pins on search are not under the
content wrapper)
* No results hides the pins

Also test that changing the new configuration options for clustering
affect the clustering. e.g. minClusterSize = 3 for the query `all` should
not show the cluster of size 2.

### DIFF
--- a/static/js/interactive-map/NewMapConfig.js
+++ b/static/js/interactive-map/NewMapConfig.js
@@ -1,0 +1,187 @@
+import { Coordinate } from './Geo/Coordinate.js';
+import { PinImages } from './PinImages.js';
+import { ClusterPinImages } from './ClusterPinImages.js';
+import { getLanguageForProvider } from './Util/helpers.js';
+
+/**
+ * The configuration for the NewMap component.
+ */
+export default class NewMapConfig {
+  /**
+   * @param {Object} rawConfig Configuration to parse
+   */
+  constructor (rawConfig) {
+    /**
+     * The provider for the map, normalized to lowercase
+     * @type {string}
+     */
+    this.mapProvider = rawConfig.mapProvider && rawConfig.mapProvider.toLowerCase();
+
+    /**
+     * The API key for the map provider (if applicable)
+     * @type {string}
+     */
+    this.apiKey = rawConfig.apiKey;
+
+    /**
+     * The client id for the map provider (if applicable)
+     * @type {string}
+     */
+    this.clientId = rawConfig.clientId;
+
+    /**
+     * The language locale for the map. This is different from
+     * the specified locale because some map providers do not support
+     * certain locales and we want to fallback in a specific pattern
+     * when we come across a locale we do not support
+     * @type {string}
+     */
+    this.language = getLanguageForProvider(rawConfig.locale, this.mapProvider);
+
+    /**
+     * The content wrapper for the floating content above the map
+     * @type {HTMLElement}
+     */
+    this.contentWrapperEl = rawConfig.contentWrapperEl;
+
+    /**
+     * Map options to be passed directly to the Map Provider
+     * @type {Object}
+     */
+    this.providerOptions = rawConfig.providerOptions || {};
+
+    const defaultCenter = rawConfig.defaultCenter 
+      || this.providerOptions.center 
+      || { lat: 37.0902, lng: -95.7129 };
+
+    /**
+     * The default center coordinate for the map, an object with {lat, lng}
+     * @type {Coordinate}
+     */
+    this.defaultCenter = new Coordinate(defaultCenter);
+
+    /**
+     * The default zoom level for the map
+     * @type {number}
+     */
+    this.defaultZoom = rawConfig.defaultZoom 
+      || this.providerOptions.zoom 
+      || 14;
+
+    /**
+     * The mobile breakpoint (inclusive max) in px
+     * @type {Number}
+     */
+    this.mobileBreakpointMax = rawConfig.mobileBreakpointMax || 991;
+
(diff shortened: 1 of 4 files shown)


static/js/interactive-map/NewMapConfig.js
```diff
@@ -0,0 +1,215 @@
+import { Coordinate } from './Geo/Coordinate.js';
+import { PinImages } from './PinImages.js';
+import { ClusterPinImages } from './ClusterPinImages.js';
+import { getLanguageForProvider } from './Util/helpers.js';
+import { defaultCenterCoordinate } from './constants.js';
+
+/**
+ * The configuration for the NewMap component.
+ */
+export default class NewMapConfig {
```
tmeyer2115, line 10: I think this change will be very helpful. We may want to change this class name as part of the "Change names of components" item. We want the HHs to understand that the config they specify in the JSON will be passed to this class, then this class might do some additional work to it.

static/js/interactive-map/NewMapConfig.js
```diff
@@ -0,0 +1,215 @@
...
+  /**
+   * @param {Object} rawConfig Configuration to parse
+   */
+  constructor (rawConfig) {
```
tmeyer2115, line 14: Could we call this `jsonConfig`? I think that would help someone who shadows this file better understand what's going on.

static/js/interactive-map/NewMapConfig.js
creotutar: Agreed! Added a reminder to SLAP-1105

static/js/interactive-map/NewMapConfig.js
creotutar: Good point, changed

tmeyer2115 (approved):

PR #625, opened 2021-02-26
WCAG Navigation

cea2aj:
Update the locator so that it is aria navigable with a screen reader or a keyboard

- Add focus listeners to any location card <a> element to trigger the pinFocus. I renamed pinClicked and pinClickListener to pinFocus and pinFocusListener because pinFocus better describes both a click and focus state.
- Make the scroll to pin feature work when tabbing through the pins
- Move the stickyBottom div so that cards are never hidden behind it
- Add focus states to the `searchThisAreaToggle` and the update location button.
- Add aria landmarks to the Map, Main location search section, and the Map controls section
- Move up the searchThisArea div so that you can tab into it right before going onto the map

Note: the `searchThisAreaToggle` focus doesn't work in IE11 because focus-within is not supported in IE11. Getting this to work in IE11 would likely require a re-work of the search this area checkbox, so this maybe makes more sense to tackle in QA

J=SLAP-1059
TEST=manual

In google chrome and safari, tab through the cards and observe the card being highlighted and observe the corresponding pin show focus. This doesn't work in firefox because firefox doesn't make <a> elements tabable by default. Since this is a larger issue with the theme, I did not address that in this PR. The solution likely involves adding tabindex="0" to those elements

Test that tabbing through the map pins causes the scroll to pin behavior.

Tab over the `searchThisAreaToggle` and the the update location button and view the focus state

Use the screen reader and the firefox landmarks extension and see the landmarks on the page. The role="region" is newer so sometimes it is only visible to newer accessibility tools.


cards/financial-professional-location/component.js
```diff
@@ -6,7 +6,7 @@ class financial_professional_locationCardComponent extends BaseCard['financial-p
   }
 
   onMount() {
-    {{> static/js/interactive-map/location-card/add-card-click-listener}}
+    {{> static/js/interactive-map/location-card/add-card-listeners}}
```
oshi97, line 9: could we rename this slightly? When I hear add-card I think of the add card command.

static/js/interactive-map/location-card/add-card-listeners.js
```diff
@@ -0,0 +1,41 @@
+const getIndexOfThisCard = () => {
```
oshi97, line 1: I think we can have this file export a single function that registers listeners, and have it be accessed through HitchhikerJS instead of a partial? Otherwise this code is added to the bundle multiple times, once per card. 
In general mixing hbs and js is a headache (I know this was how it was done before this pr but this seems like the time to change it)

cea2aj: Sure, I can definitely see how mixing JS and HBS can make things more confusing

cards/financial-professional-location/component.js
cea2aj: outdated due to the refactor

static/js/interactive-map/location-card/add-card-listeners.js
cea2aj: Added this to the locator bundle rather than HitchhikerJS

templates/vertical-interactive-map/page.html.hbs
```diff
@@ -17,10 +17,10 @@
     {{> templates/vertical-interactive-map/script/locationbias modifier="main" }}
     {{> templates/vertical-interactive-map/script/locationbias modifier="mobileMap" }}
   {{/script/core }}
-  <script src="{{relativePath}}/locator-components.js" defer></script>
+  <script src="{{relativePath}}/locator-bundle.js" defer></script>
     <div class="Answers AnswersVerticalMap CollapsibleFilters InteractiveMap InteractiveMap--listShown js-answersInteractiveMap">
       <div class="Answers-content">
-        <div class="Answers-contentWrap js-locator-contentWrap">
+        <div class="Answers-contentWrap js-locator-contentWrap"  role="region" aria-label="Main location search">
```
creotutar, line 23: does this need to be multilang?

cea2aj: ah, good catch

cea2aj: I wrapped these in calls to the translate helper. Once the strings are approved in this PR I can send them to Tom to get translated

cards/financial-professional-location/component.js
```diff
@@ -6,7 +6,7 @@ class financial_professional_locationCardComponent extends BaseCard['financial-p
   }
 
   onMount() {
-    {{> static/js/interactive-map/location-card/add-card-click-listener}}
+    window.InteractiveMap && new CardListenerAssigner({card: this}).addListenersToCard();
```
creotutar, line 9: it may be more telling that someone is using the `InteractiveMap` by the presence of the `InteractiveMap` class in the document. `window.InteractiveMap` may just mean that they accidentally left the bundle on the page.

static/js/locator-bundle.js
```diff
@@ -5,3 +5,7 @@ export { InteractiveMap }
 import { NewMap } from './interactive-map/NewMap.js'
 window.NewMap = NewMap;
 export { NewMap }
+
+import { CardListenerAssigner } from './interactive-map/CardListenerAssigner.js'
+window.CardListenerAssigner = CardListenerAssigner;
```
oshi97, line 10: do we need to put this on the window?

templates/vertical-interactive-map/page.html.hbs
```diff
@@ -17,10 +17,10 @@
     {{> templates/vertical-interactive-map/script/locationbias modifier="main" }}
     {{> templates/vertical-interactive-map/script/locationbias modifier="mobileMap" }}
   {{/script/core }}
-  <script src="{{relativePath}}/locator-components.js" defer></script>
+  <script src="{{relativePath}}/locator-bundle.js" defer></script>
```
oshi97, line 20: I didn't know this bundle existed, would it make more sense to put the card listener class in here instead, if it's only meant to be used on the locator page?

cea2aj: I did put the card listener class in that bundle (rather than in HitchhikerJS)

static/js/locator-bundle.js
cea2aj: Since this is the locator bundle which doesn't put an object on the window such as HitchhikerJS, I do think we need to put it on the window unless there is another way to access it

templates/vertical-interactive-map/page.html.hbs
oshi97: sorry missed that

oshi97 (approved):

PR #626, opened 2021-02-26
Fix locator WCAG contrast issues

cea2aj:
The locator WCAG contrast issues by darkening the shade of grey

The default grey colors are fine against a white background, but since Product requested a grey background, the text doesn't meet the WCAG contrast ratio standard. Darken these values so that they pass the WCAG contrast test.

Note: Some of the other colors on the page don't pass the contrast test, but those aren't locator-specific so they are not addressed here

J=1059
TEST=manual

Run the WAVE evaluation tool and see that the locator colors pass the contrast tests

### DIFF
--- a/static/scss/answers/interactive-map/InteractiveMap.scss
+++ b/static/scss/answers/interactive-map/InteractiveMap.scss
@@ -12,6 +12,8 @@
 
   $border-default: 1px solid #666 !default;
 
+  $locator-text-gray: #737373;
+
   position: relative;
   display: flex;
   flex-direction: column;
@@ -238,7 +240,7 @@
       display: flex;
       align-items: center;
       border-top: var(--yxt-border-default);
-      color: #737373;
+      color: $locator-text-gray;
 
       &--desktop {
         display: none;
@@ -346,7 +348,7 @@
         border-top: var(--yxt-border-default);
         padding-left: 16px;
         padding-right: 16px;
-        color: #737373;
+        color: $locator-text-gray;
 
         @include bplte($mobile-break-point-max) {
           height: 60px;
@@ -565,7 +567,7 @@
   }
 
   &.CollapsibleFilters .yxt-VerticalResultsCount {
-    color: #737373;
+    color: $locator-text-gray;
   }
 
   &.CollapsibleFilters .Answers-viewResultsButton {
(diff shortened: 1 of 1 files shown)


static/scss/answers/interactive-map/InteractiveMap.scss
```diff
@@ -238,7 +238,7 @@
       display: flex;
       align-items: center;
       border-top: var(--yxt-border-default);
-      color: var(--yxt-locationBias-text-color);
+      color: #737373;
```
tmeyer2115, line 241: Could we make a variable for this?

cea2aj: Sure, are you think adding another grey to the answers-variables, or adding a scss variable to InteractiveMap.sscs?

tmeyer2115: I guess it depends on how often we feel the HHs will change this. If we think this is styling they will likely change, a CSS variable would probably be best. Otherwise, a scss variable in this file is fine.

cea2aj: Since these are overriding SDK colors which weren't previously modifiable by the theme, i'll make it a scss variable in InteractiveMap.scss

static/scss/answers/interactive-map/InteractiveMap.scss
```diff
@@ -563,6 +566,10 @@
     background-color: var(--hh-color-gray-2);
   }
 
+  &.CollapsibleFilters .yxt-VerticalResultsCount {
```
creotutar, line 569: Is this only a contrast problem when CollapsibleFilters are available? I know we add CollapsibleFilters automatically (regardless of if you use it or not) but it may be confusing to make it this CSS selector if this is a color change that should happen agnostic to collapsible filters existing

cea2aj: it's just due to the lighter grey background rather than the white that the theme uses by default, so i think I can remove the cfilters selector from here

static/scss/answers/interactive-map/InteractiveMap.scss
```diff
@@ -12,6 +12,8 @@
 
   $border-default: 1px solid #666 !default;
 
+  $locator-text-gray: #737373;
```
tmeyer2115, line 15: Is this the color of all text in the locator? If so, I think we should name the variable `locator-text-color` or something like that.

cea2aj: It's not the color all of the locator text (the locator has blue text for 'filter results' and 'update your location', and the location cards have black text), but it is the color used by all of the gray locator text

PR #627, opened 2021-02-26
Make locator SVGs pass w3 validation

cea2aj:
Update the mapIcon and listIcon so that they pass the https://validator.w3.org/

I removed the ids which were causing errors to be thrown in the validator. One of the errors was due to the slashes in the id, and the other was due to the duplicate "mask-2" ids. I added some tabs to so that the inline partial would be consistent with the other one.

J=SLAP-1059
TEST=manual

Build the site and confirm the SVGs still look good. Run the validator against the build and see the errors go away


creotutar (approved):
👍 👍 👍 LGTM

PR #628, opened 2021-03-01
Add translations for the "Close Card" string.

tmeyer2115:
This PR adds translations for the "Close Card" string that appears in some of
our result cards. French, German, Spanish, Italian, and Japanese translations
were all added.

TEST=none


cea2aj (approved):
lgtm

PR #630, opened 2021-03-01
Hide the yxtFooter when CFilters is active

cea2aj:
Hide the YxtFooter element when CFilters is active

Previously, we were just hiding the answers footer (which contains the yext logo and location info). We also need to hide the YxtFooter so that it doesn't show up on the CFilters page.

Solve this problem by hiding the YxtFooter similarly to how the answers footer is hidden, which is done by using interactions.js to toggle CSS classes on the specific elements.

The breakpoints are needed so that the element is not hidden when someone enters the filters page and then exits the filters page by dragging the window larger. The maps page is different in that the footer is hidden at a smaller breakpoint than the vertical standard or grid layouts.

J=SLAP-1006
TEST=manual

Override the layouts/footer.hbs and add some content. Confirm that the content is hidden when the CFilters page is open. Also test that dragging the brower window to make it larger and therefore disabling the CFilters page makes the YxtFooter show up again. Test both vertical-grid and vertical-map.


static/scss/answers/collapsible-filters/inactive-elements.scss
```diff
@@ -0,0 +1,5 @@
+@media (max-width: #{$breakpoint-collapsed-filters}) {
```
oshi97, line 1: before this styling was only being applied to vertical standard/grid, is it okay to have it apply to all page templates?

cea2aj: Ya, we definitely want to hide it on all page templates, however after double checking on vertical-map I realized that it's not optimal. This is because vertical map removes the answers-footer at a smaller breakpoint than vertical-grid and vertical-standard remove the answers-footer, so if we leave it as is we will remove the YxtFooter a little bit earlier than necessary on the vertical-map. We currently can't target different page templates for the YxtFooter because that footer is outside the div which contains the css class of the vertical page template. One solution would be to update the page templates to include the page template name on the Yxt-Page-wrapper div. How does that sound?

cea2aj: Update: That won't work because Yxt-Page-wrapper is outside of the scope of the page templates so there no straightforward way of specifying that per template without some hacky js. I'm investigating solutions for this through Interactions.js

cea2aj: 2nd Update: We can put the name of the template on the footer as a css class and target it that way

oshi97 (approved):

PR #631, opened 2021-03-02
 interactive-map: Fix scroll to result off by a few px

creotutar:
Previously, when clicking on a pin, the results list would move to the
card, except with a few px of padding between the top of the results
container and the selected card. This was not the case before the theme
migration (see the test site on staging).

To fix this, we change (and clean up) the way we find the scroll
distance and whether or not the card we are looking at is within view.

Note: Tested with a big header and header offset is not a variable in
calcualting the offset.

We also
(1) remove a useless `window.scroll` placement and
(2) move out non-class-specific code to a helper function

J=SLAP-1095
TEST=manual

Test that clicking on a pin correctly moves the card to the top of the
results container.

Try to find two cards that are right next to each other. Make sure being
at the lower card and then clicking the higher card takes you to the
higher card. Make sure being at the higher card and clicking on the
lower card in view will not scroll the results list.

### DIFF
--- a/static/js/interactive-map/InteractiveMap.js
+++ b/static/js/interactive-map/InteractiveMap.js
@@ -29,12 +29,6 @@ class InteractiveMap extends ANSWERS.Component {
     this._pageWrapperEl = document.querySelector('.YxtPage-wrapper');
 
     /**
-     * The header DOM element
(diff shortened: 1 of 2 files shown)


cea2aj (approved):
🎩

PR #635, opened 2021-03-03
interactive-map: Fix cluster click moves to wrong center

creotutar:
Previously, if you clicked on a pin cluster on a mapbxo map, it would
take you to the current center of map, except zoomed in. This behavior
was a result of conflicting setZoom/setCenter calls on the map.

Instead, we want Mapbox to set the zoom center in one call. It turns
out Consulting fixed this in a recent commit. We add these changes to our
Consulting map code to allow Mapbox to setZoomCenter in its own way that
does not give undesired behavior.

J=SLAP-1134
TEST=manual

Test that clicking on a mapbox pin cluster will zoom you and take you to
the center of the pins within that cluster. The behavior before was
obviously taking you to the wrong location on a pin cluster click.

### DIFF
--- a/static/js/interactive-map/Maps/ProviderMap.js
+++ b/static/js/interactive-map/Maps/ProviderMap.js
@@ -189,21 +189,19 @@ class ProviderMap {
     throw new Error('not implemented');
   }
 
-/**
+  /**
    * @param {number} zoom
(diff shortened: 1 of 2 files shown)


cea2aj (approved):
LGTM

PR #636, opened 2021-03-03
interactive-map: Add canvas click behavior

creotutar:
When you click on a pin or a result card, it is meant to show the card
and the pin in a selected state. It is desired behavior that when you
click on the map canvas the current pin and result card should not be
focused anymore.

Because Mapbox and Google handle click events separately, we have to
have provider-specific code for finding when a canvas is clicked.

J=SLAP-1086
TEST=manual

Test that you can click on a pin and then click anywhere on the map
canvas to remove the select styling on both the results list and the
pin.

On mobile, show that clicking on the map canvas when a map pin is
focused will close the detail card and move the map to full height.

Test that you can click on a pin and then directly another pin and these
changes do not cause any unintended behvaior.

Test that you can click on a result card and then click the map canvas
and both the pin and card remove their focused styling.

Test that clicking on the map control or any other pin element on the
map will not trigger a canvas click event.

### DIFF
--- a/static/js/interactive-map/NewMap.js
+++ b/static/js/interactive-map/NewMap.js
@@ -108,6 +108,7 @@ class NewMap extends ANSWERS.Component {
       .withPadding(this.config.mapPadding)
       .build();
     this.map = map;
+    window.mappp = this.map._map.map;
     this.addMapInteractions(map);
   }
 
@@ -139,6 +140,7 @@ class NewMap extends ANSWERS.Component {
         this.updateMapPropertiesInStorage();
         this.config.zoomEndListener(this.map.getZoom(), zoomTrigger);
       });
+      map.setCanvasClickHandler(() => this.config.canvasClickListener());
     });
 
     const mapRenderTargetOptions = new MapRenderTargetOptions()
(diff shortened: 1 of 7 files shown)


static/js/interactive-map/NewMap.js
```diff
@@ -108,6 +108,7 @@ class NewMap extends ANSWERS.Component {
       .withPadding(this.config.mapPadding)
       .build();
     this.map = map;
+    window.mappp = this.map._map.map;
```
oshi97, line 111: I don't have much context but what is this.map._map.map, and could we call it something other than window.mappp? I tried to trace the code and it looks like this.map._map is a ProviderMapOptions, but I couldn't find this.map inside of the ProviderMapOptions class. Maybe putting some method onto MapOptions to get this property would make it a little clearer?

creotutar: This was totally meant for personal debugging! Removing

oshi97 (approved):

cea2aj:
Do you know why the results aren't loading on the vertical-map page on Percy?

creotutar:
@cea2aj A re-run of the tests fixed it! Seems it was showing the results a bit slow when it was running my second commit tests.

cea2aj (approved):

PR #638, opened 2021-03-04
interactive-map: Remove results-specific storage listeners

creotutar:
Previously, we would keep listeners for card clicks in storage all
searches. This is a problem because part of the listener is relying on
the index of the pinned card remaining static. However, we cannot rely
on this because the index of cards can change with every search.

We remove any results-specific listeners when the map is updated with
new pins. That way, we do not keep our old listeners from pins in a
different search state.

J=None
TEST=manual

Test that you can click on a cluster pin and it then click on a card in
the new results list. This should highlight the corresponding pin on the
map.

### DIFF
--- a/static/js/interactive-map/NewMap.js
+++ b/static/js/interactive-map/NewMap.js
@@ -359,7 +359,6 @@ class NewMap extends ANSWERS.Component {
       callback: (data) => {
         const cardIndex = data.index;
         if (cardIndex + 1 === index) {
-          console.log('setting in card focus handler');
           this.core.storage.set(StorageKeys.LOCATOR_SELECTED_RESULT, id);
         }
       }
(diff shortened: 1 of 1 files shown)


static/js/interactive-map/NewMap.js
```diff
@@ -280,16 +286,19 @@ class NewMap extends ANSWERS.Component {
...
     const id = 'js-yl-' + entity.profile.meta.id;
 
-    this.core.storage.registerListener({
+    const cardFocusUpdateListener = {
       eventType: 'update',
       storageKey: StorageKeys.LOCATOR_CARD_FOCUS,
       callback: (data) => {
         const cardIndex = data.index;
         if (cardIndex + 1 === index) {
+          console.log('setting in card focus handler');
```
oshi97, line 295: is this an intentional log?

oshi97 (approved):

static/js/interactive-map/NewMap.js
creotutar: Gah, missed one! Removed!

oshi97 (approved):

oshi97 (approved):

PR #639, opened 2021-03-04
interactive-map: Add selected cluster pin styling on card click

creotutar:
This PR is split into two commits: One is adding changes to the
Consulting map code from their latest commits. The second is
building on top of that for our personal use case.

Previously, if you were using pin clustering, if you clicked on a result
card associated with a pin in a pin cluster, the map pin cluster would
not have selected styling. By default, this styling changes the color of
the cluster pin to separate it from other pins.

This change is important because we removed ordinals from cards. It is
important for users to be able to see that the card is associated with a
pin that is not on the map but rather in a cluster.

We build on top of Consulting locator code to store the id of the pins
and find the cluster if it exists.

J=SLAP-1085
TEST=manual

Test that you can click on a card in the results list and have it
highlight a cluster pin to indicate that the card exists in the cluster.

Test that you can click out of the selected cluster pin by clicking on
another card in the results list or on the canvas.

Test on both Google and Mapbox.

### DIFF
--- a/static/js/interactive-map/NewMap.js
+++ b/static/js/interactive-map/NewMap.js
@@ -229,7 +229,8 @@ class NewMap extends ANSWERS.Component {
   }
 
   /**
-   * Update the pin and map state with select result information.
+   * Update the pin and map state with information about the selected result. This result
+   * may be selected by either a pin click or a card click.
    *
    * @param {string} id
    * @param {MapPin[]} pins
@@ -244,7 +245,8 @@ class NewMap extends ANSWERS.Component {
   }
 
   /**
-   * Update the pin and map state with selected result information.
+   * Update the pin and map state with information about the selected result. This result
+   * may be selected by either a pin click or a card click.
    * This will check if the selected result, identified by id, is in a cluster in clusters
    * and enact changes according to the selected cluster.
    * If not, fallback to the normal selected result behavior.
--- a/static/js/interactive-map/NewMap.js
+++ b/static/js/interactive-map/NewMap.js
@@ -63,6 +63,12 @@ class NewMap extends ANSWERS.Component {
      * @type {string}
      */
     this.selectedPinId = null;
+
+    /**
+     * HTML element id for the selected cluster
+     * @type {string}
+     */
+    this.selectedClusterPin = null;
   }
 
   onCreate () {
@@ -147,8 +153,10 @@ class NewMap extends ANSWERS.Component {
       .withOnPostRender((data, map) => this.config.onPostMapRender(data, map, mapRenderTarget.getPins()))
       .withPinBuilder((pinOptions, entity, index) => this.buildPin(pinOptions, entity, index))
 
+    let pinClusterer;
     if (this.config.enablePinClustering) {
-      mapRenderTargetOptions.withPinClusterer(this.getClusterer());
+      pinClusterer = this.getClusterer();
+      mapRenderTargetOptions.withPinClusterer(pinClusterer);
     }
 
     const mapRenderTarget = mapRenderTargetOptions.build();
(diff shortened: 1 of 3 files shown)


static/js/interactive-map/NewMap.js
```diff
@@ -248,7 +313,9 @@ class NewMap extends ANSWERS.Component {
    * @param {Number} index The index of the entity in the result list ordering
    */
   buildPin(pinOptions, entity, index) {
+    const id = 'js-yl-' + entity.profile.meta.id;
```
cea2aj, line 316: for my own edification, what `yl` here for?

static/js/interactive-map/PinClusterer/PinClusterer.js
```diff
@@ -4,6 +4,33 @@ import { PinOptions } from '../Maps/MapPin.js';
 import { PinProperties } from '../Maps/PinProperties.js';
 import { Type, assertType, assertInstance } from '../Util/Assertions.js';
 
+/**
+ * Represents a cluster of {@link MapPin}s on a map.
+ * @member {MapPin} clusterPin
+ * @member {MapPin[]} pins
+ * @see {PinCluster}
```
cea2aj, line 11: do we need an '@see' here if the `PInCluster` class is right below it?

static/js/interactive-map/PinClusterer/PinClusterer.js
```diff
@@ -4,6 +4,33 @@ import { PinOptions } from '../Maps/MapPin.js';
...
+  }
+
+  /**
+   * Returns true if the cluster contains the pin with the given id, false otherwise
+   *
+   * @param {string} id The unique identifier for the pin
+   * @returns {boolean}
+   */
+  containsPin(id) {
+    return this.pins.filter((pin) => pin.getId() === id).length;
```
cea2aj, line 30: nit: we can remove the parenthesis around `pin` since it's the only param to the function

static/js/interactive-map/PinClusterer/PinClusterer.js
```diff
@@ -205,6 +295,13 @@ class PinClusterer {
     }
   }
 
+  /**
+   * @returns {PinCluster[]} Current pin clusters
+   */
+  getClusters() {
+    return [...this._clusters];
```
cea2aj, line 302: Is there a reason for the spread operator here rather than just returning this._clusters?

static/js/interactive-map/NewMap.js
creotutar: Just asked Consulting! This stands for yext locator, to differentiate from their other uses of some js `id` on the page.

static/js/interactive-map/PinClusterer/PinClusterer.js
creotutar: Deferring! This is coming directly from a Consulting commit. Adding to the description of SLAP-1094

static/js/interactive-map/PinClusterer/PinClusterer.js
creotutar: Not sure! It might be so that the array can be changed, but I know it doesn't stop the inner Pin Cluster object from being changed.

Deferring to SLAP-1094!

static/js/interactive-map/NewMap.js
```diff
@@ -183,37 +191,94 @@ class NewMap extends ANSWERS.Component {
...
+
+        if (this.selectedClusterPin) {
+          this.selectedClusterPin.setStatus({ selected: false });
+          this.selectedClusterPin = null;
+        }
 
-            if (this.config.onPinSelect) {
-              this.config.onPinSelect();
-            }
+        if (!id) {
```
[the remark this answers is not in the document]
creotutar: We'd still want to de-select if the current pin if the `LOCATOR_SELECTED_RESULT` is set to null (code above this if statement).

This short circuit is just for when we newly apply select state to a pin.

static/js/interactive-map/NewMap.js
```diff
@@ -183,37 +191,94 @@ class NewMap extends ANSWERS.Component {
...
+
+        if (this.config.onPinSelect) {
+          this.config.onPinSelect();
         }
       }
     });
   }
 
+  /**
+   * Update the pin and map state with select result information.
```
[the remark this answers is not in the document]
creotutar: Select result information should be a result that is selected, whether it be through a pin click or a card click! I changed the description to describe this more

cea2aj (approved):
🔥

oshi97 (approved):

PR #641, opened 2021-03-05
Improve Overlay Loading Speed

cea2aj:
Improve Overlay Loading Speed

The current answers overlay instructions waits until the page "load"
event before loading the overlay.js script. This can slow down the
loading of the overlay quite a bit since some sites may take several
seconds before firing that event. 

We can improve the performance by loading the overlay script
immediately. All of the other scripts and files that the overlay loads
will load much faster. When doing this, we need to wait to initialize
the overlay until the DOM is ready or else the overlay load will fail.

This PR is compatible with the original overlay instructions.
This PR allows the integration to be split into two scripts:
One for the overlay script and another for the overlay settings.

Add an overlay page for testing the overlay with the test site. The
file is minified because the "build-test-site" command will minify it
during each build. Checking in the html minified means that
running the build command won't trigger a git diff.

Z=388238
TEST=manual

Test this code on the Yanswers overlay integration and observe the
overlay script start loading much sooner (~80ms vs ~800ms).
When using the chrome dev tools "Fast 3G" Throttle, observe the
overlay appear after ~12 seconds compared to the previous time
of ~16 seconds.


tmeyer2115 (approved):

tmeyer2115:
LGTM! Have you chatted with Rose about updating the integration instructions to make use of this optimization?

cea2aj:
> LGTM! Have you chatted with Rose about updating the integration instructions to make use of this optimization?

Ya, I discussed the changes with her, and she was cool with them. When this is QA'd for 1.9, I can help her with the updated instructions

PR #648, opened 2021-03-09
Dev/faq accordion title

rosiegrant:
updating the faq-accordion title to first check the new "question" field. 

TEST: Confirmed locally with the same structure that the accordion card title still worked


tmeyer2115 (approved):
LGTM! When merging the branch, can you squash the commits?

PR #650, opened 2021-03-10
Remove default star icon behavior from Universal Section Templates.

tmeyer2115:
This PR adds a check to each of the section templates such that no Icon
component is rendered unless there is actually a `_config.icon`. What was
happening was that even if there was no icon config, a star icon component
would still get rendered.

J=SLAP-1170
TEST=manual

Verified that the default star icon no longer appeared in each of the section
templates. If an icon was specified in the config, though, it did appear
properly.


creotutar (approved):
⭐ ⭐ ⭐ LGTM
`happenning` in the PR message

PR #652, opened 2021-03-10
update details link url and text

rosiegrant:


directanswercards/documentsearch-standard/component.js
```diff
@@ -17,8 +17,8 @@ class documentsearch_standardComponent extends BaseDirectAnswerCard['documentsea
     return {
       value: answer.value,
       snippet: snippet && Formatter.highlightField(snippet.value, snippet.matchedSubstrings), // Text snippet to include alongside the answer
-      viewDetailsText: relatedItemData.title, // Text below the direct answer and snippet
-      viewDetailsLink: relatedItemData.url, // Link for the "view details" text
+      viewDetailsText: relatedItemData.fieldValues.name, // Text below the direct answer and snippet
```
tmeyer2115, line 20: I think we will need to add a null check here. Could we do something like:
```
const relatedItemData = relatedItem.data || {};
const fieldValues = relatedItemData.fieldValues || {};
```

rosiegrant: good call! would this work instead?
`viewDetailsText: relatedItemData.fieldValues && relatedItemData.fieldValues.name, // Text below the direct answer and snippet`

(could get confusing redefining `fieldValues`; we could do `relatedItemDataFieldValues = relatedItemData.fieldValues || {};` but just gets a little long)

tmeyer2115: Sure, the `&&` things sounds fine!

rosiegrant: discussed over slack, former is good!

tmeyer2115 (approved):

PR #653, opened 2021-03-10
Support the 4 column layout at all breakpoints

cea2aj:
Support the 4 column layout at all breakpoints

The layout wasn't working at breakpoints because there was a min-width on the placeholder cards which interfered with the 4 column width.  Removing that fixes it. In fact, we don't need the placeholders anymore since we are setting the widths  directly, so I removed those entirely.

The page width shrinks to take into account filters on the left side of the page. Since this styling is specific to CFilters, I moved the CSS to the CFilters styling. This means that the universal page no longer shrinks to accommodate filters since filters aren't supported on the universal page.

J=SLAP-1169
TEST=manual

Confirm that the 4 column layout works at all breakpoints. Test deleting some of the cards and observe them maintain the proper size. Also test this on the 3 other unviersalsectiontemplates. Observe that the container on the universal page maintains its larger size for longer since it is no longer shrinking to try to accommodate filters.

### DIFF
--- a/static/scss/answers/collapsible-filters/sdk-filter-overrides.scss
+++ b/static/scss/answers/collapsible-filters/sdk-filter-overrides.scss
@@ -8,6 +8,7 @@
       width: $container-width-at-collapsed-filters-breakpoint / $breakpoint-collapsed-filters * 100%;
     }
   }
+
   .yxt-FilterOptions {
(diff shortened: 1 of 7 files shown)


creotutar (commented):
🏠 🏠 🏠 LGTM

PR #656, opened 2021-03-13
cardcreator: Add partials path if cards/ dir already exists

creotutar:
We were seeing behavior where if the `cards/` directory already existed
in a repository (the new card directory at the top-level of a
repository), then the cardcreator would not automatically add "cards" as
a partial in the jambo.json. This is necessary to have jambo register
the new card partial.

We add a check to make sure it exists as a partial when a card is
created, instead of adding it only when the cards directory doesn't
exist.

We also do this for direct answer cards.

J=None
TEST=manual

Test on a local jambo repository

npx jambo card --name president-fork --templateCardFolder themes/answers-hitchhiker-theme/cards/standard

When creating a card and the cards directory already exists,
and the jambo.json partials doesn't contain the cards directory
and changing a cardType to use the new card (president-fork),
(1) the card is used
(2) the cards directory is not re-created
(3) the jambo.json is updated to include the partial cards

When creating a card and the cards directory does not already exist
and the jambo.json partials doesn't contain the cards directory
and changing a cardType to use the new card (president-fork),
(1) the card is used
(2) the cards directory is created
(3) the jambo.json is updated to include the partial cards

When creating a card and the cards directory already exists
and the jambo.json partials contains the cards directory
and changing a cardType to use the new card (president-fork),
(1) the card is used
(2) the cards directory is not re-created
(3) the jambo.json is not updated

All above test cases with direct answer cards as well.

### DIFF
--- a/commands/cardcreator.js
+++ b/commands/cardcreator.js
@@ -1,5 +1,5 @@
 const fs = require('fs-extra');
-const { addToPartials } = require('./helpers/utils/jamboconfigutils');
+const { containsPartial, addToPartials } = require('./helpers/utils/jamboconfigutils');
 const path = require('path');
 const UserError = require('./helpers/errors/usererror');
 const { ArgumentMetadata, ArgumentType } = require('./helpers/utils/argumentmetadata');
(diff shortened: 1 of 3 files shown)


bryanreed (changes requested):
lg although can we also make the change to the directanswercardscreator as well?  it has the same issue i am seeing

bryanreed (approved):

PR #661, opened 2021-03-18
Fix a bug with the price formatter.

tmeyer2115:
The price formatter was using parseInt for the string-ified price value. This
led to a loss of precision. Specifically, if a price was stored in the KG as
$10.99, our formatter would report it as $10.00. This is corrected by using
parseFloat.

TEST=auto

Updated the unit tests for the formatter to ensure the precision is not lost.


oshi97 (approved):
💸 💸 💸

PR #664, opened 2021-03-22
vertical-full-page-map: allow specification of clientId in config

oshi97:
clientId is a config option that can be specified for google maps
for authentication, as an alternative to the api key.
If both an api key and client id are passed to the google maps
script, the api key will take higher priority. Even if no apiKey
is explicitly provided in the page config, we automatically provide
a default, so this commit makes it so if a clientId is provided,
we no longer provide a default map api key.

J=SLAP-1114
TEST=manual,auto

test that I can specify a clientId. see that it will be used to load
in google maps (see "gme-yextinc" work and a bogus clientId not work)

see that I can still specify a custom api key
see that an apiKey is still provided by default it no clientId is specified

### DIFF
--- a/hbshelpers/chainedLookup.js
+++ b/hbshelpers/chainedLookup.js
@@ -9,7 +9,7 @@
 module.exports = function chainedLookup(context, ...lookupChain) {
   let lookupResult = context;
   for (const lookup of lookupChain.slice(0, -1)) {
-    if (!context || typeof context !== 'object') {
+    if (!lookupResult || typeof lookupResult !== 'object') {
       return '';
     }
     lookupResult = lookupResult[lookup];
--- a/hbshelpers/chainedLookup.js
+++ b/hbshelpers/chainedLookup.js
@@ -1,0 +1,18 @@
+/**
+ * A Handlebars helper that performs multiple lookups in a row.
+ *
+ * @param {Object} context The context to perform lookup on
+ * @param {...string} lookupChain The lookups to perform.
+ * @param {import('handlebars').HelperOptions} options
+ * @returns {any} The result of the lookup
+ */
+module.exports = function chainedLookup(context, ...lookupChain) {
+  let lookupResult = context;
+  for (const lookup of lookupChain.slice(0, -1)) {
+    if (!context || typeof context !== 'object') {
+      return '';
+    }
+    lookupResult = lookupResult[lookup];
+  }
+  return lookupResult;
+}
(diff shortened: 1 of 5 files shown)


hbshelpers/chainedLookup.js
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @param {Object} context The context to perform lookup on
+ * @param {...string} lookupChain The lookups to perform.
+ * @param {import('handlebars').HelperOptions} options
+ * @returns {any} The result of the lookup
+ */
+module.exports = function chainedLookup(context, ...lookupChain) {
+  let lookupResult = context;
+  for (const lookup of lookupChain.slice(0, -1)) {
+    if (!context || typeof context !== 'object') {
```
creotutar, line 12: is this meant to be `!lookupResult`?

oshi97: ah yes good catch thank you 🙏

oshi97: also my unit tests didn't get checked in by accident, added those in!

hbshelpers/chainedLookup.js
```diff
@@ -0,0 +1,18 @@
...
+ * @param {Object} context The context to perform lookup on
+ * @param {...string} lookupChain The lookups to perform.
+ * @param {import('handlebars').HelperOptions} options
+ * @returns {any} The result of the lookup
+ */
+module.exports = function chainedLookup(context, ...lookupChain) {
+  let lookupResult = context;
+  for (const lookup of lookupChain.slice(0, -1)) {
+    if (!lookupResult || typeof lookupResult !== 'object') {
+      return '';
```
creotutar, line 13: re: Returning empty string as the fail case -- I could see there being a time we want to know that it exists and it's just empty vs. the object is malformed. Do we want to return undefined here instead?

tests/hbshelpers/chainedLookup.js
```diff
@@ -0,0 +1,33 @@
+import chainedLookup from '../../hbshelpers/chainedLookup';
+
+it('performs a chained lookup', () => {
+  const context = {
+    a: {
+      b: {
+        c: 123
+      }
+    }
+  }
```
creotutar, line 10: nit: semicolon here please

hbshelpers/chainedLookup.js
oshi97: yeah makes sense!

creotutar (approved):
🆔 🆔 🆔 LGTM

Can we make sure the commit message says vertical full page map instead of locator?

PR #667, opened 2021-03-23
Update dependencies to address vulnerabilities

oshi97:
These are dependencies pointed out by dependabot by the hitchhiker
theme solution templates. For example
https://github.com/yext-pages/answers-jambo-template/pull/87
https://github.com/yext-pages/answers-jambo-template/pull/88
https://github.com/yext-pages/answers-jambo-template/pull/89

In the theme's package.json
- lodash 4.17.15 -> 4.17.20
this was done by adding lodash as a direct devDependency
The theme's package.json is technically unused by the solution templates.
But we might as well update our dev dependency to make
the dependabot warning go away.

In the theme's static/package.json, which is the one directly
used by the solution templates
- underscore.string 3.2.3 -> 3.3.5, also moved this from devDependencies to
  a regular dependency because it's used in the browser
- lodash 4.17.15 -> 4.17.20, this updates lodash in multiple @babel/ dependencies

J=SLAP-1196
TEST=manual

test that I can load up a map page (regular vertical-map page not the
vertical-full-page-map) on ie11, and the address will be displayed correctly

### DIFF
--- a/static/package.json
+++ b/static/package.json
@@ -37,7 +37,6 @@
     "resolve-url-loader": "^3.1.1",
     "sass-loader": "^8.0.2",
     "simple-git": "^2.15.0",
-    "underscore.string": "3.2.3",
     "webpack": "^5.4.0",
     "webpack-cli": "^4.2.0"
   },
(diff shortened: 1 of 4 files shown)


static/package.json
```diff
@@ -52,9 +51,11 @@
     "iframe-resizer": "^4.1.1",
     "is-mobile": "^2.2.2",
     "libphonenumber-js": "^1.7.51",
+    "lodash": "^4.17.20",
```
tmeyer2115, line 54: Kinda a cheat move, but I think you can put this in the `package.json`, run `npm install` to get the updates in the `package-lock`, then remove this from the `package.json`. I think that will get you what you want (updating the transitive lodash dependencies), without making lodash a top-level dependency.

oshi97: nifty!

tmeyer2115 (approved):

PR #668, opened 2021-03-23
Allow custom cards to be used with the `vertical` command.

tmeyer2115:
The `vertical` command had a bug that caused it to only allow built-in cards
to be specified as the `cardName` flag. This was overly restrictive. Users of
the command should be able to configure their vertical to use one of their
custom cards as well.

J=SLAP-1188
TEST=manual

Made a test site with custom cards. Verified that the card names appeared in
the `describe` output for `vertical`. Also verified that I could pass one
of the custom card names to the `cardName` flag and that the vertical would
be configured to use it properly.

### DIFF
--- a/commands/addvertical.js
+++ b/commands/addvertical.js
@@ -93,7 +93,7 @@ class VerticalAdder {
     const customCardsDir = 'cards';
     if (fs.existsSync(customCardsDir)) {
       fs.readdirSync(customCardsDir, { withFileTypes: true })
-        .filter(dirent => !dirent.isFile())
+        .filter(dirent => !dirent.isFile() && !cards.includes(dirent.name))
         .forEach(dirent => cards.push(dirent.name));
     }
 
--- a/commands/addvertical.js
+++ b/commands/addvertical.js
@@ -84,10 +84,20 @@ class VerticalAdder {
     if (!defaultTheme || !themesDir) {
       return [];
     }
-    const cardsDir = path.join(themesDir, defaultTheme, 'cards');
-    return fs.readdirSync(cardsDir, { withFileTypes: true })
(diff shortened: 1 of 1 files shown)


commands/addvertical.js
```diff
@@ -84,10 +84,20 @@ class VerticalAdder {
...
-    const cardsDir = path.join(themesDir, defaultTheme, 'cards');
-    return fs.readdirSync(cardsDir, { withFileTypes: true })
+    const themeCardsDir = path.join(themesDir, defaultTheme, 'cards');
+
+    const cards = fs.readdirSync(themeCardsDir, { withFileTypes: true })
       .filter(dirent => !dirent.isFile())
       .map(dirent => dirent.name);
+
+    const customCardsDir = 'cards';
+    if (fs.existsSync(customCardsDir)) {
```
oshi97, line 94: will this have duplicate cards if somebody overrides a theme one? like if somebody overrides faq-accordion will there be two faq-accordions in the array?

tmeyer2115: I believe we prevent people from naming their custom card the same as a built-in one. I just tried that and the `card` command threw an error saying the name was taken.

oshi97: ah I see. I think people can still shadow the card though? (I'm just remembering mercy overriding link-standard and faq-accordion but only the component.js parts)

oshi97: Still might be worth just having this return a set instead? Just on the off chance of something weird

tmeyer2115: Yeah, I suppose people could do that. Although, they really should just make a new card in that case. But, I can update to handle!

oshi97 (approved):

PR #669, opened 2021-03-23
Use transformFacets to apply the Facets fields config

cea2aj:
Add a transformFacets function which applies the config specified in
a Facet config

J=SLAP-1160
TEST=manual, auto

Build the test site and see fields option be applied on the page. Run
unit tests.


static/js/transform-facets.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * A Facets component transformFacets function which applies the configuration
```
tmeyer2115, line 2: It might be beneficial to use the verbiage "transform" here. This function transforms the Facets.fields component configuration provided by the Hitchhikers into a `transformFacets` function that can be understood by the SDK.

static/js/transform-facets.js
```diff
@@ -0,0 +1,43 @@
...
+ * @param {FilterOptionsConfig} config from answers-search-ui
+ * @returns {DisplayableFacet[]} from answers-core
+ */
+export default function transformFacets (facets, config) {
+  if(!config || !('fields' in config)) {
+    return facets;
+  }
+
+  return facets.map(facet => {
+    const isFacetFieldConfigSpecified = facet.fieldId in config.fields;
```
tmeyer2115, line 15: What would you think of naming this `isConfigurationForFacet` or something like that? I think a name like that is a bit clearer: we want to know if the HHs have added special config for a particular facet.

static/js/transform-facets.js
```diff
@@ -0,0 +1,43 @@
...
+    if (!isFacetFieldConfigSpecified) {
+      return facet;
+    }
+    const fieldConfig = config.fields[facet.fieldId];
```
tmeyer2115, line 19: Would it be more accurate to call this `facetConfig`? That name may make more sense in the context of this method. We are looking for custom facet configuration specified by the HHs and melding that with the `DisplayableFacet`s.

static/js/transform-facets.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * A Facets component transformFacets function which applies the configuration
+ * specified by the Facets 'fields' option
+ * 
+ * @param {DisplayableFacet[]} facets from answers-core
+ * @param {FilterOptionsConfig} config from answers-search-ui
+ * @returns {DisplayableFacet[]} from answers-core
```
tmeyer2115, line 7: Is the return type a union of `DisplayableFacet` and `FilterOptionsConfig`?

static/js/transform-facets.js
cea2aj: that works

static/js/transform-facets.js
cea2aj: I was thinking fieldConfig because it's keyed by `fieldId`, but I think `facetConfig` does make more sense

static/js/transform-facets.js
cea2aj: ah, ya, thanks for catching that

static/js/transform-facets.js
```diff
@@ -0,0 +1,43 @@
...
+
+  return facets.map(facet => {
+    const isConfigurationForFacet = facet.fieldId in config.fields;
+    if (!isConfigurationForFacet) {
+      return facet;
+    }
+    const facetConfig = config.fields[facet.fieldId];
+
+    const options = facet.options.map(option => {
+      if (!('fieldLabels' in facetConfig)) {
```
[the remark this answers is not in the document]
cea2aj: Good point! We don't need to loop through options if the check fails

tmeyer2115 (approved):

tmeyer2115 (approved):

PR #670, opened 2021-03-23
ThemeMapConfig: rename mapPadding to just padding

oshi97:
J=SLAP-1114
TEST=manual

see that I can set padding.bottom to 500 and the map center will be shifted
up a lot


creotutar (approved):
🗺️ 🗺️ 🗺️ LGTM

PR #671, opened 2021-03-24
vertical-full-page-map: Handle out of bounds longitude

creotutar:
Before, if you moved the map outside of the boundaries of longitude
[-180, 180], it was possible to get outside of the range. This would
cause liveapi to error because the longitude was incorrect. Here we mod
our longitudes so it wraps from -180 to 180 and stays within range.

Note: we do not need to do this for latitude because our current map
implementation does not let you wrap along the latitude lines.

J=SLAP-1107
TEST=manual

On a vertical full page map, zoom out to the maximum value and drag the
map to the left. Make sure when you're at the -180 longitude mark, if
you drag more to the left, you will be in the positive range and
correctly a liveapi request without an error.

Test this the positive longitude direction as well.

### DIFF
--- a/static/js/theme-map/Geo/GeoBounds.js
+++ b/static/js/theme-map/Geo/GeoBounds.js
@@ -1,5 +1,6 @@
 import { Unit, Projection } from './constants.js';
 import { Coordinate } from './Coordinate.js';
+import { getNormalizedLongitude } from '../Util/helpers.js';
 
 /**
  * This class represents a bounded coordinate region of a sphere.
@@ -121,25 +122,9 @@ class GeoBounds {
     const newLon = (nw.longitude + this._ne.longitude) / 2 + (this._ne.longitude < nw.longitude ? 180 : 0);
 
     nw.add(-latDist / 2, 0, Unit.DEGREE, projection);
-    nw.longitude = this._getNormalizedLongitude(newLon);
+    nw.longitude = getNormalizedLongitude(newLon);
 
     return nw;
-  }
-
-  /**
-   * Normalize lng to the range [-180, 180]. If you give -181, for
-   * example, we wrap back to 180.
-   *
-   * @param {Number} lng The longitude
-   * @returns {Number} The normalized longitude
-   */
-  _getNormalizedLongitude(lng) {
-    const positiveModLng = ((lng % 361) + 361) % 361;
-    const normalizedLng = positiveModLng > 180
-      ? positiveModLng - 361
-      : positiveModLng;
-
-    return normalizedLng;
   }
 }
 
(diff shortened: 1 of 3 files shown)


creotutar:
image

(0, -180)

static/js/theme-map/Geo/GeoBounds.js
```diff
@@ -121,10 +121,26 @@ class GeoBounds {
...
     nw.add(-latDist / 2, 0, Unit.DEGREE, projection);
-    nw.longitude = newLon;
+    nw.longitude = this._getNormalizedLongitude(newLon);
 
     return nw;
   }
+
+  /**
+   * Normalize lng to the range [-180, 180]. If you give -181, for
+   * example, we wrap back to 180.
```
oshi97, line 131: should -181 wrap back to 179? not a map expert but the math looked a little weird (using mod 361 instead of 360 or 180) and I think that might be why?

static/js/theme-map/Util/helpers.js
```diff
@@ -55,8 +55,25 @@ const isViewableWithinContainer = (targetEl, containerEl) => {
...
+ * prefer to return -180 over 180 when wrapping, as they are the same coordinate.
+ *
+ * @param {Number} lng The longitude
+ * @returns {Number} The normalized longitude
+ */
+const getNormalizedLongitude = (lng) => {
+  if (lng === 180) {
+    return lng;
+  }
+  return ((lng + 180) % 360 + 360) % 360 - 180;
```
cea2aj, line 71: Is there any way that you could add explanatory variables to help describe the how this formula works? Just looking at it, it's not clear to me how this formula was derived. If not, could you add a comment which shows your reasoning behind the formula?

oshi97: agree on the variables! just to double check I think the process is something like 
1. add 180 to convert to 0 to 360
2. mod 360 to normalize
3. add 360 take mod 360 AGAIN to make the mod positive (in javascript -4 % 10 is -4 not 6, but here we want 6)
4. subtract 180 to convert back to -180 to 180

static/js/theme-map/Util/helpers.js
```diff
@@ -55,8 +55,36 @@ const isViewableWithinContainer = (targetEl, containerEl) => {
...
+ * example, we wrap back to 179. If lng is 180, we return 180. Otherwise, we
+ * prefer to return -180 over 180 when wrapping, as they are the same coordinate.
+ *
+ * The idea is that we must mod by the range of the longitude values (360) to
+ * be span our entire range of values. In order to have our negative to
+ * positive wrapping work with modulus, the values we mod by need to be positive.
+ *
+ * 1. Add 180 to shift values to make sure -180 to 0 map to 0 to 180 
+ *    and 0 to 180 map to 180 to 360 for example
+ * 2. Mod by 360 to make the range (-360, 360)
```
cea2aj, line 70: After adding 180, you have a range of (0 to 360). When you mod any number in that range by 360, you get the same number, right? So why is modding necessary?

cea2aj: Ah, I see. It slipped my mind that `lng` can can be any negative or positive number. This  makes sense!

cea2aj (approved):
lgtm

oshi97 (approved):

PR #674, opened 2021-03-24
remove themes/answers-hitchhiker-theme prefix from custom commands

oshi97:
This commit removes the need to prefix all card/dacards paths with
"themes/answers-hitchhiker-theme" for the jambo card command.
Additionally, this also adds top-level cards to the jambo describes for
the card and dacard commands.

J=SLAP-1137
TEST=manual

test that the card command still works
test that the dacard command still works
run jambo describe, and see that card and dacard paths do not have the prefix. also see top level cards

### DIFF
--- a/commands/cardcreator.js
+++ b/commands/cardcreator.js
@@ -73,7 +73,7 @@ class CardCreator {
     if (!defaultTheme || !themesDir) {
       return [];
     }
-    const cardsDir = path.join(themesDir, defaultTheme, 'cards');
+    const themeCardsDir = path.join(themesDir, defaultTheme, 'cards');
     const cardPaths = new Set();
     const addCardsToSet = cardsDir => {
       if (!fs.existsSync(cardsDir)) {
@@ -83,7 +83,7 @@ class CardCreator {
         .filter(dirent => !dirent.isFile())
         .forEach(dirent => cardPaths.add(path.join('cards', dirent.name)));
     };
-    [cardsDir, 'cards'].forEach(dir => addCardsToSet(dir));
+    [themeCardsDir, 'cards'].forEach(dir => addCardsToSet(dir));
     return Array.from(cardPaths);
   }
 
--- a/commands/cardcreator.js
+++ b/commands/cardcreator.js
@@ -76,7 +76,7 @@ class CardCreator {
     const cardsDir = path.join(themesDir, defaultTheme, 'cards');
     return fs.readdirSync(cardsDir, { withFileTypes: true })
       .filter(dirent => !dirent.isFile())
-      .map(dirent => path.join(cardsDir, dirent.name));
+      .map(dirent => path.join('cards', dirent.name));
   }
 
   /**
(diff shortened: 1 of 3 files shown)


commands/cardcreator.js
```diff
@@ -113,13 +113,14 @@ class CardCreator {
     }
 
     const cardFolder = `${this._customCardsDir}/${cardFolderName}`;
-    if (fs.existsSync(templateCardFolder)) {
+    const themeCardFolder = path.join(this.config.dirs.themes, defaultTheme, templateCardFolder);
+    if (fs.existsSync(themeCardFolder)) {
```
cea2aj, line 117: Do we want to check if the file exists at the templateCardFolder first? That way checking inside the theme is the fallback rather than the only location which is checked. If we don't check the templateCardFolder first, then I think we lose the ability to create a card based on a top-level card

cea2aj:
Also, can you update the README to reflect this change?

commands/cardcreator.js
oshi97: wow I didn't know we could do that! top level cards are not included in jambo describe (in this PR or before), but I can fix that here

cea2aj: Oh, maybe it's not officially supported (i'm also not sure if anyone uses that functionality). However I do remember Bryan saying that in general for describe and commands, that our logic should first check at the top level and then check the theme as a fallback

oshi97: sure makes sense, i can imagine somebody wanting to copy a top level card and make small tweaks

oshi97: actually hMMmmM it's possible they actually don't want the extra options (thinking of something like yanswers with its 20 different cards), just asked Tom what his opinion his

commands/cardcreator.js
```diff
@@ -74,9 +74,17 @@ class CardCreator {
       return [];
     }
     const cardsDir = path.join(themesDir, defaultTheme, 'cards');
```
cea2aj, line 76: nit: thoughts on renaming to `themeCardsDir`?

commands/directanswercardcreator.js
```diff
@@ -74,9 +74,17 @@ class DirectAnswerCardCreator {
       return [];
     }
     const daCardsDir = path.join(themesDir, defaultTheme, 'directanswercards');
```
cea2aj, line 76: similar nit here about possibly renaming to `themeDaCardsDir`

commands/cardcreator.js
oshi97: honestly I don't want to make more differences between the two commands

cea2aj: ok

oshi97: we originally made them two separate commands to keep them dumb and decoupled, but there have been literally 0 divergences between them in the past year, besides the different folder name

oshi97: I'll rename them both to themeCardDir though

commands/cardcreator.js
oshi97: Tom gave the ok!

cea2aj (approved):
lgtm

PR #675, opened 2021-03-24
Set anchorX and anchorY in ThemeMapConfig

oshi97:
This commit sets anchorX and anchorY for map pins
and cluster pins using config specified in ThemeMapConfig.js

J=SLAP-1114
TEST=manual

test that I can set the anchor values, and when I check the
transform: translate(percentX, percentY) in dev tools, the
css transform is the expected value for the anchorX/Y I set

### DIFF
--- a/static/js/theme-map/ThemeMapConfig.js
+++ b/static/js/theme-map/ThemeMapConfig.js
@@ -26,6 +26,7 @@ export default class ThemeMapConfig {
 
     /**
      * Controls the visual offset of each pin.
+     * @type {Object}
      */
     this.pinAnchors = {
       anchorX: 0.5,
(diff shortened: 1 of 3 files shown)


static/js/theme-map/ThemeMapConfig.js
```diff
@@ -24,6 +24,22 @@ export default class ThemeMapConfig {
      */
     this.apiKey = jsonConfig.apiKey;
 
+    /**
+     * Controls the visual offset of each pin.
```
creotutar, line 28: nit: do we want the @type here too?

oshi97: I feel like you can tell it's an object by looking at it but I'll add for consistency

tmeyer2115 (approved):

PR #678, opened 2021-03-25
Keyboard support for "Search When Map Moves"

cea2aj:
Make the "Search When Map Moves" feature work with the keyboard

Previously the "Search When Map Moves" feature only worked with the mouse because it was relying on the `dragEndHandler` rather than the `panHandler`. By switching to the `panHandler`, the feature will work with the keyboard. As a result of this change, I split up the denouncer so that zoom changes and map center changes are debounced separately.

Also, after a user runs a search, or after a user focuses on a pin which is at the edge of the screen, we don't want to run a new search. To prevent this, I created `PanTriggers` which are analogous to `ZoomTriggers` so that we could prevent running new searches in these cases.

The Google Maps 'idle' event is called repeatedly when the map is moves with the keyboard, which I believe is a bug, so in order to get around this I used a debouncer so that Google Maps only fires the `panHandler`once the map is idle for 100ms.

J=SLAP-1141
TEST=manual

Test on both Mapbox and on Google Maps that "Search When Map Moves" works when navigating the map with the keyboard. Confirm that running a search or focusing on a pin near the edge of the screen doesn't run a new search. Confirm that clicking on a cluster does run a new search after zooming in. Test that I can tab through results without new searches being ran

### DIFF
--- a/static/js/theme-map/Maps/PanTriggers.js
+++ b/static/js/theme-map/Maps/PanTriggers.js
@@ -1,0 +1,9 @@
+/**
+ * Describes the types of triggers for a map pan
+ *
+ * @enum {string}
+ */
+export default {
+  UNSET: '',
+  API: 'api',
+};
--- a/static/js/theme-map/Maps/PanTriggers.js
+++ b/static/js/theme-map/Maps/PanTriggers.js
@@ -1,9 +1,23 @@
 /**
- * Describes the types of triggers for a map pan
+ * Describes the possible of triggers for a map pan
  *
  * @enum {string}
  */
 export default {
+  /**
+   * Indicates that the panTrigger is not set, and therefore there is no specific
+   * behavior that should occur if the panHandler is called with this PanTrigger.
+   * 
+   * This should be set if a user moves the map, or if a user clicks on a cluster.
+   */
   UNSET: '',
+  /** 
+   * Indicates that the map is panning due to programatic reason, and therefore a new
+   * search should not be ran if the panHandler is called while this PanTrigger is set
+   * on the Map.
+   * 
+   * This includes panning the map after a new search is ran, or centering the map over
+   * a focused pin after clicking or tabbing onto it.
+   */
   API: 'api',
 };
--- a/static/js/theme-map/Maps/PanTriggers.js
+++ b/static/js/theme-map/Maps/PanTriggers.js
@@ -1,5 +1,5 @@
 /**
- * Describes the possible of triggers for a map pan
+ * Describes the possible triggers for a map pan
  *
  * @enum {string}
  */
@@ -7,8 +7,6 @@ export default {
   /**
    * Indicates that the panTrigger is not set, and therefore there is no specific
    * behavior that should occur if the panHandler is called with this PanTrigger.
-   * 
-   * This should be set if a user moves the map, or if a user clicks on a cluster.
    */
   UNSET: '',
   /** 
@@ -16,8 +14,9 @@ export default {
    * search should not be ran if the panHandler is called while this PanTrigger is set
    * on the Map.
    * 
-   * This includes panning the map after a new search is ran, or centering the map over
-   * a focused pin after clicking or tabbing onto it.
+   * This includes the automatic centering of the map after a new search is ran, or the
+   * automatic centering of the map over a focused pin near the edge of the screen after
+   * clicking or tabbing onto it.
    */
   API: 'api',
 };
--- a/static/js/theme-map/Maps/PanTriggers.js
+++ b/static/js/theme-map/Maps/PanTriggers.js
@@ -5,10 +5,15 @@
  */
 export default {
   /**
-   * Indicates that the panTrigger is not set, and therefore there is no specific
-   * behavior that should occur if the panHandler is called with this PanTrigger.
+   * Indicates that the panTrigger is not set.
    */
   UNSET: '',
+  /**
+   * Indicates that the pan occured as a result of user interaction.
+   * 
+   * This includes moving the map or clicking on a pin cluster.
+   */
+  USER: 'user',
   /** 
    * Indicates that the map is panning due to programatic reason, and therefore a new
    * search should not be ran if the panHandler is called while this PanTrigger is set
--- a/static/js/theme-map/ThemeMap.js
+++ b/static/js/theme-map/ThemeMap.js
@@ -139,7 +139,10 @@ class ThemeMap extends ANSWERS.Component {
    */
   addMapInteractions(map) {
     this.map.idle().then(() => {
-      map.setPanHandler(() => this.updateMapPropertiesInStorage());
+      map.setPanHandler((prevousBounds, currentBounds) => {
+        this.updateMapPropertiesInStorage();
+        this.config.panHandler(prevousBounds, currentBounds);
+      });
       map.setDragEndHandler(() => {
         this.updateMapPropertiesInStorage();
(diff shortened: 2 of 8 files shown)


cea2aj:
I noticed that the panHandler for MapBox is only called after the map completely stops moving, however for Google Maps, the panHandler is called repeatedly as the map moves. This means that the map re-renders as it goes outside of the denouncer threshold on Google, but on MapBox it doesn't re-render until the user has stopped moving the map. 

I'm wondering if it's worth adding another check for google maps so that the panHandler doesn't run if an arrow key is being pressed

static/js/theme-map/Maps/PanTriggers.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Describes the possible of triggers for a map pan
```
tmeyer2115, line 2: Describes the possible triggers for a map pan

static/js/theme-map/Maps/PanTriggers.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @enum {string}
+ */
+export default {
+  /**
+   * Indicates that the panTrigger is not set, and therefore there is no specific
+   * behavior that should occur if the panHandler is called with this PanTrigger.
+   * 
+   * This should be set if a user moves the map, or if a user clicks on a cluster.
```
tmeyer2115, line 11: I think the terminology here is a bit confusing. If a user moves the map, that sounds to me like they did "pan" the map. Does pan have a specific definition in the context of WCAG?

static/js/theme-map/ThemeMap.js
```diff
@@ -139,7 +139,10 @@ class ThemeMap extends ANSWERS.Component {
    */
   addMapInteractions(map) {
     this.map.idle().then(() => {
-      map.setPanHandler(() => this.updateMapPropertiesInStorage());
+      map.setPanHandler((prevousBounds, currentBounds, zoomTrigger) => {
+        this.updateMapPropertiesInStorage();
+        this.config.panHandler(prevousBounds, currentBounds, zoomTrigger);
```
tmeyer2115, line 144: Is the handler fired on the end of panning?

static/js/theme-map/ThemeMapConfig.js
```diff
@@ -186,6 +186,12 @@ export default class ThemeMapConfig {
      */
     this.dragEndListener = jsonConfig.dragEndListener || function () {};
 
+    /**
+     * Callback for when a map pan event has finished
+     * @type {Function}
+     */
+    this.panHandler = jsonConfig.panHandler || function () {};
```
tmeyer2115, line 193: What do you think of calling this `panEndListener` for consistency? It may also be a bit more descriptive as the listener is fired on the end of the pan event

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -157,21 +159,28 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
     const pinClusterClickListener = () => this.searchOnMapMove && this.searchThisArea();
 
     /**
-     * Dragging the map searches the new area, if desired
-     *
-     * @param {Map} map The map object
+     * The listener called when the map drag ends
      */
-    const dragEndListener = () => this.handleMapAreaChange();
+    const dragEndListener = () => {};
```
tmeyer2115, line 164: Do we sill need the `dragEndListener`?

static/js/theme-map/Maps/PanTriggers.js
cea2aj: I'm using the word 'pan' in the same way that Google Maps and MapBox use it, which means that the map moved. Would it help if I removed the second sentence here?

static/js/theme-map/ThemeMap.js
cea2aj: For MapBox it isn't fired until the map has stopped moving completely. On Google Maps it gets called as the map moves.

static/js/theme-map/ThemeMapConfig.js
cea2aj: I left it as is because it was already named `panListener ` in Map.js (consulting's code). Also, it would be misleading because Google Maps fires repeatedly. We could make a new handler called `panEndListener`
which is the same as the `panHandler` for MapBox, and for Google Maps, we debounce it and call it when the map has settled. What do you think of that idea?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
cea2aj: nope. I left it in here so that it would be easy to modify if we ever wanted to change it, however I could remove it if you'd prefer

tmeyer2115: If it's not used right now, I think we should remove. We can add it back in if we ever need it.

static/js/theme-map/Maps/PanTriggers.js
tmeyer2115: I guess the thing that's tripping me up is that if a user moves the map, it leads to a trigger of `UNSET`. In that case, it seems like the user took a manual action that triggered a pan of the map, almost as if the trigger should be `MANUAL` or something.

cea2aj: My thinking here is that "API" is the only trigger that we care about for this item, so it felt unnecessary to set a "USER" trigger, however I could add one to match the ZoomTriggers (which have API, USER, and UNSET)

static/js/theme-map/ThemeMapConfig.js
cea2aj: Correct, that would mean that Google would would have both a `panHandler` and a `panEndListener`, however our code would only use the `panEndListener`

static/js/theme-map/Maps/PanTriggers.js
cea2aj: Update: just added PanTriggers.User which behaves the same as the User ZoomTrigger

static/js/theme-map/ThemeMapConfig.js
creotutar: I would feel like we want to only `updateMapPropertiesInStorage` on panEnd. What do you think of making that `panEnd` like you suggested, following the zoomend in Google's case?

cea2aj: Sure, I can make a `panEndListener`, but what do you mean by "following the zoomend in Google's case"?

creotutar: The case where we simulate a zoomend by listening for idle once, after a zoom_changed event!

cea2aj: The Google maps 'idle' event works correctly when dragging the map by mouse, but it fires repeatedly when the keyboard is used, which suggests that this is a Google maps issue. I have tried preventing calling the `panHandler` if the navigation keys are pressed, however it appears that Google's map captures the navigation key presses, and therefore I'm unable to detect if they are pressed from our javascript. So at this point, we have a. couple of options:
1. Move forward with this PR as-is
2. Add a simple time-based debouncer to help reduce the frequency of panHandler calls performed when the map is moving.
I also think we may be able to optimize the javascript on the main thread so that we don't drop frames every time a search is performed, but that's out of scope of this item. What are your thoughts?

tmeyer2115 (approved):

PR #680, opened 2021-03-26
vertical-full-page-map: location card click listener only on desktop

creotutar:
Previously, if you were on a mobile device and you clicked on a card,
the "selecting" behavior from desktop would occur. That is, the card
would enter its selected state, which is opening a detail card. This is
not correct because the only entry to a selected result should be
through a pin click on the map view when on mobile.

To fix this, we only enact the selected listeners on card click when we
are on desktop.

J=SLAP-1202
TEST=manual

Test that on mobile/tablet
* Clicking on a result card does not hide the card anymore
* Clicking on a pin in the map view will show a detail card

Test that on desktop
* Clicking on a result card still selects the correct pin (the selected
  styling on the pin shows).
* Clicking on a result card also selects a cluster pin if applicable.

### DIFF
--- a/static/js/theme-map/CardListenerAssigner.js
+++ b/static/js/theme-map/CardListenerAssigner.js
@@ -9,6 +9,20 @@ class CardListenerAssigner {
      * @type {Answers.Component}
      */
     this.card = card;
+
+    /**
+     * The upper bound (inclusive) of the mobile breakpoint in px
+     *
+     * @type {Number}
+     */
+    this.mobileBreakpointMax = 991;
+
+    /**
+     * The matcher to determine if the window width is within the mobile breakpoint
+     *
+     * @type {MediaQueryList}
+     */
+    this.mediaMatcher = window.matchMedia(`(max-width: ${this.mobileBreakpointMax}px)`);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


static/js/theme-map/CardListenerAssigner.js
```diff
@@ -9,6 +9,20 @@ class CardListenerAssigner {
...
+     * @type {Number}
+     */
+    this.mobileBreakpointMax = 991;
+
+    /**
+     * The matcher to determine if the window width is within the mobile breakpoint
+     *
+     * @type {MediaQueryList}
+     */
+    this.mediaMatcher = window.matchMedia(`(max-width: ${this.mobileBreakpointMax}px)`);
```
oshi97, line 25: could we name this mobileMediaMatcher or something that makes it obvious it this.mediaMatcher.matches returns true for mobile?

static/js/theme-map/CardListenerAssigner.js
```diff
@@ -9,6 +9,20 @@ class CardListenerAssigner {
      * @type {Answers.Component}
      */
     this.card = card;
+
+    /**
+     * The upper bound (inclusive) of the mobile breakpoint in px
+     *
+     * @type {Number}
+     */
+    this.mobileBreakpointMax = 991;
```
oshi97, line 18: would we ever use this variable directly as opposed to this.mediaMatcher?

static/js/theme-map/CardListenerAssigner.js
creotutar: Good point! Changed

static/js/theme-map/CardListenerAssigner.js
creotutar: We would not! Removed the one-use variable

oshi97 (approved):

cea2aj (approved):
lgtm

PR #681, opened 2021-03-29
Intelligent defaults for `vertical` command's `cardName`.

tmeyer2115:
This PR makes the `cardName` defaulting logic more intelligent. Initially,
if no `cardName` was supplied, a default of `standard` was used. But, all
vertical templates prescribe their own default `cardType`. This is now used
as the default. If the template has no `cardType` (which is not currently
the case), a default of `standard` will still be used.

A validation layer was added for the `vertical` command, on top of the
structural validation done already by YArgs. This layer will throw UserErrors
if certain conditions are not met. It allows the other code in the class to
make certain assumptions safely.

J=SLAP-1186
TEST=manual

Tested the following:
- If an invalid `cardName` is provided, an error is thrown.
- If an invalid `template` is provided, an error is thrown.
- If a valid `cardName` is specified, that appears in the new vertical page's
  config as the `cardType`.
- If no `cardName` is specified, but the template has a default `cardType`,
  that is used over `standard`.
- If no `cardName` is specified and the template has no default `cardType`,
  then `standard` is used.

### DIFF
--- a/commands/addvertical.js
+++ b/commands/addvertical.js
@@ -211,6 +211,8 @@ class VerticalAdder {
     const configFile = `config/${name}.json`;
     let parsedConfig = parse(fs.readFileSync(configFile, { encoding: 'utf-8' }));
 
+    parsedConfig.verticalKey = verticalKey;
+
     parsedConfig.verticalsToConfig[verticalKey] = 
       parsedConfig.verticalsToConfig['<REPLACE ME>'];
     delete parsedConfig.verticalsToConfig['<REPLACE ME>'];
--- a/commands/addvertical.js
+++ b/commands/addvertical.js
@@ -1,5 +1,6 @@
 const fs = require('fs-extra');
 const path = require('path');
+const { parse, stringify } = require('comment-json');
 const { spawnSync } = require('child_process');
 
 const UserError = require('./helpers/errors/usererror');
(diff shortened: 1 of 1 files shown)


commands/addvertical.js
```diff
@@ -1,5 +1,6 @@
 const fs = require('fs-extra');
 const path = require('path');
+const { parse, stringify } = require('comment-json');
```
cea2aj, line 3: Note: 'comment-json' was added in theme v1.17, which means when we create the updated patch files for themes older than 1.16, the patch files will need to modify the correct package.json file to include the dependency

tmeyer2115: Good catch. Hopefully this won't make it much more difficult to craft the pre-1.16 patch files.

commands/addvertical.js
```diff
@@ -119,12 +120,71 @@ class VerticalAdder {
...
+   * @param {Object} args The command parameters.
+   */
+  _validateArgs(args) {
+    if (args.template === 'universal-standard') {
+      throw new UserError('A vertical cannot be initialized with the universal template');
+    }
+
+    const themeDir = this._getThemeDirectory(this.config);
+    const templateDir = path.join(themeDir, 'templates', args.template);
+    if (!fs.existsSync(templateDir)) {
```
cea2aj, line 144: I remember Bryan discussing that in general, we should first check for folders in the top-level of the site, and then check inside the theme as a fallback. Do you think that we should do the same here as well?

tmeyer2115: Hmm, that's not a use case that Product anticipated. But, it could happen! Someone could shadow the `templates` directory and we'd want to honor that here. If you don't mind, I can do that as a separate item/PR. The command in it's current state doesn't support this either.

cea2aj: Sure, I think a separate item for that makes sense

cea2aj (approved):
lgtm

commands/addvertical.js
cea2aj: As @oshi97 pointed out, comment-json is actually in the theme all the way back to v1.8 (and possibly a little earlier), so we probably don't have to worry about adding the dependency. I was mistakenly looking at package.json rather than static/package.json

PR #682, opened 2021-03-29
Document search card tweaks

rosiegrant:
Addressing two QA items:

1) Fall back to relatedItemData.fieldValues.landingPageUrl

2) Update the default on the documentsearch card to be "Read more about" instead of "Read more from"

### DIFF
--- a/directanswercards/documentsearch-standard/component.js
+++ b/directanswercards/documentsearch-standard/component.js
@@ -18,7 +18,7 @@ class documentsearch_standardComponent extends BaseDirectAnswerCard['documentsea
       value: answer.value,
       snippet: snippet && Formatter.highlightField(snippet.value, snippet.matchedSubstrings), // Text snippet to include alongside the answer
       viewDetailsText: relatedItemData.fieldValues && relatedItemData.fieldValues.name, // Text below the direct answer and snippet
-      viewDetailsLink: relatedItemData.website, // Link for the "view details" text
+      viewDetailsLink: relatedItemData.website || relatedItemData.fieldValues && relatedItemData.fieldValues.landingPageUrl, // Link for the "view details" text
       viewDetailsEventOptions: this.addDefaultEventOptions({
         ctaLabel: 'VIEW_DETAILS'
       }), // The event options for viewDetails click analytics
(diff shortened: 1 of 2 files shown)


directanswercards/documentsearch-standard/component.js
```diff
@@ -18,7 +18,7 @@ class documentsearch_standardComponent extends BaseDirectAnswerCard['documentsea
       value: answer.value,
       snippet: snippet && Formatter.highlightField(snippet.value, snippet.matchedSubstrings), // Text snippet to include alongside the answer
       viewDetailsText: relatedItemData.fieldValues && relatedItemData.fieldValues.name, // Text below the direct answer and snippet
-      viewDetailsLink: relatedItemData.website, // Link for the "view details" text
+      viewDetailsLink: relatedItemData.website || relatedItemData.fieldValues && relatedItemData.fieldValues.landingPageUrl, // Link for the "view details" text
```
tmeyer2115, line 21: I'm wondering if it will be difficult for the Hitchhikers to understand the order of operations here (the `&&` is grouped more tightly than the `||`). It may be easier for them to read if we had something like:
```
const fieldValues = relatedItemData.fieldValues || {};
...
viewDetailsLink: `relatedItemData.website` || `fieldValues.landingPageUrl`;
```

rosiegrant: yes good point! I polled a couple and they got it (I think the && is present in so many places and it luckily reads a little like a sentence). my other reservation is that adding another constant might get a little long (https://github.com/yext/answers-hitchhiker-theme/pull/652#discussion_r592566504), and I think it's slightly clearer to see the points of failure this way. but can change if we feel strongly, or maybe add parenthesis to make it as obvious as possible `relatedItemData.website || (relatedItemData.fieldValues && relatedItemData.fieldValues.landingPageUrl)`?

rosiegrant: @tmeyer2115 bump on this when you get a chance, thanks!

tmeyer2115: Okay, thanks for polling! If you don't mind, I think we should add some parentheses to make it as obvious as possible. I think even experience programmers can get tripped up by expressions like this, some languages have different binding rules for operators.

tmeyer2115 (approved):

PR #683, opened 2021-03-29
Comment in default initial search

rosiegrant:
Now that we have add vertical in the theme, we can have a default initial search set by default. 

(Note the search bar config already allows an empty search by default in all of the page config templates)


tmeyer2115 (approved):
This lgtm! We will need to update the `vertical` command to make sure it replaces this new, uncommented instance of `<REPLACE ME>` as well. Currently, it just replaces the one at the top and in `verticalsToConfig`.

tmeyer2115:
We can handle the `vertical` command update though!

rosiegrant:
@tmeyer2115 thanks! ah I could've sworn it replaced it but yes good call  💯

PR #684, opened 2021-03-29
Fix the vertical full page map zooming out quickly bug

cea2aj:
Fix a bug that incorrectly caused the map the zoom back in while
zooming out

The issue was that the setting and the deleting of `LOCATOR_FROM_SEARCH_THIS_AREA` would sometimes get out of sync when zooming out of the map quickly. This caused the map to automatically update the zoom, which made it difficult to zoom out in some locations because the map would keep zooming back in automatically.

This PR solves this by keeping track of the number of concurrent calls of `searchThisArea`, and only updating the map zoom if that value is equal to zero.

J=SLAP-1142
TEST=manual

Test zooming out quickly and confirm that when the the map does not automatically zoom back in in situations where it would previously occur. Test that performing a search correctly updates the zoom automatically.

### DIFF
--- a/static/js/theme-map/ThemeMap.js
+++ b/static/js/theme-map/ThemeMap.js
@@ -397,9 +397,12 @@ class ThemeMap extends ANSWERS.Component {
     const universalData = transformDataToUniversalData(data);
     let entityData = verticalData.length ? verticalData : universalData;
 
-    const fromSearchThisArea = this.core.storage.get(StorageKeys.LOCATOR_FROM_SEARCH_THIS_AREA);
-    this.core.storage.delete(StorageKeys.LOCATOR_FROM_SEARCH_THIS_AREA);
-    let updateZoom = !fromSearchThisArea;
+    const numConcurrentSearchThisAreaCalls = this.core.storage.get(StorageKeys.NUM_CONCURRENT_SEARCH_THIS_AREA_CALLS);
+    if (numConcurrentSearchThisAreaCalls > 0) {
+      this.core.storage.set(StorageKeys.NUM_CONCURRENT_SEARCH_THIS_AREA_CALLS, numConcurrentSearchThisAreaCalls - 1); 
+    }
+
+    let updateZoom = numConcurrentSearchThisAreaCalls <= 0;
 
     const isNoResults = data.resultsContext === 'no-results';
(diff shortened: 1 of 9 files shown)


static/js/theme-map/ThemeMap.js
```diff
@@ -397,9 +397,17 @@ class ThemeMap extends ANSWERS.Component {
...
+      this.core.storage.get(StorageKeys.LOCATOR_NUM_CONCURRENT_SEARCH_THIS_AREA_CALLS);
+
+    if (numConcurrentSearchThisAreaCalls > 0) {
+      this.core.storage.set(
+        StorageKeys.LOCATOR_NUM_CONCURRENT_SEARCH_THIS_AREA_CALLS,
+        numConcurrentSearchThisAreaCalls - 1
+      ); 
+    }
+
+    let updateZoom = numConcurrentSearchThisAreaCalls <= 0;
```
creotutar, line 410: if this isn't changing can we make it a const?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -415,6 +415,14 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
    * the visible area.
    */
   searchThisArea() {
+    const numConcurrentSearchThisAreaCalls =
+      this.core.storage.get(StorageKeys.LOCATOR_NUM_CONCURRENT_SEARCH_THIS_AREA_CALLS);
+    const updatedNumSearchThisAreaCalls = numConcurrentSearchThisAreaCalls + 1 || 1;
+    this.core.storage.set(
+      StorageKeys.LOCATOR_NUM_CONCURRENT_SEARCH_THIS_AREA_CALLS,
```
creotutar, line 422: From our conversation yesterday: It is unfortunate that the only way, currently, to communicate that the trigger of a search (to all of the people listening to a search) is through this asynchronous storage. However, we do not have a way to go around this. Just a note.

creotutar (approved):
👍 👍 👍 LGTM

static/js/theme-map/ThemeMap.js
cea2aj: it changes so it needs to be a `let`

PR #685, opened 2021-03-30
vertical-full-page-map: Keep height on map view fixed

creotutar:
Previously, on mobile phone browsers with navigation bars, we would see
that 100vh would not necessarily be the 100% of the viewport visible to
the user. For instance, on Chrome mobile for iOS, 100vh is obscured by
about 60px to make room for the Chrome navigation tools. This is not a
problem in the mock-mobile on Chrome desktop.

We keep the height of the map view fixed when it is in view. That way,
you are always seeing the important parts of the map. and cannot scroll
away from it. Because of this change, we must fix a few other things,
including the detail card and the wrapper of the map (so we cannot
invisibly scroll to the hidden footer).

NOTE: The Answers logo footer and the customer footer will no longer be
visible on the map view. They will only be visible on the list view.

NOTE: Browserstack s21 Chrome, Firefox will NOT work currently, because
there is a bug with the Navigation component styling causing an overflow.

J=SLAP-1207
TEST=manual

Test on physical iOS device (iPhone 8+) Chrome Safari
Browserstack iPhone 12 Chrome, Safari
Test on Chrome, Firefox, Safari locally on MacOS

You scroll down the results list and see the Answers logo footer and
custom footer.

You can click the map toggle and see the Search This Area and Location
Bias immediately. It is not obscured by any navigation tools or cut off
in the viewport. You cannot scroll in the map view.

You can click on a pin and see its detail card with the correct height.

### DIFF
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -365,6 +365,7 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
           window.scrollTo(0, 0);
           this._container.classList.toggle('VerticalFullPageMap--listShown');
           this._container.classList.toggle('VerticalFullPageMap--mapShown');
+          this._pageWrapperEl.classList.toggle('YxtPage-wrapper--mapShown');
           this._container.classList.remove('VerticalFullPageMap--detailShown');
         });
       }
--- a/static/scss/answers/interactive-map/VerticalFullPageMap.scss
+++ b/static/scss/answers/interactive-map/VerticalFullPageMap.scss
@@ -175,7 +175,12 @@
       top: 0;
 
       @include bplte($mobile-break-point-max) {
-        max-height: 100vh;
+        overflow: auto;
+        position: fixed;
+        top: 0;
+        left: 0;
+        right: 0;
+        bottom: 0;
       }
     }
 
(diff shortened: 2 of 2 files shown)


creotutar:
Percy is not passing because we are getting new facets back from the backend

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -365,6 +365,7 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
           window.scrollTo(0, 0);
           this._container.classList.toggle('VerticalFullPageMap--listShown');
           this._container.classList.toggle('VerticalFullPageMap--mapShown');
+          this._pageWrapperEl.classList.toggle('YxtPage-wrapper--mapShown');
```
tmeyer2115, line 368: Is `YxtPage` a new selector prefix that was added as part of the locator work? It sounds like this class is added to the `div` that wraps the entire page. I'm wondering if we should refactor some of our existing CSS to use this prefix as well. If we have styling that applies to any page generated by the Theme, it seems like we could use this selector prefix with it. Out of scope of this item! Just an observation.

tmeyer2115 (approved):

static/js/theme-map/VerticalFullPageMapOrchestrator.js
creotutar: Definitely agree! Was here before locator work: https://github.com/yext/answers-hitchhiker-theme/blob/master/layouts/html.hbs#L153

PR #686, opened 2021-03-30
Add a `locales` flag to the `vertical` command.

tmeyer2115:
This PR adds the `locales` flag to the `vertical` command. This flag is piped
through to the `page` command that is ultimately run. If specified, localized
copies of the vertical's page will be created.

As a side note, this PR further highlights the need for Jambo to export it's
ArgumentMetadata model. We are effectively having to maintain an identical
copy of it in the Theme because we don't have access. Moving the `page` command
to the Theme would have also helped here. The method for obtaining the valid
page locales could then be extracted to a common helper.

J=SLAP-1185
TEST=manual

Tested the following:

- The output of `npx jambo describe` is correct.
- The output of `npx jambo vertical --help` is correct.
- Running the command with no locales works as expected.
- Running the command with an invalid additional locale works as expected.
- Running the command with valid, additional locales works as expected.

### DIFF
--- a/commands/addvertical.js
+++ b/commands/addvertical.js
@@ -88,6 +88,7 @@ class VerticalAdder {
   }
 
   /**
+   * @param {Object} jamboConfig The Jambo configuration of the site.
    * @returns {Array<string>} The additional locales that are configured in 
    *                          locale_config.json
    */


cea2aj:
> As a side note, this PR further highlights the need for Jambo to export it's ArgumentMetadata model.
Adding onto this, it would be great if Jambo also exported its UserError and SystemError models so that they don't have to be duplicated in the theme

commands/addvertical.js
```diff
@@ -71,11 +77,53 @@ class VerticalAdder {
...
+    const pageLocales = [];
+    const localeContentsRaw = fs.readFileSync(localeConfig, 'utf-8');
+    let localeContentsJson;
+    try {
+      localeContentsJson = parse(localeContentsRaw);
+    } catch(err) {
+      throw new UserError('Could not parse locale_config.json ', err.stack);
+    }
+    const defaultLocale = localeContentsJson.default;
+    for (const locale in localeContentsJson.localeConfig) {
```
cea2aj, line 118: nit: Can you declare `pageLocales` above the for loop since it's not needed until the loop? Alternatively, you could remove `pageLocales` and use an array filter

commands/addvertical.js
```diff
@@ -149,6 +197,15 @@ class VerticalAdder {
     if (args.cardName && !availableCards.includes(args.cardName)) {
       throw new UserError(`${args.cardName} is not a valid card`);
     }
+
+    if (args.locales.length) {
+      const supportedLocales = VerticalAdder._getAdditionalPageLocales(this.config);
+      args.locales.forEach(locale => {
+        if (!supportedLocales.includes(locale)) {
+          throw new UserError(`${locale} is not supported by your site`);
```
cea2aj, line 205: nit: I think it might be helpful to include the fact that it's a locale which isn't supported. What do you think of updating the message to `${locale} is not a locale supported by your site`?

commands/addvertical.js
```diff
@@ -71,11 +77,54 @@ class VerticalAdder {
...
         }
       }
     };
   }
 
+  /**
+   * @returns {Array<string>} The additional locales that are configured in 
+   *                          locale_config.json
+   */
+  static _getAdditionalPageLocales(jamboConfig) {
```
creotutar, line 94: nit: can we add a params jsdoc?

tmeyer2115:
> > As a side note, this PR further highlights the need for Jambo to export it's ArgumentMetadata model.
> 
> Adding onto this, it would be great if Jambo also exported its UserError and SystemError models so that they don't have to be duplicated in the theme

Yeah, it would be great to have those exported as well. I think we have an item to export all this, but it hasn't been scheduled.

cea2aj (approved):
lgtm

creotutar (approved):
🆙 🆙 🆙 LGTM

PR #692, opened 2021-04-01
Only collapse filters on searchbar searches

oshi97:
This fixes a regression where collapsible filters were
collapsing whenever any search was run, not just ones made
using the searchbar. This is done by checking for the query trigger
corresponding to the search.

This happened because of the way collapsibe filters was differentiating
between vertical results updated caused by the searchbar, and vertical
results updated made by filter components and pagination. Previously,
any update to vertical results that had a corresponding update to query
was considered to be caused by the searchbar. In SDK 1.8, this is no longer
the case, and all searches now have a corresponding query update.

J=SLAP-1204
TEST=manual

test that updating filters in collapsible filters will not collapse the filters
test that back navigating on the filters view (when the filters are expanded)
does not collapse the filters


static/js/collapsible-filters/interactions.js
```diff
@@ -123,6 +123,18 @@ export default class Interactions {
...
       storageKey: 'query',
       callback: () => {
+        const repeatedSearchTriggers = [
+          'filter-component',
+          'pagination',
+          'suggest',
+          'initialize',
+          'query-parameter'
+        ];
+        const queryTrigger = ANSWERS.core.storage.get('queryTrigger');
```
creotutar, line 133: Can we add queryTrigger to the new storage-keys.js and reference from there?

static/js/collapsible-filters/interactions.js
```diff
@@ -123,6 +123,18 @@ export default class Interactions {
       eventType: 'update',
       storageKey: 'query',
       callback: () => {
+        const repeatedSearchTriggers = [
```
tmeyer2115, line 126: What is a 'repeated' search? I'm a bit confused by that terminology.

static/js/collapsible-filters/interactions.js
```diff
@@ -123,6 +123,18 @@ export default class Interactions {
...
+          'filter-component',
+          'pagination',
+          'suggest',
+          'initialize',
+          'query-parameter'
+        ];
+        const queryTrigger = ANSWERS.core.storage.get('queryTrigger');
+        const isRepeatedSearch =  repeatedSearchTriggers.includes(queryTrigger);
```
tmeyer2115, line 134: I don't suppose we could go the reverse direction? If the `queryTrigger` is equal to a particular value, then we set `pendingQueryUpdate`. That may make the code a bit simpler. But, I recognize that trigger may not exist.

oshi97: This is actually every single query trigger. We could have it be that if there is no `queryTrigger`, `set pendingQueryUpdate` I had this originally but felt like it was a little less obvious what's happening? On the off chance we add another trigger

static/js/collapsible-filters/interactions.js
oshi97: it's any search that wasn't triggered by typing in a query into the searchbar, I can update!

static/js/collapsible-filters/interactions.js
oshi97: yes definitely, I kinda want to import these directly from the SDK src in the future (or maybe provide some export in the SDK) but that's a good middle ground!

static/js/collapsible-filters/interactions.js
oshi97: Actually I can put up a PR to add a trigger for searchbar searches? I was worried about breaking somebody's random code, but I think I'd rather do that than add a trap for us in the future

static/js/collapsible-filters/interactions.js
oshi97: oh wait I thought you meant the actual query triggers themselves, will also update the storage key

tmeyer2115 (approved):

PR #694, opened 2021-04-02
vertical-full-page-map: Fix styling for iFrame experience

creotutar:
Previously, the map would resize indefinitely when iframed. This is
because the library we use, the iFrame resizer, listens for changes to
the height of the content in the child frame. In turn, the child frame
is listening for changes to the height in its frame. This is because we
are relying on 100vh heights for the map to give a full screen effect.

This is a know bug (below) with the iframe resizing library that is used
with all of our current iframe experiences.

iframe-resizer/issues/ 692

In order to fix this, we set an explicit height for the maps on mobile
and desktop. This allows the content to not rely on its viewport height.
We set the heights to defaults that look good across many mobile
browsers and desktop browsers.

An important part in the height decision was being able to see the
Location Bias right away when you landed on a page with results.

NOTE: As a result of this, we had to change our result list from being
very long on mobile to being inner scrolled. This has a UI effect of two
scroll bars.

NOTE: Because we are optimizing for iframes, you will notice that the
mobile toggle will be a bit higher on the mobile list view than the
mobile map view in the non-iframed experience. This is so that the
height issues for map are resolved. The mobile toggle is not this way
on an iFramed experience. If you want the normal experience to have
better spaced mobile toggles, use 100vh as your mobile maps height.

J=SLAP-1201
TEST=manual

Create a ./iframe_test.html with the correct iframe div and a script tag
referencing a local iframe.js

Make sure there are JAMBO_INJECTED_DATA in your env variables set
to use localhost as a staging domain. This makes it so the iframe shows
a local iFrame.

Test on Samsung s21 / Pixel 4 / Pixel 5 Chrome
Test on iPhone 12 / iPhone 8 Plus / iPad Safari, Chrome
Test on IE11, Chrome, Firefox Desktop

That interactions still work, the difference on an iframe is that the height
is set and not necessarily the entire height of the device.

### DIFF
--- a/static/scss/answers-variables.scss
+++ b/static/scss/answers-variables.scss
@@ -63,6 +63,8 @@ $container-width: 1000px;
   --yxt-maps-desktop-results-container-width: 410px;
   --yxt-maps-mobile-results-header-height: 185px;
   --yxt-maps-mobile-results-footer-height: 97px;
+  --yxt-maps-desktop-height: 950px;
+  --yxt-maps-mobile-height: 620px;
 }
 
 // breakpoint variables for use in media queries within the theme styling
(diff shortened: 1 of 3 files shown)


oshi97 (approved):
css :godmode:

cea2aj (approved):

PR #695, opened 2021-04-02
Fix behavior when clicking away from detail cards

cea2aj:
Fixes the problem that occurs when a users clicks away from a mobile detail card

Previously, the behavior when a user closed a detail card by clicking on the map and by clicking the close card button. This PR updates the behavior so that both methods of closing the detail card results in the same behavior. This fixes the bug where the detail card would be present below the results when the users switches back to list view

J=SLAP-1205
TEST=manual

Verify that the bug is no longer present. Test that closing a card by clicking on the map and by clicking on the 'x' button results in the same behavior.

### DIFF
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -111,6 +111,12 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
      * @type {SearchDebouncer}
      */
     this.searchDebouncer = new SearchDebouncer();
+
+    /**
+     * The detail card which apears on mobile after clicking a pin
+     * @type {Element}
+     */
+    this._detailCard = null;
   }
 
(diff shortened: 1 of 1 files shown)


static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -313,6 +319,8 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
       el.classList.remove('yxt-Card--pinFocused');
     });
 
+    this._detailCard && this._detailCard.remove();
```
tmeyer2115, line 322: Should we update the name of this method? Now, the card is actually getting removed, correct? The method name suggests we're just toggling CSS classes/focus states.

cea2aj: updated!

tmeyer2115 (approved):

PR #696, opened 2021-04-05
vertical-full-page-map: Add `alternativeVerticalsConfig` passthrough

creotutar:
Previously, the recommended other alternative verticals would show as a
suggestion, even if the query was an empty string. We have opted not to
show the vertical suggestions at all.

However, this is not toggle-able by the component in the SDK. In order
to remove that HTML, we must override the template to hide the HTML.

In order to pass a translated hbs file to the Alternative Verticals child of
the Vertical Full Page Map, we add an alternative verticals config
option to the Vertical Full Page Map component. This is better than just
a specific alternative verticals template config option.

We copy the `alternativeverticals.hbs` directly from the Answers Search
UI SDK except we add an exception for no-query suggested verticals. We
also escapeHTML all the translated files and copy the translations directly
from the Answers Search UI SDK .po files.

Note: An unused config option `searchbarConfig` was replaced with
this `alternativeVerticalsConfig` option.

J=SLAP-1203
TEST=manual

Test that when you get to a no results state on the new map, you will
not see alternative verticals show for any query.

Test that on the benchmark account, if you zoom into a search this
area search with no results (and no query) you do not get alternative
verticals. However, if you type e.g. `wallet`, you will get suggested
verticals to `products` (given you added the products vertical).

### DIFF
--- a/theme-components/vertical-full-page-map-alternative-verticals/alternativeverticals.hbs
+++ b/theme-components/vertical-full-page-map-alternative-verticals/alternativeverticals.hbs
@@ -1,0 +1,8 @@
+<div class="yxt-AlternativeVerticals{{#unless isShowingResults}} yxt-AlternativeVerticals--notShowingResults{{/unless}}">
+  <div class="yxt-AlternativeVerticals-noResultsInfo">
+    {{translate phrase='<em class="yxt-AlternativeVerticals-noResultsInfo--emphasized">No results found</em> in [[currentVerticalLabel]].' currentVerticalLabel=currentVerticalLabel }}
+    {{#if isShowingResults}}
+      {{translate phrase='Showing <em class="yxt-AlternativeVerticals-noResultsInfo--emphasized">all [[currentVerticalLabel]]</em> instead.' currentVerticalLabel=currentVerticalLabel }}
+    {{/if}}
+  </div>
+</div>
--- a/theme-components/vertical-full-page-map-alternative-verticals/alternativeverticals.hbs
+++ b/theme-components/vertical-full-page-map-alternative-verticals/alternativeverticals.hbs
@@ -5,4 +5,48 @@
       {{translate phrase='Showing <em class="yxt-AlternativeVerticals-noResultsInfo--emphasized">all [[currentVerticalLabel]]</em> instead.' currentVerticalLabel=currentVerticalLabel escapeHTML=false}}
     {{/if}}
   </div>
+  {{#if (all verticalSuggestions query)}}
+    <div class="yxt-AlternativeVerticals-suggestionsWrapper">
+      <div class="yxt-AlternativeVerticals-details">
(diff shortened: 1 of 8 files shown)


theme-components/vertical-full-page-map/script.js
```diff
@@ -49,14 +49,11 @@ ANSWERS.addComponent('VerticalFullPageMapOrchestrator', Object.assign({},
       {{/if}}
     {{/each}}
   ],
-  searchbarConfig: Object.assign({},
```
tmeyer2115, line 52: For my own edification, why do we no longer need the `searchbarConfig`?

tmeyer2115 (commented):
Did Rose decide we never want to display AlternativeVerticals on the Locator? My initial read of the item was that only if the query is empty do we not want to display AlternativeVerticals.

theme-components/vertical-full-page-map/script.js
creotutar: That was a residual from when we were trying out all components as child components of the orchestrator! That is no longer the case to keep the structure HHs are used to

creotutar:
@tmeyer2115 Confirming w Product! I believe in the meeting, we decided not to special case it because it would be a higher LOE but it seems we pass query in the Alternative Verticals and we would be able to do it in this template. Will LYK.

creotutar:
@tmeyer2115 Updated to make the exception for no query suggested queries! w/ translations!

theme-components/vertical-full-page-map-alternative-verticals/alternativeverticals.hbs
```diff
@@ -0,0 +1,52 @@
...
+  <div class="yxt-AlternativeVerticals-noResultsInfo">
+    {{translate phrase='<em class="yxt-AlternativeVerticals-noResultsInfo--emphasized">No results found</em> in [[currentVerticalLabel]].' currentVerticalLabel=currentVerticalLabel escapeHTML=false}}
+    {{#if isShowingResults}}
+      {{translate phrase='Showing <em class="yxt-AlternativeVerticals-noResultsInfo--emphasized">all [[currentVerticalLabel]]</em> instead.' currentVerticalLabel=currentVerticalLabel escapeHTML=false}}
+    {{/if}}
+  </div>
+  {{#if (all verticalSuggestions query)}}
+    <div class="yxt-AlternativeVerticals-suggestionsWrapper">
+      <div class="yxt-AlternativeVerticals-details">
+        {{translate phrase='The following search category yielded results for <span class="yxt-AlternativeVerticals-details--query">"[[query]]"</span>:' pluralForm='The following search categories yielded results for <span class="yxt-AlternativeVerticals-details--query">"[[query]]"</span>:' count=verticalSuggestions.length query=query escapeHTML=false}}
```
oshi97, line 11: is it possible to have break up these translate helpers into multiple lines? (also below)

creotutar: @oshi97 Unfortunately, I don't think this is supported in the theme yet! (See SLAP-1178)

oshi97: ah ok got it

oshi97 (approved):

tmeyer2115 (approved):

PR #700, opened 2021-04-05
Fix logic for exiting the active state of a pin

cea2aj:
Make it such that the active pin is deactivated if a user selects the search bar, or if a new search is ran

J=SLAP-1208
TEST=manual

Test that clicking on the search bar closes the detail card on mobile. Test that the detail card closes during "Search When Map Moves", and after pressing the "Search This Area" button. Test that the detail card closes when tabbing onto the search bar.

### DIFF
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -237,7 +237,8 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
       zoomChangedListener: zoomChangedListener,
       zoomEndListener: zoomEndListener,
       panHandler: panHandler,
-      canvasClickListener: () => this.deselectAllResults()
+      canvasClickListener: () => this.deselectAllResults(),
+      searchBarFocusListener: () => this.deselectAllResults()
     }));
   }
 
@@ -439,6 +440,8 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
       StorageKeys.LOCATOR_NUM_CONCURRENT_SEARCH_THIS_AREA_CALLS,
       updatedNumSearchThisAreaCalls
     );
+
+    this.deselectAllResults();
 
     this._container.classList.remove('VerticalFullPageMap--showSearchThisArea');
 
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -428,13 +428,21 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
 
   /**
    * Register listeners so that any active pins are deselected when a user clicks
-   * on the search bar
+   * or focuses on the searchbar.
    */
   setupSearchBarListeners () {
     const searchBarForm = this._container.querySelector('.yxt-SearchBar-form');
     searchBarForm && searchBarForm.addEventListener('click', () => {
       this.deselectAllResults()
     });
+    const searchBarInput = this._container.querySelector('.yxt-SearchBar-input');
+    searchBarInput && searchBarInput.addEventListener('focus', () => {
+      this.deselectAllResults();
+    });
+    const searchBarButton = this._container.querySelector('.yxt-SearchBar-button');
+    searchBarButton && searchBarButton.addEventListener('focus', () => {
+      this.deselectAllResults();
+    })
   }
 
   /**
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -431,8 +431,8 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
    * on the search bar
    */
   setupSearchBarListeners () {
-    const searchBarInput = this._container.querySelector('.yxt-SearchBar-form');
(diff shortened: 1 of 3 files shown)


static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -410,6 +411,20 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
     }
   }
 
+  /**
+   * Register listeners so that any active pins are deselected when a user clicks
+   * or focuses on the search bar
+   */
+  setupSearchBarListeners () {
+    const searchBarInput = document.querySelector('#yxt-SearchBar-input--SearchBar');
+    searchBarInput && searchBarInput.addEventListener('click', () => {
```
creotutar, line 420: Does this account for clicking on the searchbar button to search?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -410,6 +411,20 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
     }
   }
 
+  /**
+   * Register listeners so that any active pins are deselected when a user clicks
+   * or focuses on the search bar
+   */
+  setupSearchBarListeners () {
+    const searchBarInput = document.querySelector('#yxt-SearchBar-input--SearchBar');
```
creotutar, line 419: Can we use `this._container` instead? I think we want the Orchestrator scoped to elements inside its specified container

creotutar: Do we want this selector to be a configuration option? It seems this selector could change if a different `name` is given in the search bar options

static/js/theme-map/VerticalFullPageMapOrchestrator.js
cea2aj: It doesn't, however new searches deselect all results, so the behavior will still occur

static/js/theme-map/VerticalFullPageMapOrchestrator.js
cea2aj: Ya, i'll try this._container. Also, I think there's another CSS class I can target instead which won't be susceptible to name changes

cea2aj:
> It seems like we have `deselectAllResults` scattered throughout the code to account for the various scenarios. Could we consolidate this logic? Specifically, I'm wondering it we could add a single listener to `storage`. That listener would listen to changes in `StorageKeys.VERTICAL_RESULTS`. If the "search loading" state is entered, we invoke `deselectAllResults`.

I can investigate how far that will get us. I don't think it will work for deselecting all results when a user clicks on a map, however that might be an acceptable trade off

cea2aj:
> > It seems like we have `deselectAllResults` scattered throughout the code to account for the various scenarios. Could we consolidate this logic? Specifically, I'm wondering it we could add a single listener to `storage`. That listener would listen to changes in `StorageKeys.VERTICAL_RESULTS`. If the "search loading" state is entered, we invoke `deselectAllResults`.
> 
> I can investigate how far that will get us. I don't think it will work for deselecting all results when a user clicks on a map, however that might be an acceptable trade off

Update:  Most of the times when we want to call `deselectAlllResults`, it isn't associated with the vertical results search loading state, so that wouldn't get us the behavior that we're looking for

tmeyer2115:
> > > It seems like we have `deselectAllResults` scattered throughout the code to account for the various scenarios. Could we consolidate this logic? Specifically, I'm wondering it we could add a single listener to `storage`. That listener would listen to changes in `StorageKeys.VERTICAL_RESULTS`. If the "search loading" state is entered, we invoke `deselectAllResults`.
> > 
> > 
> > I can investigate how far that will get us. I don't think it will work for deselecting all results when a user clicks on a map, however that might be an acceptable trade off
> 
> Update: Usually when we want to call `deselectAlllResults`, it isn't associated with the vertical results search loading state, so that wouldn't get us the behavior that we're looking for

Gotcha. Thanks for investigating!

tmeyer2115 (approved):

static/js/theme-map/VerticalFullPageMapOrchestrator.js
cea2aj: updated to account for clicking on the search bar button

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -425,6 +426,17 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
     }
   }
 
+  /**
+   * Register listeners so that any active pins are deselected when a user clicks
+   * on the search bar
+   */
+  setupSearchBarListeners () {
+    const searchBarForm = this._container.querySelector('.yxt-SearchBar-form');
+    searchBarForm && searchBarForm.addEventListener('click', () => {
```
tmeyer2115, line 435: I notice we no longer have a `focus` listener in this PR. Does that have any WCAG implications?

cea2aj: I removed it because
- It's not necessary for us to have it to be WCAG compliant, and its a rare use case where someone is on mobile and using keyboard
- I felt that it would make the code a little more confusing since I would have to add focus listeners to two additional elements

However I can support it if you'd like

cea2aj: Updated to include focus listeners!

tmeyer2115 (approved):

PR #703, opened 2021-04-06
Map list view fix for tall mobile screens

cea2aj:
Fix a bug where the map would extend below the results on mobile.

This bug occurred on the list view page only within the mobile
breakpoint, and only on non-iframed experiences.

Fix the bug by adding a new YxtPage-wrapper--mobileListView CSS
class which lets us apply a fix to the Answers-footer. We fix the footer
by removing the margin using padding instead so that the white
background of the footer covers the map.

This PR updates the map orchestrator to improve the handling of the
mobileListView and the mobileMapView CSS classes. These changes
ensure that classes associated with mobile views only exist when the
page is within the mobile breakpoint. This has the side effect of fixing
layout issues when someone drags the screen between mobile and
desktop sizes.

J=SLAP-1230
TEST=manual

Test very tall screen sizes within the mobile breakpoint on list view
and confirm that the bug is fixed. Test on iframe and non-iframe
experiences to make sure things work as expected. Observe the
listShown CSS class being properly set on the page. Ensure that
a YxtFooter can be put on the page and have it look as expected.

### DIFF
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -6,6 +6,7 @@ import { defaultCenterCoordinate } from './constants.js';
 
 import ZoomTriggers from './Maps/ZoomTriggers.js';
 import PanTriggers from './Maps/PanTriggers.js';
+import MobileViews from './MobileViews';
 
 import StorageKeys from '../constants/storage-keys.js';
 
@@ -107,6 +108,11 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
     this.mobileBreakpointMax = 991;
 
     /**
+     * The current view for mobile
+     */
+    this._mobileView = MobileViews.LIST_VIEW;
+
+    /**
      * Determines whether or not another search should be ran
      * @type {SearchDebouncer}
      */
@@ -159,11 +165,87 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
       this.searchThisArea();
     });
 
-    this._pageWrapperEl.classList.add('YxtPage-wrapper--listShown');
-
+    this.setupMapViewCssClasses();
     this.addMapComponent();
   }
 
+  /**
+   * Properly set CSS classes for mobile and desktop
+   */
+  setupMapViewCssClasses () {
+    if (!this.isMobile()) {
+      this.updateCssForDesktop();
+    }
+    window.addEventListener('resize', () => {
+      if (this.isMobile()) {
+        this.updateCssForMobile();
+      } else {
+        this.updateCssForDesktop();
+      }
+    }, { passive: true });
+  }
+
+  isMobile () {
+    return window.innerWidth <= this.mobileBreakpointMax;
+  }
+
+  updateCssForMobile () {
+    if (this._mobileView === MobileViews.LIST_VIEW) {
+      this.addMobileListViewCssClasses();
+      this.removeMobileMapViewCssClasses();
+      this.removeMobileDetailShownCssClasses();
+    } else if (this._mobileView === MobileViews.MAP_VIEW) {
+      this.addMobileMapViewCssClasses();
+      this.removeMobileListViewCssClasses();
+    }
+  }
+
+  updateCssForDesktop () {
+    this.removeMobileListViewCssClasses();
+    this.removeMobileMapViewCssClasses();
+    this.removeMobileDetailShownCssClasses();
+  }
+
+  setMobileMapView () {
+    this._mobileView = 'mapView';
+    this.updateCssForMobile();
+  }
+
+  setMobileListView () {
+    this._mobileView = 'listView';
+    this.updateCssForMobile();
+  }
+
+  addMobileListViewCssClasses () {
+    this._container.classList.add('VerticalFullPageMap--mobileListView');
+    this._pageWrapperEl.classList.add('YxtPage-wrapper--mobileListView');
+  }
+
+  addMobileMapViewCssClasses () {
+    this._container.classList.add('VerticalFullPageMap--mobileMapView');
+    this._pageWrapperEl.classList.add('YxtPage-wrapper--mobileMapView');
+  }
+
+  addMobileDetailShownCssClasses () {
+    this._container.classList.add('VerticalFullPageMap--detailShown');
+    this._pageWrapperEl.classList.add('YxtPage-wrapper--detailShown');
+  }
+
+  removeMobileMapViewCssClasses () {
+    this._container.classList.remove('VerticalFullPageMap--mobileMapView');
+    this._pageWrapperEl.classList.remove('YxtPage-wrapper--mobileMapView');
+  }
+
+  removeMobileListViewCssClasses () {
+    this._container.classList.remove('VerticalFullPageMap--mobileListView');
+    this._pageWrapperEl.classList.remove('YxtPage-wrapper--mobileListView');
+  }
+
+  removeMobileDetailShownCssClasses () {
+    this._container.classList.remove('VerticalFullPageMap--detailShown');
+    this._pageWrapperEl.classList.remove('YxtPage-wrapper--detailShown');
+  }
+  
   addMapComponent () {
     /**
      * Sets up mobile view toggles and search bar listeners
@@ -331,8 +413,7 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
    * updating global storage.
    */
   deselectAllResults () {
-    this._container.classList.remove('VerticalFullPageMap--detailShown');
-    this._pageWrapperEl.classList.remove('YxtPage-wrapper--detailShown');
+    this.removeMobileDetailShownCssClasses();
 
     document.querySelectorAll('.yxt-Card--pinFocused').forEach((el) => {
       el.classList.remove('yxt-Card--pinFocused');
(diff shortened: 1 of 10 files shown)


static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -159,6 +159,8 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
       this.searchThisArea();
     });
 
+    this._pageWrapperEl.classList.add('YxtPage-wrapper--listShown');
```
creotutar, line 162: Instead of adding it through JavaScript on initial page load. Do we want to add this CSS class to the page directly on vertical-full-page-map pages?

cea2aj: That would be ideal, however YxtPage-wrapper is defined inside layotus/html.hbs, and this seems like a better solution than conditionally adding YxtPage-wrapper--listShown to layouts/html.hbs

creotutar: My worry is that this can cause a flash of CSS on page load, where you might see the map and then it hides away after a second. Is your worry about adding some `pageType` parameter to the >layouts/html partial?

cea2aj: The `pageType` parameter idea actually sounds like a good solution. I was more concerned about putting the 'VerticalFullPageMap' string somewhere in layouts/html.hbs, but we could avoid that with a param. I'll try that.

I tested it though, and no page flashing occurs because the CSS class is added to the page before the map component is added

static/scss/answers/interactive-map/VerticalFullPageMap.scss
```diff
@@ -1,3 +1,5 @@
+$mobile-break-point-max: sm !default;
```
creotutar, line 1: Hm, this makes the SCSS variable global to any file that imports this file (e.g. `_default.scss`) instead of scoped to the VerticalFullPageMap. How do we feel about removing the uses of `$mobile-break-point-max` and `$desktop-break-point-min` in this file and instead using some variable in `answers-variables`? Maybe we could make a new one for vertical-full-page-maps?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
cea2aj: Updated so that YxtPage-wrapper-mobileListView is default, however if the page loads on desktop, we remove the CSS class. This causes no visible CSS flashing.

static/scss/answers/interactive-map/VerticalFullPageMap.scss
cea2aj: I'm not sure about exposing this an answers-variables, because we have the mobile breakpoint hardcoded if a few places, and we didn't really design the map to support different breakpoints. Due to a refactor, this change is no longer necessary, however the ability to change the breakpoint might make sense in a future enhancement

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -407,18 +487,20 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
       if (!listToggle.dataset.listened) {
         listToggle.dataset.listened = 'true';
         listToggle.addEventListener('click', () => {
+          this.deselectAllResults();
           window.scrollTo(0, 0);
-          this._container.classList.toggle('VerticalFullPageMap--listShown');
-          this._container.classList.toggle('VerticalFullPageMap--mapShown');
-          this._pageWrapperEl.classList.toggle('YxtPage-wrapper--mapShown');
-          this._container.classList.remove('VerticalFullPageMap--detailShown');
+          if (this._mobileView === 'listView') {
```
creotutar, line 499: LOVE the refactor. Can we use the enum value here?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -159,9 +165,87 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
   }
 
+  /**
+   * Properly set CSS classes for mobile and desktop
+   */
+  setupMapViewCssClasses () {
+    if (!this.isMobile()) {
+      this.updateCssForDesktop();
+    }
+    window.addEventListener('resize', () => {
```
creotutar, line 179: I'm worried about this event being called a lot every time you resize. On Consulting we had a throttle based on animation frames for resize events to prevent it from being called too many times in one resize. Is that something we can do here?

creotutar: We may also be interested in the `addListener` for the media query object (so we can use both matches in the `isMobile` function and `addListener` for listening to breakpoint changes)

https://blog.logrocket.com/introduction-to-js-media-queries/

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -159,9 +165,87 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
+      if (this.isMobile()) {
+        this.updateCssForMobile();
+      } else {
+        this.updateCssForDesktop();
+      }
+    }, { passive: true });
+  }
+
+  isMobile () {
```
creotutar, line 188: can we add jsdoc for types?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -159,9 +165,87 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
+    this._mobileView = 'mapView';
+    this.updateCssForMobile();
+  }
+
+  setMobileListView () {
+    this._mobileView = 'listView';
+    this.updateCssForMobile();
+  }
+
+  addMobileListViewCssClasses () {
```
creotutar, line 219: How do you feel about some
```
applyCssClassesForState(state) {
    let classModifier = getModifierForState(state)
    this._container.classList.add('VerticalFullPageMap--${classModifier}');
    this._pageWrapperEl.classList.add('YxtPage-wrapper--${classModifier}');
}
```

where state -> css class modifier can come from your MobileViews.js enum?

cea2aj: sure

cea2aj: So I renamed MobileViews to MobileStates, so that we could include detailShown in there. So all three are states, however only map view and list view are views.

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -159,9 +167,99 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
+   * @param {MobileStates} mobileState 
+   * @returns {string}
+   */
+  getModifierForState(mobileState) {
+    switch (mobileState) {
+      case MobileStates.LIST_VIEW:
+        return 'mobileListView'
+      case MobileStates.MAP_VIEW:
+        return 'mobileMapView'
+      case MobileStates.DETAIL_SHOWN:
```
creotutar, line 258: nit Can we have a default case?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -159,9 +167,99 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
+    const classModifier = this.getModifierForState(mobileState)
+    this._container.classList.add(`VerticalFullPageMap--${classModifier}`);
+    this._pageWrapperEl.classList.add(`YxtPage-wrapper--${classModifier}`);
+  }
+
+  /**
+   * @param {MobileStates} mobileState 
+   */
+  removeCssClassesForState(mobileState) {
+    const classModifier = this.getModifierForState(mobileState)
```
creotutar, line 241: nit: semicolon please

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -159,9 +167,99 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
+  /**
+   * Properly set CSS classes for mobile and desktop
+   */
+  setupCssForBreakpoints () {
+    if (!this.isMobile()) {
+      this.updateCssForDesktop();
+    }
+
+    const mediaQuery = window.matchMedia(`(max-width: ${this.mobileBreakpointMax}px)`);
+    mediaQuery.addEventListener('change', () => {
```
creotutar, line 183: Does this listener persist outside of this function, given the variable is defined within this scope? Do we want to move this to a class variable so it can be used in both here and `isMobile`?

cea2aj: updated to put the media query in the class

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -257,6 +261,8 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
         return 'mobileMapView'
       case MobileStates.DETAIL_SHOWN:
         return 'mobileDetailShown'
+      default:
```
creotutar, line 264: This feels like a default we'd want to surface an error for or check for in the add/remove functions. Otherwise we might end up with `VerticalFullPageMap--` surfacing on the DOM. What do you think?

cea2aj: getting to the default here indicates a problem with our code, so I think throwing an error is reasonable

creotutar (approved):
📵 📵 📵 LGTM

PR #704, opened 2021-04-06
update "search this area" button to be visible in no results

oshi97:
This commit makes the "search this area" button visible
during the no results state. Before this commit, the no
results block would cover up "search this area".

This was done by moving "search this area" directly
below Answers-resultsWrapper in the html.
In order for this to work, the css for "search this area"
was changed to use static positioning on mobile.

getBoundingClientRect() was used to calculate the new
margin-top for Answers-centerTop

J=SLAP-1225
TEST=manual

open up a side by side of a maps page before and after this commit
see that on mobile and desktop, the "search this area" button is in
the same position
see that I can swap back and forth between list and map view on mobile
see that in the no results state, the "search this area" button is directly
below the no results block, instead of being covered up by it

### DIFF
--- a/static/scss/answers/interactive-map/VerticalFullPageMap.scss
+++ b/static/scss/answers/interactive-map/VerticalFullPageMap.scss
@@ -245,13 +245,14 @@
     }
 
     &-centerTop {
-      position: absolute;
+      display: flex;
+      margin-top: 11px;
       color: black;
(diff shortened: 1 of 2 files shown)


oshi97:
image

oshi97:
double checked with rose and she said it's ok that the zoom controls are hidden in no results

creotutar (approved):
👍 👍 👍 LGTM

PR #707, opened 2021-04-06
dont duplicate vertical-full-page-map page-setup

oshi97:
This commit slightly reduces the amount of code
in the full page map's setup by reusing a function
instead of having two copies of the same js partials

J=none
TEST=manual

see that ANSWERS.registerTemplate("theme-components/theme-map only occurs
once in the final build's html, and that before this change it would be in the
html twice
smoke test the map


creotutar (approved):
👍 👍 👍 LGTM

PR #709, opened 2021-04-07
dynamically resize the map on no results

oshi97:
This commit dynamically resizes the map when the no results
state appears/disappears. This is because the no results info
that pops up would otherwise cover up part of the map. This
would be a problem primarily if somebody had displayAllResults
set to true for the map, in which case a majority of the pins
would be blocked by alternative verticals suggestions.

J=SLAP-1237
TEST=manual

on no results, the map dynamically resizes
refreshing the page while on no results goes to the list
view, and the map is not visible
desktop looks correct for no results and regular results
"search this area" button is visible in no results

### DIFF
--- a/static/js/collapsible-filters/interactions.js
+++ b/static/js/collapsible-filters/interactions.js
@@ -214,8 +214,7 @@ export default class Interactions {
     if (!this._disableScrollToTopOnToggle) {
       this.scrollToTop();
     }
-    const filterLink = ANSWERS.components.getActiveComponent('FilterLink');
-    filterLink && filterLink.setState({
+    ANSWERS.components.getActiveComponent('FilterLink').setState({
       panelIsDisplayed: !shouldCollapseFilters
     });
   }
--- a/static/js/collapsible-filters/interactions.js
+++ b/static/js/collapsible-filters/interactions.js
@@ -214,7 +214,8 @@ export default class Interactions {
     if (!this._disableScrollToTopOnToggle) {
       this.scrollToTop();
     }
-    ANSWERS.components.getActiveComponent('FilterLink').setState({
+    const filterLink = ANSWERS.components.getActiveComponent('FilterLink');
+    filterLink && filterLink.setState({
       panelIsDisplayed: !shouldCollapseFilters
     });
   }
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -505,7 +505,7 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
       return;
     }
     const noResults = this._noResultsContainer;
-    const map = document.querySelector(this._mapContainerSelector);
+    const map = this._container.querySelector(this._mapContainerSelector);
     if (!noResults || !map) {
       return;
     }
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -514,9 +514,10 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
       const noResultsHeight = noResults.getBoundingClientRect().height;
       map.style.top =
         `calc(var(--yxt-maps-mobile-results-header-height) + ${noResultsHeight}px)`;
-      const headerFooterHeight = 
-        'var(--yxt-maps-mobile-results-header-height) - var(--yxt-maps-mobile-results-footer-height)';
-      map.style.height = `calc(100% - ${headerFooterHeight} - ${noResultsHeight}px)`;
+      map.style.height = 'calc(100%' +
+      ' - var(--yxt-maps-mobile-results-header-height)' + 
+      ' - var(--yxt-maps-mobile-results-footer-height) ' + 
+      ` - ${noResultsHeight}px)`;
     } else {
       map.style.top = '';
       map.style.height = '';
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -21,12 +21,13 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
 
     /**
      * The container in the DOM for the interactive map
-     * @type {HTMLElement}
+     * @type {string}
      */
     this._mapContainerSelector = '#js-answersMap';
 
     /**
      * The container in the DOM for the no results block.
+     * @type {HTMLElement}
      */
     this._noResultsContainer = document.querySelector('.js-answersNoResults');
 
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -8,6 +8,7 @@ import ZoomTriggers from './Maps/ZoomTriggers.js';
 import PanTriggers from './Maps/PanTriggers.js';
 
 import StorageKeys from '../constants/storage-keys.js';
+import ResultsContext from '../constants/results-context.js';
 
 /**
  * The component to control the interactions for an interative map.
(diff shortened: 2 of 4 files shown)


static/js/collapsible-filters/interactions.js
```diff
@@ -214,7 +214,8 @@ export default class Interactions {
     if (!this._disableScrollToTopOnToggle) {
       this.scrollToTop();
     }
-    ANSWERS.components.getActiveComponent('FilterLink').setState({
+    const filterLink = ANSWERS.components.getActiveComponent('FilterLink');
```
creotutar, line 217: Is this change related?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -478,14 +494,42 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
+
   setState(data) {
     if (data.searchState === 'search-loading') {
       return;
     }
 
     this._data = data;
 
-    if (data.resultsContext === 'no-results') {
+    if (data.resultsContext === ResultsContext.NO_RESULTS) {
```
creotutar, line 532: I love the enum

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -127,6 +133,12 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
      * @type {Object}
      */
     this.alternativeVerticalsConfig = config.alternativeVerticalsConfig;
+
+    /**
+     * Whether the page is displaying the map on mobile.
+     * @type {boolean}
+     */
+    this.isInMobileMapView = false;
```
creotutar, line 141: This is going to have a conflict with #703 
We may want to update to use the MapStates enum

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -24,6 +25,11 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
      */
     this._mapContainerSelector = '#js-answersMap';
 
+    /**
+     * The container in the DOM for the no results block.
+     */
+    this._noResultsContainer = document.querySelector('.js-answersNoResults');
```
creotutar, line 31: Can we query `this._container` instead of `document` to keep it scoped to this component?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -24,6 +25,11 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
      */
     this._mapContainerSelector = '#js-answersMap';
 
+    /**
+     * The container in the DOM for the no results block.
```
creotutar, line 29: nit: type jsdoc please

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -478,14 +494,42 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
+    const noResults = this._noResultsContainer;
+    const map = document.querySelector(this._mapContainerSelector);
+    if (!noResults || !map) {
+      return;
+    }
+    if (resultsContext === ResultsContext.NO_RESULTS) {
+      const noResultsHeight = noResults.getBoundingClientRect().height;
+      map.style.top =
+        `calc(var(--yxt-maps-mobile-results-header-height) + ${noResultsHeight}px)`;
+      const headerFooterHeight =
```
creotutar, line 516: I think it's a bit confusing because I'd expect it to be an addition but because it's a string interpolation it has to be a subtraction. Was this moved out because the line was too long? I am of the camp that long strings are fine, what do you think?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -478,14 +494,42 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
+      return;
+    }
+    const noResults = this._noResultsContainer;
+    const map = document.querySelector(this._mapContainerSelector);
+    if (!noResults || !map) {
+      return;
+    }
+    if (resultsContext === ResultsContext.NO_RESULTS) {
+      const noResultsHeight = noResults.getBoundingClientRect().height;
+      map.style.top =
```
creotutar, line 514: I'm a bit worried about having CSS directly applied from the JS. It makes it hard to pinpoint where styling is coming from and haves us worried about two different places for a page's styling when making changes. However I don't see another way because

(1) flex-grow cannot be used as the `mapWrapper` is not sibling to the results header and results footer
(2) the map should be fixed to contain the height on mobile phones

No change request, just voicing the concern

static/js/theme-map/VerticalFullPageMapOrchestrator.js
oshi97: oh ok I'll wait for 703 to be merged in

oshi97: then rebase

static/js/theme-map/VerticalFullPageMapOrchestrator.js
oshi97: yes love that!

static/js/collapsible-filters/interactions.js
oshi97: I was occasionally running into issues with this component not existing (not sure why it would ever not exist, could be related to the map code running before collapsible filters is ready)

oshi97: I'll make it a separate pr

static/js/theme-map/VerticalFullPageMapOrchestrator.js
oshi97: hmm, I can split the string up into multiple lines. I'm fine with long strings but not long lines

oshi97: maybe I would have a different opinion if my monitor was bigger

static/js/theme-map/VerticalFullPageMapOrchestrator.js
cea2aj: 703 has been merged, so you should be good to rebase

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -340,6 +353,11 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
     this.core.storage.set(StorageKeys.LOCATOR_SELECTED_RESULT, null);
   }
 
+  isMobile() {
```
cea2aj, line 356: 703 has this exact code, haha

oshi97: 🧠

static/js/theme-map/VerticalFullPageMapOrchestrator.js
cea2aj: Would it be possible to toggle a css class as usual, and include some new variable, say `--yxt-no-results-height`, which we update from the javascript when we toggle the no results css class?

cea2aj: This is what i'm thinking: https://css-tricks.com/updating-a-css-variable-with-javascript/

oshi97: oh wow I like that idea a lot, I was trying out a different way of setting the height from what I have in the PR and was running into occasional desync issues (with the update method running before some other random part of the page was done) but I think setting a variable might fix it

static/js/theme-map/VerticalFullPageMapOrchestrator.js
oshi97: just realized 703 is an area code

oshi97:
talked with @rosiegrant and @creotutar and we decided to not resize the map for no results! future improvement might be to add a close button to the no results pop up (like google maps)

PR #711, opened 2021-04-07
vertical-full-page-map: Move default VerticalResults noResults

creotutar:
We move from the page-config.json to the script.hbs. This was a request
from Product to not expose this default "hacky" behavior in the JSON.

For context: We must do this because the AlternativeVerticals cannot
live as a standalone component with the current SDK. We need AVs to show
on both the list view and the map view. Thus we make our outside of the
Vertical Results and empty the template for the Vertical Results.

NOTE: If you change the template in the JSON config, you will
effectively show two AVs on the page. This is a detail that Product is
going to show in documentation of how to edit No Results for the map
page.

J=None
TEST=manual

Test that the JSON config always supersedes the default noResults in
script.hbs.

Test that when noResults isn't in the JSON config, we default to the
empty template.

Test creating a new vertical and by default only one AV component shows.

### DIFF
--- a/templates/vertical-full-page-map/page-config.json
+++ b/templates/vertical-full-page-map/page-config.json
@@ -35,8 +35,7 @@
     },
     "VerticalResults": {
       "noResults": {
-        "displayAllResults": false, // Optional, whether to display all results in the vertical when no results are found.
-        "template": "<span></span>"
+        "displayAllResults": false // Optional, whether to display all results in the vertical when no results are found.
       },
       "hideResultsHeader": true
     },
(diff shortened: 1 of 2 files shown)


cea2aj:
> NOTE: If you change the template in the JSON config, you will effectively show two AVs on the page.
 Why is this?

creotutar:
>>NOTE: If you change the template in the JSON config, you will effectively show two AVs on the page.
>
>Why is this?

This is because we want to have the JSON config supersede the `script.hbs`. This is also because we want to clear out the default Alternative Verticals that lives in the Vertical Results so we can show our VerticalFullPageMapOrchestrator Alternative Verticals instead.

If you were to change the VerticalResults noResults template in the JSON config, it would clear the `<span></span>` that we have as default.

The reason we have two AVs is because of an SDK limitation, where (1) AV cannot live as a standalone component and (2) VerticalResults has no toggle option to remove an AV

cea2aj (approved):

PR #714, opened 2021-04-08
Fix long titles on vertical full page map

cea2aj:
Fix a bug where long titles overflowed into the card distance and the close card button.

Using relative rather than absolute causes long texts to wrap rather than overflow into the location distance. I adjusted it by 1px so that the distance to the top of the card is the same as before the change.

On mobile detail view, this causes the card body to be pushed down slightly. I spoke with product and they were okay with the extra spacing.

J=none
TEST=manual

Test this on the turtle head tacos site were this was originally reported, and test this on the theme test site. Confirm that long titles wrap rather than overflow into the distance text. Test the location standard, professional location card, and the financial professional location cards.

### DIFF
--- a/static/scss/answers/interactive-map/VerticalFullPageMap.scss
+++ b/static/scss/answers/interactive-map/VerticalFullPageMap.scss
@@ -562,9 +562,8 @@
     }
 
     &-closeCardButtonWrapper {
-      position: absolute;
-      right: 16px;
-      top: 16px;
+      position: relative;
+      top: 2px;
     }
 
     &-closeCardButton {
--- a/static/scss/answers/interactive-map/VerticalFullPageMap.scss
+++ b/static/scss/answers/interactive-map/VerticalFullPageMap.scss
@@ -564,6 +564,7 @@
     &-closeCardButtonWrapper {
       position: relative;
       top: 2px;
+      margin-left: auto;
     }
 
     &-closeCardButton {
--- a/static/scss/answers/interactive-map/VerticalFullPageMap.scss
+++ b/static/scss/answers/interactive-map/VerticalFullPageMap.scss
@@ -563,7 +563,7 @@
 
     &-closeCardButtonWrapper {
       position: relative;
-      top: 2px;
+      top: 1px;
       margin-left: auto;
     }
 


static/scss/answers/interactive-map/VerticalFullPageMap.scss
```diff
@@ -562,9 +562,8 @@
     }
```
oshi97, line 562: would something like changing `.HitchhikerLocationStandard-distance` to be 
```css
position: absolute;
right: 0px;
```
 fix the card body being lower?

oshi97: ^there might be some weirdness with long distances paired with long titles

cea2aj: Ya, that would fix the card body being lower, however that still makes it susceptible to being overlapped by long titles or lots of text in the card body

static/scss/answers/interactive-map/VerticalFullPageMap.scss
```diff
@@ -562,9 +562,8 @@
...
 
     &-closeCardButtonWrapper {
```
tmeyer2115, line 564: Instead of changing the close card button, could we use target the styling for the title itself? Potentially, could we use `overflow-wrap: break-word` or something like that? It feels a bit strange that to fix the title overflow issue we're altering the styling of things around the title.

cea2aj: The root problem is that the closeCardButtonWrapper is using `position: absolute` which takes the div out of the document's normal flow, so the adjacent `ordinalAndTitle` div will expand right into it. Even if we set `overflow-wrap: break-word`, the title will overflow because the page will act as if the `closeCardButtonWrapper` div isn't even there.

static/scss/answers/interactive-map/VerticalFullPageMap.scss
oshi97: ah got it

static/scss/answers/interactive-map/VerticalFullPageMap.scss
```diff
@@ -562,9 +562,8 @@
...
-      position: absolute;
```
oshi97, line 565: is there some way we could add a max-width HitchhikerLocationStandard-title instead? if not I'm happy with the current solution

static/scss/answers/interactive-map/VerticalFullPageMap.scss
tmeyer2115: Got it!

static/scss/answers/interactive-map/VerticalFullPageMap.scss
cea2aj: Ya, we could add a max width, however we would need to account for the size of the distance text which can vary. It will change based on the distance and it would change if someone customized the font or changed the text size. 

We would also need to modify the size of the card content because that is also at risk of overflowing. That'a a lot trickier because we don't have guarantees about the layout of the content (it will change based on the location card). We could add a mix width to the whole content div as well, however that would prevent the content from filling the whole card.

oshi97: does changing the closeCardButton to position: relative take care of the card content overflowing? I don't seem to be able to get the card content to overflow but I could be gumming something up.

I was thinking realistically the distance text will only be at most something like "20,000km" (7 characters + comma), since the earth's circumference is 40,000km (maybe slightly longer if some locale's number format has extra commas/spaces. Do you think something like a max width of 75% on the title and 25% on the closeCardWrapper could work? Or maybe a fixed max-width on the closeCardWrapper and a `100% - {width}` on the title? I know I'm being kind of persistent about this, but I think most of the time the title won't overflow the card. If possible I'd rather optimize our default css for the title not overflowing, than vice versa

cea2aj: Setting closeCardButton to position: relative didn't seem to take care of the card content overflowing:

Screen Shot 2021-04-12 at 11 50 19 AM

I tried out your suggested title fixes, however the card body can still overflow if there is a long address or a large body of text in the description right at the top of the card. Rose said that she'd rather have some extra spacing than having the text overflow. And she mentioned that a lot of healthcare entities have really long titles, so long titles do occur pretty frequently.

I think that if we want to avoid the extra spacing, we should change our layout so that the location distance is to the left of the close card button. However, Rose didn't want to make such a layout change at this point in time.

While testing out the other cards, I noticed that I need to make a small flexbox change, so I will update that now

oshi97 (approved):
lgtm!

static/scss/answers/interactive-map/VerticalFullPageMap.scss
oshi97: ah ok, it sucks we can't get that extra real estate on mobile but it sounds like we kind of have an impossible (or very difficult) css problem

cea2aj: Yep, I think in the future we want to be careful with `position: absolute` because it can cause these sorts of overflow issues. I would also like to fix the spacing, and I think that moving the distance to the left of the close button is the right way to do it. I'll make an item for it

PR #716, opened 2021-04-09
Take percy snapshots of the vertical full page map

cea2aj:
Take snapshots of the vertical full page map

In mobile, we want to take snapshots of the map view, list view, and the card detail view. For the map view and card detail view, I took those snapshots only at the mobile width because snapshots of those at desktop width are unnecessary.

I experimented with enabling Javascript on Percy, however we don't need Javascript enabled to get good snapshots of the  layout, so overall I don't think there is a need for it. 

J=SLAP-1163
TEST=visual

Inspect the generated snapshots in Percy.


static/js/theme-map/Maps/Providers/Mapbox.js
```diff
@@ -101,6 +101,7 @@ class MapboxPin extends ProviderPin {
     this._pinEl.style.position = 'absolute';
     this._pinEl.style.top = '0';
     this._pinEl.style.left = '0';
+    this._pinEl.classList.add('Answers-mapPin');
```
cea2aj, line 104: Do you all think it's okay to just update the pin CSS class for MapBox? It doesn't look like Google Maps supports custom CSS classes on its pins, and I don't think those pins are intended to be customizable with CSS either.

oshi97: is it possible to select this pin without adding a css class? if adding a css class makes it a lot simpler this lgtm!

tests/percy/snapshots.js
```diff
@@ -51,4 +53,18 @@ async function captureVerticalMapSearch (page, percySnapshot) {
   await page.goto(`${TEST_SITE}/locations?query=a`);
   await waitTillHTMLRendered(page)
   await percySnapshot('vertical-map-search');
+}
+
+async function captureVerticalFullPageMapSearch (page, percySnapshot) {
+  await page.goto(`${TEST_SITE}/locations_full_page_map?query=`);
+  await waitTillHTMLRendered(page)
+  await percySnapshot('vertical-full-page-map-list-view');
```
tmeyer2115, line 61: For my own edification, it seems like we're taking snapshots of the list view on desktop, map view on mobile, and detail view on mobile. Is that correct?

static/js/theme-map/Maps/Providers/Mapbox.js
tmeyer2115: Is this test currently only taking screenshots with Mapbox maps? That's totally fine for now, but it's possible in the future we want Google Maps snapshots as well. Does Google have CSS classes on its pins that we can target?

cea2aj: I'll investigate if more advanced query selectors can find the pin without this CSS class. And yes Tom, we only take screenshots with Mapbox maps. Google doesn't have CSS classes on its pins, however I will investigate if there is a more advanced selector that I could use to find one

tests/percy/snapshots.js
cea2aj: Percy takes snapshots of desktop and mobile by default. So the list view we get mobile and desktop. And the map view and detail view are only mobile. I can split up the list view snapshots to make that a little more clear

static/js/theme-map/Maps/Providers/Mapbox.js
cea2aj: I've updated to use a query that no longer requires us to add a CSS class.  If we add Google maps support in the future, we can find a pin with the following: `document.querySelector('.gm-style img').parentElement`

tmeyer2115 (approved):

oshi97 (approved):

oshi97 (approved):

PR #724, opened 2021-04-15
Fix Hitchhiker universal templates not being used for all verticals

oshi97:
This commit fixes a bug where the Hitchhiker universal section template
was only being used for verticals that had pages made for them. For example,
if the backend returned a universal section for vertical "fruits", but I did
not have a page created with verticalKey "fruits", the SDK universal section
template would be used instead of the Hitchhiker one.

This was fixed by overriding the "results/verticalresults" template.

J=SLAP-1130
TEST=manual

test that I can specify a specific universal section template
see that the hitchhiker template is always used even for verticals
by making only a universal page and seeing the hitchhiker template for
all verticls

### DIFF
--- a/templates/universal-standard/script/universalresults.hbs
+++ b/templates/universal-standard/script/universalresults.hbs
@@ -1,3 +1,9 @@
+{{!-- Override the default VerticalResults template used in UniversalResults --}}
+ANSWERS.registerTemplate(
+  "results/verticalresults",
+  {{{ stringifyPartial (read "universalsectiontemplates/standard") }}}
+);
+
 ANSWERS.addComponent("UniversalResults", Object.assign({}, {
     container: ".js-answersUniversalResults",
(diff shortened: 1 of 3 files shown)


oshi97:
checked with Rose and she's good with this fix!

cea2aj (approved):
LGTM!

cea2aj:
Should we update the card used so that the percy snapshots have some data in it for the KM vertical?

oshi97:
> Should we update the card used so that the percy snapshots have some data in it for the KM vertical?

ah yeah that's more helpful than blank boxes! Ideally the API response doesn't ever change so we don't get more red herrings (but either way that's a separate issue)

cea2aj (approved):
LGTM. I'm pretty sure that the financial-professional-location distance text should be on the right, however that's a separate bug

PR #726, opened 2021-04-15
dont remove .git folder in postupgrade script

oshi97:
Jambo now takes care of this, so we don't need to do it
in the postupgrade script. fs-extra's removeSync does not
error out if you try to remove a folder that does not exist,
so this change isn't strictly necessary. Thanks to this,
we also don't have to worry about Jambo backwards compatibility
with older theme versions.

J=SLAP-1135
TEST=manual

test upgrading to this branch, see .git is still removed


oshi97:
will merge this in after we update jambo on the off chance somebody wants to upgrade to a develop branch

tmeyer2115 (approved):

PR #727, opened 2021-04-16
open status formatter: add css classes for HH styling

rosiegrant:


cea2aj (approved):

cea2aj (changes requested):
It looks like we're going to need to update the HoursBuilder tests

oshi97:
Hijacking this PR

J=SLAP-1269
TEST=manual

check that the new css classes show up on the test-site's full-page-map cards

PR #728, opened 2021-04-16
Fix styling of nlp filters, when no collapsible-filters present

oshi97:
This commit fixes the styling of nlp filters on the vertical full page
map. This was done by removing the justify-content: space-between on
Answers-resultsHeaderTop for the full page map's css. This should not have
any negative repercussions.

This statement is based on a few ideas. First, we can divide full page maps
into two categories, has collapsible filters and does not have collapsible filters.
For pages with collapsible filters, the FilterLink component's container styling
has margin-left: auto, which will squish all of the other divs in -resultsHeaderTop
to the left, in the same way that justify-content: flex-start would (the browser default).
For pages without collapsible filters, we don't want justify-content: between, because
it causes any nlp filters that get displayed to fly all the way to the right of the page.

J=SLAP-1262
TEST=manual, auto

test that before and after of the page, after removing justify-content: space between,
is exactly the same. test for page with filters, and page without filters, using searches
that returned nlp filters and do not return nlp filters.

### DIFF
--- a/static/scss/answers/interactive-map/VerticalFullPageMap.scss
+++ b/static/scss/answers/interactive-map/VerticalFullPageMap.scss
@@ -348,7 +348,6 @@
 
     &-resultsHeaderTop {
       display: flex;
-      justify-content: space-between;
       padding-left: 16px;
     }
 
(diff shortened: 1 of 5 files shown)


oshi97:
checked with Christian and he doesn't recall any specific reason against removing this line

cea2aj (approved):

PR #730, opened 2021-04-19
Update RTF to support new elements

cea2aj:
Update RTF to support strikethroughs, superscripts, subscripts, code, code blocks, blockquotes, and headings

The styling for these elements are the defaults for Chrome, with a few tweaks. The headings use `--yxt-font-size` values rather than the defaults. The blockquote has a shorter indent because I think it works better on our cards. Finally, I made the code font smaller to make it more consistent with the default font.

J=none
TEST=manual

Use the Rose test account and the Answers rich text formatter on a card with all of the RTF that the KG supports

### DIFF
--- a/static/scss/answers/common/mixins.scss
+++ b/static/scss/answers/common/mixins.scss
@@ -84,6 +84,112 @@
   {
     margin-bottom: var(--yxt-base-spacing);
   }
+
+  s
+  {
+    text-decoration: line-through;
+  }
+
+  sup
+  {
+    vertical-align: super;
+    font-size: smaller;
+  }
+
+  sub 
+  {
+    vertical-align: sub;
+    font-size: smaller;
+  }
+
+  code 
+  {
+    font-family: monospace;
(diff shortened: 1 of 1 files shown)


cea2aj:
Screen Shot 2021-04-19 at 10 39 28 AM

static/scss/answers/common/mixins.scss
```diff
@@ -84,6 +84,112 @@
...
+    margin-inline-start: 0px;
+    margin-inline-end: 0px;
+    font-weight: bold;
+  }
+
+  h2 
+  {
+    display: block;
+    font-size: var(--yxt-font-size-lg);
+    margin-block-start: .33em;
```
oshi97, line 143: do we want to make a mixin for the margin?

cea2aj: sure, I can do that

oshi97 (approved):

oshi97 (approved):

PR #731, opened 2021-04-19
Version 1.20.1

cea2aj:
**Bugfixes**
- Remove an extraneous comma from the vertical-full-page-map template config (#725)
- Fix a layout bug with custom headers and footers on mobile list view on the vertical-full-page-map template (#729)
- Fix styling of NLP filters when no collapsible-filters are present (#728)


cea2aj:
This hotfix branch somehow included a few commits from develop including Rose's small config fix, and the updated test site with better test coverage. I actually think it's a good idea to include the updated test site because it just means we have better test coverage, and it's easier to test the new map page. What do you all think?

oshi97:
> This hotfix branch somehow included a few commits from develop including Rose's small config fix, and the updated test site with better test coverage. I actually think it's a good idea to include the updated test site because it just means we have better test coverage, and it's easier to test the new map page. What do you all think?

Yeah makes sense to me! I think the tests and comma change are basically risk free.

tmeyer2115 (commented):
Was adding all these changes to the test-site intentional?

cea2aj:
> Was adding all these changes to the test-site intentional?

Originally, no. However I think it'a s good idea because of the extra test coverage. See: https://github.com/yext/answers-hitchhiker-theme/pull/731#issuecomment-822696616

tmeyer2115:
> > Was adding all these changes to the test-site intentional?
> 
> Originally, no. However I think it'a s good idea because of the extra test coverage. See: [#731 (comment)](https://github.com/yext/answers-hitchhiker-theme/pull/731#issuecomment-822696616)

I agree with Oliver that the risk of including the testing changes is small. But, patches should be scoped to just bug fixes and the test additions related to those bug fixes. If these tests aren't in response to one of the bugs we fixed, I would argue they shouldn't be included. It becomes a slippery slope of what constitutes a patch. Having said that, I'm fine leaving these commits in for now. Just something to keep in mind moving forward.

tmeyer2115 (approved):

PR #734, opened 2021-04-19
tweak spellcheck margins

oshi97:
This commit tweaks the spellcheck component's margin. It also moves it on top of
the results header on the full-page-map.

For universal standard, it was necessary to give spellcheck a negative margin.
This is because we want the spacing above the results to be 2.35 (in terms of yxt-base-spacing)
when there is no spellcheck, but only 1.5 when there is spellcheck (a decrease of 0.85).
It is not possible to use some kind of sibling selector, because the existence of spellcheck
is determined not by whether an Answers-spellcheck exists, but whether a yxt-Spellcheck
exists inside of an Answers-spellcheck. Sadly, css doesn't support a "parent" selector.
In this case, we would need a "has a sibling of Answers-spellcheck that is a parent of
a yxt-Spellcheck", which isn't possible.

J=SLAP-1130
TEST=manual,auto

test that on vertical-grid, vertical-standard, full-page-map, and universal-standard, the
spellcheck margin matches what Rose decided each should be (with and without a spellcheck
suggestion). Rose said she does not care about the old vertical-map page anymore.

### DIFF
--- a/static/scss/answers/templates/vertical-grid.scss
+++ b/static/scss/answers/templates/vertical-grid.scss
@@ -167,9 +167,8 @@
     {
       .yxt-SpellCheck
       {
-        margin-top: calc(var(--yxt-base-spacing) / 2);
-        margin-bottom: var(--yxt-base-spacing);
-        margin-left: var(--yxt-base-spacing);
+        margin-bottom: calc(var(--yxt-base-spacing) * 1.25);
+        margin-left: calc(var(--yxt-base-spacing) * 0.25);
       }
     }
   }
--- a/static/scss/answers/templates/vertical-grid.scss
+++ b/static/scss/answers/templates/vertical-grid.scss
@@ -168,7 +168,11 @@
       .yxt-SpellCheck
       {
         margin-bottom: calc(var(--yxt-base-spacing) * 1.25);
-        margin-left: calc(var(--yxt-base-spacing) * 0.25);
+        margin-left: var(--yxt-base-spacing);
+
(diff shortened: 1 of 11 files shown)


static/scss/answers/templates/vertical-grid.scss
```diff
@@ -167,9 +167,12 @@
     {
       .yxt-SpellCheck
       {
-        margin-top: calc(var(--yxt-base-spacing) / 2);
-        margin-bottom: var(--yxt-base-spacing);
+        margin-bottom: calc(var(--yxt-base-spacing) * 1.25);
         margin-left: var(--yxt-base-spacing);
+
+        @include bpgte(sm) {
```
cea2aj, line 173: for my own edification, how are we deciding to use one of the the breakpoint mixins rather than the mobile breakpoint directly?

oshi97: I think in general we've been using the breakpoint mixins when possible, though I'm not sure if I agree with this. A lot of the time I find myself having to dife into the css plumbing to find out what breakpoint "sm" corresponds to. It's also pretty confusing that `bplte(xs)` corresponds to the same breakpoint as `bpgte(sm)` , that is 767 and 768px. Should I just use the breakpoints directly here?

oshi97: I imagine consulting had fancy reasons for breakpoint mixins, that we either didn't port over or don't take advantage of

cea2aj (approved):

PR #735, opened 2021-04-20
Use answers-modern with type="module"

oshi97:
This commit updates the answers.min.js script tags
to instead load answers-modern.min.js on modern browsers,
and answers.min.js only on browsers like ie11 that need
polyfilling. This was done using type="module" and no module
feature detections.

J=SLAP-1257
TEST=manual,auto

test that the page works on both ie11 and chrome
see that on chrome, only answers-modern is loaded,
and that answers-modern is smaller than answers.min.js

test changing default locale to 'it', and building
then adding sdkLocaleOverride of 'es', and building
then removing locale_config and specifying locale in the global_config


script/sdk-js-script-tags.hbs
```diff
@@ -0,0 +1,27 @@
+{{!--
```
tmeyer2115, line 1: What would you think of abstracting out the `sdkUrl` and `localePrefix` partials somewhere so we could also use them when generating the compiled templates script tags?

oshi97: yes that's a great idea! I updated the theme's test hbs instance so that we can test partial usages like this. I think long term we'll want jambo to be usable directly in javascript instead of only via cli, which would alleviate the need to keep this "mock jambo" around.

tmeyer2115 (approved):

PR #737, opened 2021-04-20
Add iframe Percy snapshots

cea2aj:
Add snapshots of the test site from within an iframe

J=SLAP-1264
TEST=visual

Observe updated snapshots in Percy


tests/percy/iframephotographer.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  async gotoVerticalPage({ vertical, queryParams = '' } = {}) {
+    const url = `${this._siteUrl}/${this._iframePage}?verticalUrl=${vertical}&${queryParams}`;
+    await this._page.goto(url);
+    await waitTillHTMLRendered(this._page);
+  }
+
+  async snapshot(snapshotName) {
+    await this._percySnapshot(snapshotName + '-iframe');
```
oshi97, line 36: instead of separating everything with just dashes, would we want to try something similar to BEM? it looks like the different dials we have for snapshots are page template, width, and then description? It's unfortunate that percy doesn't let us have organization beyond just a string

tests/percy/index.js
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  const iframePhotographer = new IframePhotographer({
+    page: page,
+    percySnapshot: percySnapshot,
+    siteUrl: `http://localhost:${PORT}`,
+    iframePage: 'iframe_test'
+  });
+
+  const snapshotDirector = new SnapshotDirector(standardPhotographer);
```
oshi97, line 30: this refactor is super cool!

tests/percy/photographer.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * Responsible for navigating pages and capturing snapshots
+ */
+class Photographer {
+  constructor() {}
```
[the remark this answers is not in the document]
cea2aj: I thought that I would include it to make it clear that a constructor is part of the Photographer's public interface, however I can try removing it

tests/percy/photographer.js
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Navigate to a universal page
+   * 
+   * @param {Object} obj
+   * @param {string} obj.queryParams params to be put on the page. A leading '?' or '&' should
+   *                                    not be supplied.
+   */
+  async gotoUniversalPage({ queryParams = '' } = {}) {}
```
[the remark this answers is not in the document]
cea2aj: Sure. I'm thinking of having a Photographer which handles the navigation and a Camera which takes the snapshots

tests/percy/iframephotographer.js
cea2aj: For BEM, are you think that we just use two dashes? Or something else?

tests/percy/index.js
cea2aj: Ya, I had fun with this one, and I'm hoping that it will make it easy to write code for new snapshots

tests/percy/photographer.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Responsible for navigating pages and capturing snapshots
+ */
+class Photographer {
```
tmeyer2115, line 4: Is this JSDoc still accurate? It seems like this class is only responsible for page navigation. If that's the case, I think we should change the name from `Photographer`. Maybe something like `PageNavigator`.

tests/percy/photographer.js
```diff
@@ -0,0 +1,32 @@
...
+  /**
+   * Navigate to a universal page
+   * 
+   * @param {Object} obj
+   * @param {string} obj.queryParams params to be put on the page. A leading '?' or '&' should
```
tmeyer2115, line 9: We could potentially make the interface of this and the other methods friendlier by accepting a `queryParams` object and doing the serialization internally. What do you think?

tests/percy/camera.js
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Responsible for taking Percy snapshots
+ */
+class Camera {
+  /**
+   * @param {Object} obj
+   * @param {function} obj.percySnapshot The percy snapshot function
+   * @param {boolean} obj.iframeMode Enables capturing iframe snapshots
+   */
+  constructor({ percySnapshot, iframeMode }) {
```
tmeyer2115, line 12: Since there are only two parameters here, do we need to wrap them in an object?

tests/percy/iframephotographer.js
oshi97: I was thinking something like `verticalFullPageMap--mobile__listView` or `verticalStandard__search`? It could be worth another item to think about this a little more, but I kind of like making some improvements right now since we're adding in the -iframe suffix.

oshi97: maybe we even have 4 different categories like `pageTemplate--width__description-integrationType`?

tests/percy/snapshotdirector.js
```diff
@@ -0,0 +1,91 @@
...
+  async _captureVerticalSearch () {
+    await this._photographer.gotoVerticalPage({ vertical: 'events', queryParams: 'query=a'});
+    await this._camera.snapshot('vertical-search');
+
+    await this._photographer.gotoVerticalPage({ vertical: 'events', queryParams: 'query=vrginia'});
+    await this._camera.snapshot('vertical-search-spellcheck');
+  }
+  
+  async _captureVerticalGridSearch () {
+    await this._photographer.gotoVerticalPage({ vertical: 'people', queryParams: 'query=a' });
```
oshi97, line 51: nit but I feel like a photographer normally is the one holding a camera? Here it's reads like the director is telling the photographer to drive them around while he uses the camera to take pictures which is a little different than what I first expect

oshi97: maybe snapshot directory should be renamed to Photographer and Photographer to something like PageNavigator or SnapshotNavigator (or maybe something like TourGuide if we want to be more colorful?)

tests/percy/photographer.js
cea2aj: sure, I like that

tests/percy/camera.js
cea2aj: no, I can simplify it

tests/percy/snapshotdirector.js
cea2aj: I'll update it so that the Director is now the Photographer, and the Photographers are now PageNavigators

tests/percy/photographer.js
cea2aj: Ya, that jsdoc got out of date with a refactor. PageNavigator works!

tests/percy/iframephotographer.js
cea2aj: What do you think of putting the layout next to the integratinType? Like: `pageTemplate__description-layout-integrationType` For example, that could look like: `vertical-full-page-map__nlp-filters-desktop-view-iframe`

oshi97: that's fine by me! I would just change it so that description, layout, and integration type aren't also separated by dashes.
maybe `vertical-full-page-map--nlp-filters__desktop-view__iframe`?

cea2aj: Sure, that works! And maybe a single underscore before `iframe`?

oshi97: yeah makes sense to me!

test-site/public/iframe_test.html
```diff
@@ -0,0 +1 @@
+<html> <head><meta name="viewport" content="initial-scale=1,maximum-scale=1,minimum-scale=1,user-scalable=no,width=device-width"></head> <body> <div id="answers-container"></div> <script src="iframe.js"></script> </body> </html>
```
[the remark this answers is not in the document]
cea2aj: When webpack runs on the test site, the iframe gets minified and it shows up as a diff in git, which isn't ideal. We could exclude the test file from the webpack build, however the test site uses the prod webpack config, so we would have to update it in our prod webpack config, which I'm not sure we want to do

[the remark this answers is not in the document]
oshi97: this isn't being built by webpack though, isn't this just a tracked by git (and preserved by jambo) file?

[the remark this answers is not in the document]
cea2aj: I tested this and it gets minified when webpack runs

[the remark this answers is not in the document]
oshi97: ah I see we make an html plugin instance per .html file in output regardless of whether jambo made it, sorry about that

oshi97 (approved):
looks awesome!

PR #738, opened 2021-04-21
consolidate webpack-config for locator-bundle and other bundles

oshi97:
This commit merges the locator-bundle's bespoke webpack config
into the config for the rest of the bundles. A following commit
will change the return value of the webpack-config to a single object
instead of an array. This refactor makes it easier to
modify the config based on whether the config is in development
or production mode. This commit also renames bundle.js and bundle.css
to HitchhikerJS.js and HitchhikerCSS.css.

J=SLAP-1132
TEST=manual

test that I can build the test-site, and the full-page-map and
regular page loads and can run searches


static/webpack-config.js
```diff
@@ -79,53 +79,37 @@ module.exports = function () {
...
-        filename: '[name].js',
-        path: path.resolve(__dirname, jamboConfig.dirs.output),
-        library: {
-          name: 'VerticalFullPageMap',
-          type: 'window'
-        },
-        publicPath: ''
-      },
-      module: {
-        rules: [javascriptModuleRule],
```
cea2aj, line 101: Does this mean that now the locator-bundle uses all of the same module rules as the other bundle?

static/webpack-config.js
```diff
@@ -79,53 +79,37 @@ module.exports = function () {
...
         alias: {
           static: path.resolve(__dirname, jamboConfig.dirs.output, 'static'),
         }
       },
       output: {
-        filename: '[name].js',
+        filename: (pathData) => {
+          const chunkName = pathData.chunk.name;
+          if (chunkName === 'VerticalFullPageMap') {
+            return 'locator-bundle.js';
```
cea2aj, line 107: What are your thoughts on also renaming locator-bundle to VerticalFullPageMap.js?

static/webpack-config.js
oshi97: the locator-bundle only imports .js files, so it always did!

static/webpack-config.js
oshi97: yeah we can do that! I think it would let us avoid using a function here. I was worried locator-bundle.js was something product decided, but on second thought that seems very unlikely.

templates/vertical-full-page-map/page.html.hbs
```diff
@@ -17,7 +17,7 @@
     {{> templates/vertical-full-page-map/script/locationbias modifier="main" }}
     {{> templates/vertical-full-page-map/script/locationbias modifier="mobileMap" }}
   {{/script/core }}
-  <script id="js-answersLocatorBundleScript" src="{{relativePath}}/locator-bundle.js"
+  <script id="js-answersLocatorBundleScript" src="{{relativePath}}/VerticalFullPageMap.js"
     onload="window.locatorBundleLoaded = true;" defer></script>
```
cea2aj, line 21: nit: now window.locatorBundleLoaded is a little out of sync with the other naming. Maybe we should rename to `verticalFullPageMapBundleLoaded`?

oshi97: yes definitely! renamed other usages of LocatorBundle too

cea2aj (approved):

PR #739, opened 2021-04-22
Fix locator bundle race condition

cea2aj:
Fix a race condition where the vertical-full-page-map cards would break if the search loads before the locator bundle

J=SLAP-1276
TEST=manual

Use Fiddler to delay the loading of the locator bundle and observe the cards rendering properly. The cards aren't clickable while the bundle is loading, but as soon as it does, the cards become clickable. Also test that when the bundle loads before the search, the listeners register properly.

### DIFF
--- a/templates/vertical-full-page-map/page-setup.js
+++ b/templates/vertical-full-page-map/page-setup.js
@@ -7,5 +7,25 @@ if (window.locatorBundleLoaded) {
   loadFullPageMap();
 } else {
   const locatorBundleScript = document.querySelector('script#js-answersLocatorBundleScript');
-  locatorBundleScript.onload = loadFullPageMap;
+  locatorBundleScript.onload = () => {
+    window.locatorBundleLoaded = true;
+    locatorBundleScript.dispatchEvent(new Event('locator-bundle-loaded'));
+    loadFullPageMap();
+  }
+}
+
+/**
+ * Registers listeners on the card once the locator bundle is loaded
+ *
+ * @param {ANSWERS.Component} card A location card
+ */
+function registerVerticalFullPageMapCardListeners(card) {
+  if (window.locatorBundleLoaded) {
+    new VerticalFullPageMap.CardListenerAssigner({card: card}).addListenersToCard();
+    return;
+  }
+  const locatorBundleScript = document.querySelector('script#js-answersLocatorBundleScript');
+  locatorBundleScript.addEventListener('locator-bundle-loaded', () => {
+    new VerticalFullPageMap.CardListenerAssigner({card: card}).addListenersToCard();
+  });
 }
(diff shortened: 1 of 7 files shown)


tmeyer2115 (commented):
Can we not guarantee that the locator bundle loads first? If we place it in the document before `answers.min.js` and give it a `defer` tag, would that also solve this issue?

templates/vertical-full-page-map/page-setup.js
```diff
@@ -7,5 +7,25 @@ if (window.locatorBundleLoaded) {
...
 }
+
+/**
+ * Registers listeners on the card once the locator bundle is loaded
+ *
+ * @param {ANSWERS.Component} card A location card
+ */
+function registerVerticalFullPageMapCardListeners(card) {
+  if (window.locatorBundleLoaded) {
+    new VerticalFullPageMap.CardListenerAssigner({card: card}).addListenersToCard();
```
oshi97, line 24: Would it be possible to move CardListenerAssigner into HitchhikerJS? I know in terms of bundle size it's not optimal, but I think it could be much simpler? Long term I think we would want to get the cards into per page bundles, instead of inlining them, and tree shake the theme (possibly using the cards as entry points)

oshi97: oh I think Tom's suggestion also makes sense! I was worried that flipping the script order would cause the search on load to be delayed, but
1. not sure if that's the case and
2. even if it is, we may prefer that over moving the CardListenerAssigner to the HitchhikerJS and pushing it onto other pages.

cea2aj:
> Can we not guarantee that the locator bundle loads first? If we place it in the document before `answers.min.js` and give it a `defer` tag, would that also solve this issue?

Sure, I can try that. As Oliver mentioned, that will likely delay the firing of any initial search. I will investigate if it's an acceptable amount of delay

tmeyer2115 (approved):

cea2aj:
> > Can we not guarantee that the locator bundle loads first? If we place it in the document before `answers.min.js` and give it a `defer` tag, would that also solve this issue?
> 
> Sure, I can try that. As Oliver mentioned, that will likely delay the firing of any initial search. I will investigate if it's an acceptable amount of delay

Update: the issue with loading the locator first is that the locator bundle depends on the ANSWERS instance since the locator components are ANSWERS components

templates/vertical-full-page-map/page-setup.js
cea2aj: Ya, I think that would work. There isn't much code in the CardListenerAssigner, so the impact on the HH bundle would be minimal. Since Tom is good to move forward with this solution, I'll make an item for investigating that approach in the future. That might resolve the issue where the map loads while being zoomed in.

PR #741, opened 2021-04-23
add prod and dev webpack configs

oshi97:
This commit adds additional optimizations to the webpack
build when IS_DEVELOPMENT_PREVIEW='true'.
It does this by adding dev and prod webpack configs, which
are merged with the shared webpack config using webpack-merge.

J=SLAP-1132
TEST=manual

test building the page with unset IS_DEVELOPMENT_PREVIEW,
IS_DEVELOPMENT_PREVIEW='false', and IS_DEVELOPMENT_PREVIEW='true',
and that the correct build mode is run for each

see that the build is much faster in development mode, and
that on ie11 I get a "Promise is not defined" error

see that ie11 still loads properly when in production mode,
and that assets are inlined
noticed an issue with ie11 and the full-page-map. this issue existed before
the webpack config changes (checked on master and develop), the issue
seems unrelated to the recent race condition found, because during my
testing it happened 100% of the time on browserstack


cea2aj:
> It does this by adds

oshi97:
> > It does this by adds

fixed!

cea2aj (approved):
lgtm

PR #742, opened 2021-04-23
VerticalFullPageMap IE11 compatibility improvements

cea2aj:
Update the VerticalFullPageMap to be fully IE11 compliant

We were using some syntax which wasn't supported by IE11. Update the locator to account for that. Also remove the flex on the mobile view toggle button because that was not being formatted properly on IE11. Removing the flex didn't change anything that I could see, and Percy didn't pick up any changes either.

Fixes for IE11 compatibility:
- For MediaQueryLists, use addListener rather than addEventListener
- Add a `removeElement()` helper function with IE11 support (essentially a ponyfill)
- Use setAttribute and removeAttribute for updating the detail card styling

J=none
TEST=manual

Test on IE11 and chrome and make sure that erros are no longer produced in the console on IE11, and test various functionality including pin clustering, opening and closing details cards, and toggling between mobile views.

### DIFF
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -1,6 +1,6 @@
 import { Coordinate } from './Geo/Coordinate.js';
 import { smoothScroll } from './Util/SmoothScroll.js';
-import { getLanguageForProvider, isViewableWithinContainer } from './Util/helpers.js';
+import { getLanguageForProvider, isViewableWithinContainer, removeElement } from './Util/helpers.js';
 import { SearchDebouncer } from './SearchDebouncer';
 import { defaultCenterCoordinate } from './constants.js';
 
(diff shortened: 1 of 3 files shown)


static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -443,11 +449,27 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
 
     this.core.storage.set(StorageKeys.LOCATOR_SELECTED_RESULT, null);
   }
 
+  /**
+   * Removes an element from the DOM, with support for IE11
+   * 
+   * @param {Element} element
+   */
+  _removeElement(element) {
```
oshi97, line 462: could this go somewhere in the "Util" folder?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -474,9 +496,9 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
 
       if (!isDetailCardOpened) {
         window.requestAnimationFrame(() => {
-          this._detailCard.style = 'height: 0;';
+          this._detailCard.setAttribute('style', 'height: 0;');
```
oshi97, line 499: do we need to worry about other usages of `.style` in the theme?

oshi97 (commented):
👼 🙏 🙇

static/js/theme-map/VerticalFullPageMapOrchestrator.js
cea2aj: The error happens only when you try to modify the entire style object like `element.style = 'height: 0;'` This is the only spot in the theme where we were modifying style like that.  Modifying it like `style.height = '300px'` or `style['height'] = '200px'` works fine in IE11. That's how we usually modify style in the theme

cea2aj: I think the `setAttribute` call works here because it corresponds with `removeAttribute`

static/js/theme-map/VerticalFullPageMapOrchestrator.js
cea2aj: sure

tmeyer2115 (approved):

PR #749, opened 2021-04-29
Setup Acceptance Test Infrastructure

cea2aj:
Setup infrastructure for running browserstack tests on the theme test site

J=SLAP-1268
TEST=acceptance

Confirm that the tests run in browswerstack. Observe the browserstack video and see that the test site loads


.github/workflows/acceptance.yml
```diff
@@ -0,0 +1,24 @@
...
+          project-name: 'Answers Hitchhiker Theme'
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - run: npm ci
+      - run: npm run setup-test-site
+      - run: npm run build-test-site
+      - name: Run Acceptance Tests
+        run: 'npx testcafe "browserstack:chrome,browserstack:firefox,browserstack:safari,browserstack:ie@11" tests/acceptance/suites/*.js'
```
oshi97, line 24: Do we want to use headless browsers for chrome and firefox? I think with the current setup each PR into the theme will use all 8 of our threads. I think it's slightly better to use a browserstack browser, but would also feel bad about asking for another thread bump so soon.

.github/workflows/acceptance.yml
```diff
@@ -0,0 +1,24 @@
...
+        uses: 'browserstack/github-actions/setup-env@master'
+        with:
+          username:  ${{ secrets.BROWSERSTACK_USERNAME }}
+          access-key: ${{ secrets.BROWSERSTACK_ACCESS_KEY }}
+          project-name: 'Answers Hitchhiker Theme'
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - run: npm ci
```
oshi97, line 20: I think we should group together tests under the same PR at https://automate.browserstack.com/dashboard/v2/ 
Currently we're getting one build per browser tested, with one session each, as opposed to
having multiple sessions within a single build
the SDK does this by exporting some browserstack env vars
https://github.com/yext/answers-search-ui/blob/fc4fd42f94d2a693f727b2b480611b8ba790affb/.circleci/run_browserstack_acceptance.sh

We could also try using a browserstack-config.json which I think would be slightly cleaner. I can't remember why I didn't do this originally for the SDK.
https://github.com/DevExpress/testcafe-browser-provider-browserstack#configuration-file
https://www.browserstack.com/docs/automate/selenium/getting-started/nodejs/testcafe#executing-tests-on-browserstack

tests/acceptance/server.js
```diff
@@ -0,0 +1,23 @@
+const HttpServer = require('../test-utils/server');
+const path = require('path');
+
+/**
+ * Initalizes an http server for the provided fixture context
+ * @param {*} ctx the context from the fixture
+ */
+export async function setupServer (ctx) {
```
oshi97, line 8: since we're starting fresh, could we try using the testcafe server infrastructure? I ran into issues trying to parallelize our acceptance tests in the past because we were hardcoding it to a single port
https://testcafe.io/documentation/402641/reference/testcafe-api/runner

.github/workflows/acceptance.yml
cea2aj: That works. I wasn't sure if our extra capacity would let us run all 4 browsers or not

.github/run_browserstack_acceptance.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
```
oshi97, line 2: nit but what do you think about removing this space?

cea2aj: sure

.github/workflows/acceptance.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - run: npm ci
+      - run: npm run setup-test-site
+      - run: npm run build-test-site
+      - name: Run Acceptance Tests
+        run: node tests/acceptance/run-tests.js --browsers chrome:headless safari:headless
```
oshi97: safari:headless isn't a testcafe browser yet, though we could probably download our own if we really wanted (or just use browserstack)

.github/workflows/acceptance.yml
```diff
@@ -0,0 +1,37 @@
...
+        shell: bash
```
oshi97, line 37: do we need this `shell: bash`?

tests/acceptance/run-tests.js
```diff
@@ -0,0 +1,35 @@
+const createTestCafe = require('testcafe');
+const { PORT } = require('./constants');
+const yargs = require('yargs');
+
+const argv = yargs
```
oshi97, line 5: This is a cool solution! I didn't think about the fact that we run two separate instances of this script with different browsers

.github/workflows/acceptance.yml
```diff
@@ -0,0 +1,37 @@
...
+      BROWSERSTACK_ACCESS_KEY: ${{ secrets.BROWSERSTACK_ACCESS_KEY }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - run: npm ci
+      - run: npm run setup-test-site
+      - run: npm run build-test-site
+      - run: chmod +x ./.github/run_browserstack_acceptance.sh
```
oshi97, line 20: could we have github track this script as an executable file instead?

.github/workflows/acceptance.yml
cea2aj: Updated so that sfari is in browserstack and firefox is headless, and the tests passed

.github/workflows/acceptance.yml
cea2aj: oh sure, I didn't realize that was possible

.github/workflows/acceptance.yml
cea2aj: I'll test it without it

tests/acceptance/run-tests.js
cea2aj: Thanks. I figured it was a nice way of running the tests

package.json
```diff
@@ -35,8 +36,11 @@
     "postcss": "^8.2.1",
     "serve": "^11.3.2",
     "simple-git": "^2.24.0",
+    "testcafe": "^1.14.0",
+    "testcafe-browser-provider-browserstack": "^1.13.2-alpha.1",
     "underscore.string": "^3.3.5",
-    "urijs": "1.18.12"
+    "urijs": "1.18.12",
+    "yargs": "^16.2.0"
```
oshi97, line 43: oh nit but can we try yargs 17? It was released yesterday

oshi97: I was looking at yargs@17 for my jambo acceptance stuff, and they improved a lot of async/promise related things

cea2aj: > It was released yesterday

Here at Yext we live on the cutting edge

oshi97 (approved):
lgtm!

PR #751, opened 2021-05-04
Vertical Full Page Map Acceptance Tests

cea2aj:
Test core functionality of the vertical-full-page-map on desktop

Tests:
- Performing a search
- Clicking a pin and focusing a card
- Search when map moves
- Search this area button

Contrary to the SDK acceptance tests, these tests do not organize blocks into pages. They simply use the blocks directly from the tests. I think this leads to simpler and cleaner code since there are less layers of abstraction that need to be maintained. I believe the SDK pages architecture was setup to mimic some of the testcafe examples where tests are organized into pages. I think that makes sense when each page has different sets of functionality, however, since the SDK is component based, I think organizing the components into blocks is more effective.

SLAP-1268
TEST=acceptance

Run tests and confirm they pass


oshi97 (approved):
🍍 🍍 🍍 !

PR #754, opened 2021-05-06
Add iframe JWT support

cea2aj:
Add iframe JWT support which is enabled when useJWT is true in global config

For iframe experiences, add iframe-jwt.js (along with prod and staging versions) which provide an `initAnswersFrameJWT` function. The answers experience will load once `initAnswersFrameJWT` is called.

For non-iframe experiences, `initAnswersJWT` must be called rather than `initAnswersFrameJWT`

The next PR will ensure that the apiKey from the jambo injected data will not appear in the output of any builds.

This PR does not include JWT overlay support.

J=SLAP-1118
TEST=manual

Test iframe-jwt.js and see that the page doesn't load until initAnswersFrame is ran. Test supplying valid and invalid tokens. Test navigating between pages and refreshing the page. Test that if the answers experience with `useJWT=true` is loaded directly (not through the iframe), and error message appears in the console. Smoke test the non-iframe experience and the the overlay integration with `useJWT=false`.  Test JWT locally and on SGS. Test in a cross-domain manner with a local test site which points to an answers experience on SGS which is running this branch. On SGS, test iframe-jwt-prod.js and iframe-jwt-staging.js

Test JWT on non-iframe experiences.

### DIFF
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -123,9 +123,8 @@
       window.iframeLoaded = new Promise(resolve => {
         iframeLoadedResolve = resolve;
       });
-      let iframeJWTResolve;
-      window.iframeJWT = new Promise(resolve => {
-        iframeJWTResolve = resolve;
+      window.answersJWT = new Promise(resolve => {
+        window.initAnswersJWT = resolve;
       });
       window.iFrameResizer = {
         onReady: function() {
@@ -138,7 +137,7 @@
         },
         onMessage: (message) => {
           if (message.token) {
-            iframeJWTResolve(message.token);
+            initAnswersJWT(message.token);
             return;
           }
           window.isOverlay && window.Overlay.onMessage(message);
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -126,7 +126,7 @@
       let iframeJWTResolve;
       window.iframeJWT = new Promise(resolve => {
         iframeJWTResolve = resolve;
-      })
+      });
       window.iFrameResizer = {
         onReady: function() {
           window.parentIFrame.sendMessage(JSON.stringify({
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -123,6 +123,10 @@
       window.iframeLoaded = new Promise(resolve => {
         iframeLoadedResolve = resolve;
       });
+      let iframeJWTResolve;
+      window.iframeJWT = new Promise(resolve => {
+        iframeJWTResolve = resolve;
+      })
       window.iFrameResizer = {
         onReady: function() {
           window.parentIFrame.sendMessage(JSON.stringify({
@@ -132,7 +136,13 @@
           }));
           iframeLoadedResolve();
         },
-        onMessage: window.isOverlay ? window.Overlay.onMessage : function() {},
+        onMessage: (message) => {
+          if (message.token) {
+            iframeJWTResolve(message.token);
+            return;
+          }
+          window.isOverlay && window.Overlay.onMessage(message);
+        }
       };
     {{/babel}}
     </script>
--- a/script/core.hbs
+++ b/script/core.hbs
@@ -34,10 +34,10 @@
       const jwtNotProvidedTimeout = setTimeout(() => {
         console.error(
           'A JWT is required to initialize this Answers experience because "useJWT" is set to true.\n' +
-          'To specify a JWT, this page must be loaded through iframe-jwt.js or similar.'
+          'Load the experience by calling initAnswersJWT(token) or initAnswersFrameJWT(token).'
         );
       }, 2000);
-      window.iframeJWT.then(token => {
+      window.answersJWT.then(token => {
         clearTimeout(jwtNotProvidedTimeout);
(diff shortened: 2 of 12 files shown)


layouts/html.hbs
```diff
@@ -123,6 +123,10 @@
       window.iframeLoaded = new Promise(resolve => {
         iframeLoadedResolve = resolve;
       });
+      let iframeJWTResolve;
+      window.iframeJWT = new Promise(resolve => {
+        iframeJWTResolve = resolve;
+      })
```
oshi97, line 129: nit missing semicolon!

layouts/html.hbs
```diff
@@ -132,7 +136,13 @@
           }));
           iframeLoadedResolve();
         },
-        onMessage: window.isOverlay ? window.Overlay.onMessage : function() {},
+        onMessage: (message) => {
+          if (message.token) {
+            iframeJWTResolve(message.token);
```
oshi97, line 141: does an error occur if iframeJWTResolve is resolved multiple times?

script/core.hbs
```diff
@@ -28,10 +30,32 @@
         {{/if}}
       {{/with}}
     };
+    {{#if global_config.useJWT}}
+      const jwtNotProvidedTimeout = setTimeout(() => {
+        console.error(
+          'Since useJWT is set, a JWT is required to initialize this Answers experience.\n' +
+          'To specify a JWT, this page must be loaded through iframe-jwt.js or similar.'
+        );
+      }, 2000);
```
oshi97, line 39: Should we use a larger timeout here, or is it very unlikely to get a false positive here even on slow connections?

layouts/html.hbs
cea2aj: Nope. Also, only the first resolve sets the promise's fulfilled value

script/core.hbs
cea2aj: It's quite a safe timeout. I didn't get any false positives even when setting the network to 'Slow 3G' and the cpu to '6x slowdown'. The timer starts ticking when the experience loads, and it stops once the iframe receives the JWT from the parent. So the timeout only has to be larger than the time it takes for the iframeResizer to send the initial message from the parent to the iframe, which is quick.

static/js/iframe-common.js
```diff
@@ -82,6 +82,11 @@ export function generateIFrame(domain, queryParam, urlParam) {
   // For dynamic iFrame resizing
   iFrameResize({
     checkOrigin: false,
+    onInit: function(iframe) {
+      iframe.iFrameResizer.sendMessage({
```
tmeyer2115, line 86: Should we guard this call by the presence of a `token`? In the case of a normal, non-JWT iFrame integration, would this cause us to be sending empty messages around?

tmeyer2115 (commented):
One question I had, and this may be more for @rosiegrant , how would JWT support work for someone using the Theme without an iFrame integration? I'm assuming that's a use case we should support. Would they override `core.hbs` to invoke the token getter somewhere and pass that to the `ANSWERS.init`?

static/js/iframe-common.js
cea2aj: yep, good catch

layouts/html.hbs
```diff
@@ -123,6 +123,10 @@
       window.iframeLoaded = new Promise(resolve => {
         iframeLoadedResolve = resolve;
       });
+      let iframeJWTResolve;
+      window.iframeJWT = new Promise(resolve => {
```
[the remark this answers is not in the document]
cea2aj: html.hbs executes before core.hbs, right? I want to prevent the possibility of that function trying to be called before it is put on the window

script/core.hbs
cea2aj: Since this PR now supports non-iframe experiences, I've bumped the timeout to 5 seconds and am switching it to a warning rather than an error.

cea2aj:
> One question I had, and this may be more for @rosiegrant , how would JWT support work for someone using the Theme without an iFrame integration? I'm assuming that's a use case we should support. Would they override `core.hbs` to invoke the token getter somewhere and pass that to the `ANSWERS.init`?

Updated so that non-iframe experiences can be loaded with JWT with initAnswersJWT

script/core.hbs
```diff
@@ -28,10 +30,32 @@
         {{/if}}
       {{/with}}
     };
+    {{#if global_config.useJWT}}
+      const jwtNotProvidedTimeout = setTimeout(() => {
+        console.warn(
+          'A JWT is required to initialize this Answers experience because "useJWT" is set to true.\n' +
```
oshi97, line 36: Should we change this to say something along the lines of "a JWT has not been received within 5s of page load"

cea2aj: Ya I think that makes sense. I'll check with Rose on this because that was the error message originally requested in the spec

oshi97: ah I didn't realize the error message was from the spec, thought it was just a nice thing you came up with

tmeyer2115 (approved):

PR #755, opened 2021-05-06
use direct answer's timeZoneUtcOffset in openStatus

oshi97:
use direct answer's timeZoneUtcOffset in openStatus

This commit updates the allfields standard direct
answer cards to use the timeZoneUtcOffset in the
direct answer. This timezone offset is the UTC offset
for the location that was returned as the direct answer.
Without this piece of information, the open status can be
formatted incorrectly if the location is in a different
timezone than the user.

J=SLAP-1270
TEST=manual

checked that, on a universal page with without this change, the direct
answer open status, and the open status of a card displaying the same entity
could be out of sync, but with this change they would be in sync

check that the page loads in ie11

will add a percy snapshot (or acceptance test) in another PR after investigating
issues with percy snapshots not passing query params properly on universal pages
see this build, where the url for the iframe main page and then the iframe-ee are
inserted into the html
https://percy.io/cd93112c/answers-hitchhiker-theme/builds/10280939/changed/578794826

### DIFF
--- a/directanswercards/allfields-standard/component.js
+++ b/directanswercards/allfields-standard/component.js
@@ -103,10 +103,21 @@ class allfields_standardComponent extends BaseDirectAnswerCard['allfields-standa
         value = isArray ? arrayValue : regularValue;
         break;
       case 'hours':
+        const timezoneOffsetForLocation = relatedItem?.data?.fieldValues?.timeZoneUtcOffset;
         if (isArray) {
-          arrayValue = answer.value.map((value) => `<div>${Formatter.openStatus({hours: value})}</div>`);
+          arrayValue = answer.value.map((value) => {
+            const openStatus = Formatter.openStatus({
+              hours: value,
+              timeZoneUtcOffset: timezoneOffsetForLocation
+            });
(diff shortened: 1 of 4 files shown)


directanswercards/allfields-standard/component.js
```diff
@@ -103,10 +103,21 @@ class allfields_standardComponent extends BaseDirectAnswerCard['allfields-standa
         value = isArray ? arrayValue : regularValue;
         break;
       case 'hours':
+        const timezoneOffsetForLocation = relatedItem?.data?.fieldValues?.timeZoneUtcOffset;
```
tmeyer2115, line 106: This is some pretty snazzy syntactic sugar.

tmeyer2115 (approved):

directanswercards/allfields-standard/component.js
oshi97: haha yes! just double checked that it works on ie11 thanks to babel

cea2aj: I agree. It's probably a good idea for us to use optional chaining more frequently in our javascript

PR #756, opened 2021-05-07
Clean the output when useJWT=true

cea2aj:
Remove the apiKey from the build output when useJWT=true

The apiKey needs to be removed in the templatedataformatter and in the JAMBO_INJECTED_DATA that weback puts in the bundles.

J=J=SLAP-1118
TEST=manual, auto

When useJWT is true, run a build and search for the apiKey and confirm that it does not appear anywhere in the build output. Test this with JAMBO_INJECTED_DATA which contains apiKeys. Smoke test the JWT integration and the standard integration. Add unit test.


static/webpack-config.js
```diff
@@ -25,11 +26,29 @@ module.exports = function () {
...
+    ? getSecuredJamboInjectedData(jamboInjectedData)
+    : jamboInjectedData
+
   const plugins = [
     new MiniCssExtractPlugin({ filename: '[name].css' }),
     ...htmlPlugins,
-    new webpack.EnvironmentPlugin({
-      JAMBO_INJECTED_DATA: null
+    new webpack.DefinePlugin({
+      'process.env.JAMBO_INJECTED_DATA': JSON.stringify(JSON.stringify(updatedJamboInjectedData))
```
tmeyer2115, line 51: For my own edification, what's the scope of using the `JAMBO_INJECTED_DATA` in our Webpack toolchain? I believe we use it in the iFrame assets to get domains and such. Are we embedding the whole object in any of our static assets or using the apiKey in any of the static assets?

static/webpack/getSecuredJamboInjectedData.js
```diff
@@ -0,0 +1,32 @@
+const _ = require('lodash');
+
+/**
+ * Returns JAMBO_INJECTED_DATA with instances of the global config's apiKey removed
+ * 
+ * @param {Object} data JAMBO_INJECTED_DATA
+ * @returns {Object}
+ */
+function getSecuredJamboInjectedData (data) {
```
tmeyer2115, line 9: Small nit: The phrase "secured" implies certain things to me. For instance, it suggests that maybe that data is encrypted or something like that. What would you think of using another adjective? We're kind of "cleaning" the data here or "scrubbing" it.

oshi97:
it looks like there's something funny with the iframe snapshots

static/webpack/getSecuredJamboInjectedData.js
```diff
@@ -0,0 +1,32 @@
...
+
+  const removeApiKeyFromConfig = config => {
+    if ('apiKey' in config) {
+      delete config['apiKey'];
+    }
+  }
+
+  Object.values(experiences).forEach(config => {
+    removeApiKeyFromConfig(config);
+    if (!('configByLabel' in config)) {
```
oshi97, line 24: nit but it might be simpler to just do if `if ('configByLabel' in config) {` instead of the negation?

hooks/templatedataformatter.js
```diff
@@ -50,4 +54,16 @@ function getLocalizedGlobalConfig(globalConfig, currentLocaleConfig, locale) {
...
   }
   return localizedGlobalConfig;
+}
+
+function getSecuredTemplateData(templateData) {
+  const jamboInjectedData = templateData.env.JAMBO_INJECTED_DATA;
+  const globalConfig = templateData.global_config;
+  return {
+    ...templateData,
+    global_config: Object.assign({}, globalConfig, { apiKey: undefined }),
```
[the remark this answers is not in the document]
cea2aj: sure

static/webpack/getSecuredJamboInjectedData.js
cea2aj: sure

static/webpack-config.js
cea2aj: We do put the whole object in a few spots, and that ends up in many of our assets.

Webpack puts it into the HitchhikerJS.getInjectedProp() function. The only spot where that accesses the apiKey in our code is in core.hbs, but I disabled that access when useJWT is true. It's possible that someone might be trying to get the apiKey from getInjectedProp(), but that really doesn't make sense if a site is built with useJWT=true.

Secondly, it's put in the InjectedData class which simply provides info about the domains, so the apiKey is not needed there.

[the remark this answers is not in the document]
cea2aj: I prefer to leave it stringified twice so that we don't have to update outside classes, but let me know what you think! I tried the single stringified approach and it seems to cause some problems

cea2aj:
> it looks like there's something funny with the iframe snapshots

The issue was that I needed to fall back to null if JAMBO_INJECTED_DATA was not supplied

static/webpack-config.js
```diff
@@ -25,11 +26,31 @@ module.exports = function () {
     });
   }
 
+  const globalConfigPath = `./${jamboConfig.dirs.config}/global_config.json`;
+  let globalConfig = null;
```
oshi97, line 30: if we default globalConfig to empty object we can change `useJwt = globalConfig && globalConfig.useJWT` to just `useJWT = globalConfig.useJWT` or even `{ useJWT } = globalConfig`

static/webpack-config.js
```diff
@@ -25,11 +26,31 @@ module.exports = function () {
...
+    globalConfig = parse(globalConfigRaw);
+  }
+
+  const useJWT = globalConfig && globalConfig.useJWT
+  let jamboInjectedData = process.env.JAMBO_INJECTED_DATA || null;
+  jamboInjectedData = (jamboInjectedData && JSON.parse(jamboInjectedData));
+
+  const getCleanedJamboInjectedData =
+    require(`./${jamboConfig.dirs.output}/static/webpack/getCleanedJamboInjectedData.js`);
+
```
oshi97, line 42: process.env.JAMBO_INJECTED_DATA comes in as a string, so would it be possible to only do the parsing and restringifying if useJWT is true? maybe something like the below. Then I think we can also remove the ternary operator in the DefinePlugin config below

JSON.parse(null) just returns null
```js
let jamboInjectedData = process.env.JAMBO_INJECTED_DATA || null;
if (useJWT && jamboInjectedData) {
  const getCleanedJamboInjectedData =
    require(`./${jamboConfig.dirs.output}/static/webpack/getCleanedJamboInjectedData.js`);
  jamboInjectedData = JSON.parse(jamboInjectedData)
  jamboInjectedData = getCleanedJamboInjectedData(jamboInjectedData)
  jamboInjectedData = JSON.stringify(jamboInjectedData)
}
```

cea2aj: I like that better. updated!

oshi97 (approved):
lgtm!

PR #759, opened 2021-05-07
add Coveralls github workflow

oshi97:
Adds a code coverage workflow using Coveralls
Only static .js files have their coverage collected.
Trying to collect code coverage from hbs files masquerading
as js files causes the coverage reporter to crash, since it
can't parse the hbs portions of the file correctly.

Also adds a code coverage badge to the readme.

J=SLAP-873
TEST=auto

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,30 @@
+# This workflow will run our tests, generate an lcov code coverage file,
+# and send that coverage to Coveralls 
+
+name: Code Coverage
+
+on:
+  push:
+    branches-ignore: dev/*
+  pull_request:
+
+jobs:
+  Coveralls:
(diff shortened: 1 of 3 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,30 @@
...
+        node-version: [15.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npx jest --coverage
```
tmeyer2115, line 26: Since we only generate coverage for `tests/static/**/*.js`, could we just run the jest tests in that directory?

tmeyer2115 (approved):

PR #766, opened 2021-05-11
use express instead of serve-handler for snapshots

oshi97:
This pr updates our tests to use express instead of serve-handler.
serve (likely due to using serve-handler) currently has a bug where,
if any redirects occur, it will lose any query params you had in the url.
See: vercel/serve#643

On top of that, serve will automatically redirect urls like
/index.html to just /, which breaks our iframe integration for universal.

J=SLAP-1313
TEST=auto

percy snapshots for universal on iframe are now corrected and
show the correct query


tmeyer2115 (approved):
Out of scope for this PR, but should our iFrame integration be able to handle redirects from /index.html to /?

oshi97:
> Out of scope for this PR, but should our iFrame integration be able to handle redirects from /index.html to /?

Our iframe integration *is* able to handle redirects like that, it's just that serve isn't

PR #776, opened 2021-05-12
Add dynamic iframe config

cea2aj:


static/js/iframe-common.js
```diff
@@ -1,15 +1,23 @@
...
+ */
+export function generateIFrame(domain, hitchhikerConfigManager) {
   var isLocalHost = window.location.host.split(':')[0] === 'localhost';
   var containerEl = document.querySelector('#answers-container');
   var iframe = document.createElement('iframe');
   var pathToIndex = containerEl.dataset.path;
   iframe.allow = 'geolocation';
 
   domain = domain || '';
-  queryParam = queryParam || 'query';
```
cea2aj, line 11: I removed these because I could not find anywhere that they were being used

layouts/html.hbs
```diff
@@ -126,6 +126,10 @@
       window.answersJWT = new Promise(resolve => {
         window.initAnswersJWT = resolve;
       });
+      const hhConfigMetadata = {
+        lastReceived: -1,
```
tmeyer2115, line 130: What would you think about adding a `lastUpdatedTimestamp` or something like that to the `ConfigManager` itself? We could then use that in the comparison below.

layouts/html.hbs
```diff
@@ -136,8 +140,20 @@
...
+          const messageContainsNewHHConfig = message.hitchhikerConfig
+             && message.timestamp >= hhConfigMetadata.lastReceived;
+          if (messageContainsNewHHConfig) {
+            window.HitchhikerConfig = {
+              ...window.HitchhikerConfig,
+              ...message.hitchhikerConfig
+            };
+            hhConfigMetadata.lastReceived = message.timestamp;
+          }
+          if (message.hitchhikerConfig?.token && !hhConfigMetadata.hasInitializedWithJWT) {
```
tmeyer2115, line 154: Since Answers can't be initialized twice, I'm wondering if it would be beneficial to have a more general variable than `hasInitializedWithJWT`. Something like `hasAnswersInitialized`. We could then use that here and in other situations potentially.

static/js/hitchhikerconfigmanager.js
```diff
@@ -0,0 +1,55 @@
...
+  getObject () {
+    return { ...this._hitchhikerConfig };
+  }
+
+  set (key, value) {
+    this._validate(key, value);
+    this._hitchhikerConfig[key] = value;
+  }
+
+  sendConfigToIframe() {
```
tmeyer2115, line 20: I think it might be a bit cleaner if we leave this logic to `iframe-common`. Without this method, we could just make this class `HitchhikerConfig.js`. It's purely a run-time data store. Consumers of the data store can add additional behaviors, such as sending the info to an iFrame via a message. In theory, someone could use this thing in the non-iFrame integration.

static/js/iframe-common.js
oshi97: untangle one 🍝  a day to keep the smell away

static/js/utils/getHHConfigAttributes.js
```diff
@@ -0,0 +1,28 @@
...
+ * Returns an object which contains the data-hhconfig attributes on the element specified by the selector
+ * 
+ * @example
+ * If the attributes on the element contain 'data-hhconfig-link-target="_top"', the returned
+ * object will be { link-target: "_top" }
+ *
+ * @param {string} selector
+ * @returns {Object}
+ */
+export default function getHHConfigAttributes (selector) {
```
oshi97, line 11: can we get some unit tests for this function? would make it easier to visualize as well

static/js/utils/getHHConfigAttributes.js
```diff
@@ -0,0 +1,28 @@
...
+  const element = document.querySelector(selector);
+  if (!element) {
+    return {};
+  }
+  const attributes = element.attributes;
```
oshi97, line 16: I think it'd be simpler to access the data-* attributes directly (`element.dataset`)

static/js/iframe-prod.js
```diff
@@ -1,5 +1,9 @@
 import { generateIFrame } from './iframe-common';
 import InjectedData from './models/InjectedData';
+import { HitchhikerConfigManager, exposeOnWindow } from './hitchhikerconfigmanager';
+
+const hitchhikerConfigManager = new HitchhikerConfigManager();
+exposeOnWindow(hitchhikerConfigManager);
```
oshi97, line 6: do the static iframe scripts need hitchhikerConfigManager?

test-site/public/iframe_test.html
```diff
@@ -1 +1 @@
-<html><head><meta name="viewport" content="initial-scale=1,maximum-scale=1,minimum-scale=1,user-scalable=no,width=device-width"></head><body><div id="answers-container"></div><script src="iframe.js"></script></body></html>
\ No newline at end of file
+<html><head><meta name="viewport" content="initial-scale=1,maximum-scale=1,minimum-scale=1,user-scalable=no,width=device-width"></head><body><div id="answers-container" data-hhconfig-test="test-data"></div><script src="iframe.js"></script></body></html>
```
oshi97, line 1: are we committing this file just for easier manual testing in the future?

static/js/utils/getInitAnswersFrame.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Returns a function which generates an iframed Answers experience.
+ * If the function is not called by the timeout, a warning is logged.
+ * 
+ * @param {string} domain 
+ * @param {HitchhikerConfigManager} hitchhikerConfigManager 
+ * @returns {Function}
+ */
+export default function getInitAnswersFrame (domain, hitchhikerConfigManager) {
```
oshi97, line 11: would something like generateInitAnswersFrame be better? We're not getting a preexisting function but generating a new one

static/js/hitchhikerconfigmanager.js
cea2aj: sure, that sounds good

layouts/html.hbs
cea2aj: Sure, and include that data within the hitchhikerConfig data which is sent (as opposed to having it at the top level of the message)?

static/js/hitchhikerconfigmanager.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+}
+
+/**
+ * Puts a setter on the window which allows the hitchhiker config to be updated
+ *  
+ * @param {HitchhikerConfigManager} hitchhikerConfigManager 
+ */
+export function exposeOnWindow (hitchhikerConfigManager) {
```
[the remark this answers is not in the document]
cea2aj: ya, that makes sense

static/js/iframe-prod.js
cea2aj: ya, I will check with product, but having it here allows the someone with a regular iframe experience to update the config from the parent, even it it doesn't happen during the init

test-site/public/iframe_test.html
cea2aj: Ya, I feel that it's helpful to have an example of using the data attribute

static/js/utils/getInitAnswersFrame.js
cea2aj: ya, that works

static/js/utils/getHHConfigAttributes.js
cea2aj: I was hesitant to use that due to the automatic conversion to camelcase which I thought might make complicate things. I'll get product's input on it

oshi97: ah that's fair, feel free to ack this

layouts/html.hbs
cea2aj: updated!

static/js/utils/getHHConfigAttributes.js
cea2aj: switched to data attributes so unit tests are no longer necessary

layouts/html.hbs
```diff
@@ -123,9 +118,36 @@
...
+            updateRuntimeConfig(config);
+            clearTimeout(runtimeConfigNotProvidedTimeout);
+            initAnswers();
+          };
+        {{else}}
+          initAnswers();
+        {{/if}}
       });
+      const answersMetadata = {
+        hasInitialized: false
```
oshi97, line 149: would it be out of scope to refactor the window.iframeResizer logic into some static js file? maybe we can also do something about passing RuntimeConfig over the window?

cea2aj: Can you give a little more info on how that would work?

tmeyer2115 (commented):
I think it's great that we have an actual domain model for `RuntimeConfig`. It's a first-class thing in the Theme, so it definitely deserves it. I'm wondering if we should add some other class(es) around managing it. I'm a bit wary of adding even more complexity to the `layouts/html.hbs` partial. There's a lot going on in there already. 

I think anything we can do to factor some of the complexity of run-time configuration into other files/classes would be good. I know this is kind of hand-wavy. I will try and come up with some more concrete ideas. Maybe we have some kind of `RuntimeConfigDirector` that does de-duping of updates and handles message passing to frames. 

In general, if we have complex JS, I think we should try and put as much of it as we can into static. By putting things in `static/js` we can manage complexity better with separate classes, etc. Complicated JS in partials is difficult to maintain. Historically, it's what we've done, but we didn't always have time to do things the "best way".

cea2aj:
> I think it's great that we have an actual domain model for `RuntimeConfig`. It's a first-class thing in the Theme, so it definitely deserves it. I'm wondering if we should add some other class(es) around managing it. I'm a bit wary of adding even more complexity to the `layouts/html.hbs` partial. There's a lot going on in there already.
> 
> I think anything we can do to factor some of the complexity of run-time configuration into other files/classes would be good. I know this is kind of hand-wavy. I will try and come up with some more concrete ideas. Maybe we have some kind of `RuntimeConfigDirector` that does de-duping of updates and handles message passing to frames.
> 
> In general, if we have complex JS, I think we should try and put as much of it as we can into static. By putting things in `static/js` we can manage complexity better with separate classes, etc. Complicated JS in partials is difficult to maintain. Historically, it's what we've done, but we didn't always have time to do things the "best way".

That makes sense. I will experiment with putting this logic in HitchhikerJS

layouts/html.hbs
oshi97: sorry I didn't mean to make it sound like it was easy, and didn't have really concrete plans in mind. I was imagining moving this DOMContentLoaded listener into a static file (the on-document-loaded partial would probably stay here. we can just make them separate listeners). For the RuntimeConfig, instead of getting it from the window in core.hbs, maybe we could have a HitchhikerJS.getRuntimeConfig() ? I also thought we might want to refactor window.iFrameResizer into it's own file, but looking at it again that can probably wait.

cea2aj: I updated the new code to be inside static files rather than in the handlebars. I will leave the refactoring of the window.iFrameResizer to another PR since that is out of scope here. Changing it to HitchhikerJS.getRuntimeConfig() would also change the public interface for accessing the RuntimeConfig. Maybe I should check with product on that? Personally I think that accessing 'window.RuntimeConfig.linkTarget` is a little simpler than 'HitchhikerJS.getRuntimeConfig().linkTarget', since Hitchhikers will be referencing it from inside the cards.

layouts/html.hbs
```diff
@@ -136,13 +134,19 @@
...
           }
-          window.isOverlay && window.Overlay.onMessage(message);
         }
       };
+      {{#if global_config.initializeManually}}
+        const initializeManually = true;
+      {{else}}
+        const initializeManually = false;
+      {{/if}}
+      const experienceInitializer = new HitchhikerJS.ExperienceInitializer(initializeManually).setup();
```
tmeyer2115, line 149: Do we only need an `experienceInitializer` in the `initializeManually` case or no?

static/js/experience-initializer.js
```diff
@@ -0,0 +1,60 @@
...
+      })
+    }
+  }
+
+  /**
+   * Executes the provided function on DOM load
+   * @param {Function} cb 
+   */
+  _fireOnDomLoad (cb) {
+    if  (document.readyState == 'loading') {
```
oshi97, line 31: nit extra space

static/js/experience-initializer.js
```diff
@@ -0,0 +1,60 @@
...
+    this._initializeManually = initializeManually;
+    this._hasAnswersInitialized = false;
+  }
+
+  /**
+   * Initializes on DOM load, or puts an init function on the window if initializeManually is true
+   */
+  setup () {
+    if (this._initializeManually) {
+      window.initAnswersExperience = this._generateInitFunction();
```
tmeyer2115, line 17: Does the iFrame integration end up invoking this as well?

oshi97 (commented):
besides nit lgtm!

static/js/iframe-prod.js
```diff
@@ -1,5 +1,15 @@
...
+const runtimeConfig = new RuntimeConfig();
+window.setRuntimeConfig = (key, value) => {
+  runtimeConfig.set(key, value);
+  sendToIframe({ runtimeConfig: runtimeConfig.getObject() });
+}
+
+const prodDomain = new InjectedData().getProdDomain();
+generateIFrame(prodDomain, runtimeConfig);
+
+window.initAnswersExperienceFrame = generateInitAnswersFrame(runtimeConfig);
```
tmeyer2115, line 15: Here and in the other iFrame script, are we waiting for them to set the `runtimeConfig` before calling `initAnswers`? If so, it's a little hard to follow how that's happening. For my own edification, could you explain that flow?

tmeyer2115 (commented):
This is definitely much more stream-lined! Having these additional classes is really helpful. It allows us to compartmentalize the complexity better.

layouts/html.hbs
cea2aj: Currently it's used for both automatic and manual initialization. I could re-work it so that the automatic initialization occurs in layouts/html.hbs, and the experience initializer is just for initializeManually. (I would also probably update the naming). That might be a better separation of concerns

static/js/experience-initializer.js
cea2aj: Yes, if the RuntimeConfigReceiver gets a message from the parent with initAnswersFrame set to true, then it will call the same initAnswersExperience function

static/js/iframe-prod.js
cea2aj: Ah, after refactoring the 'generateInitAnswersFrame' naming is no longer entirely accurate. I will update it. What happens is that the iframe is always generated immediately. If there were any data attributes set, those will be included in the runtime config during initialization. Next, the `initAnswersExperienceFrame` function is put on the window which updates the runtime config and sends the message to the iframe that the Answers experience should initialize

layouts/html.hbs
tmeyer2115: Got it, I'm fine either way!

tmeyer2115 (approved):

static/js/manual-initializer.js
```diff
@@ -0,0 +1,52 @@
...
+        'An initialization function has not been called within 5 seconds of page load, and "initializeManually" is set to true.\n' +
+        'Load the experience by calling initAnswersExperience() or initAnswersExperienceFrame() for iframe integrations.'
+      );
+    }, 5000);
+    return runtimeConfig => {
+      this._fireOnDomLoad(() => {
+        updateRuntimeConfig(runtimeConfig);
+        clearTimeout(runtimeConfigNotProvidedTimeout);
+        if (!this._hasAnswersInitialized) {
+          initAnswers();
```
oshi97, line 34: could we pass in initAnswers into the constructor?

cea2aj: sure, that's a nice way of reducing the reliance on globals

cea2aj: updated to pass in the init function

oshi97 (approved):
lgtm! 🐹 🐹 🐹  happy about all the progress

PR #778, opened 2021-05-12
fix acceptance tests false positives

oshi97:
Previously, the acceptance test github check would always pass, even
if tests failed. There currently is some issue with how we load mapbox-gl
and firefox:headless, which would cause a "Failed to initialize WebGL."
error, which only occurs for firefox:headless, and not chrome:headless
or regular firefox.

As a short term fix we now run firefox in browserstack.

J=SLAP-1319
TEST=auto

see that the process will exit with 1 (by checking `$?` in terminal) if
I throw an error inside of acceptance/index.js's `try` block, and also
that the github acceptance test check will report a failure

see that the acceptance test check will report a failure if any tests fail, and 
exit with status code 1

### DIFF
--- a/tests/acceptance/index.js
+++ b/tests/acceptance/index.js
@@ -27,10 +27,7 @@ async function runTests (browsers) {
       .src('tests/acceptance/suites/*.js')
       .browsers(browsers)
       .startApp(`npx serve -p ${PORT} test-site/public`, 4000)
-      .run({
-        quarantineMode: true,
-        pageLoadTimeout: 25000
-      });
+      .run({ quarantineMode: true });
     if (result > 0) {
       process.exit(1);
     }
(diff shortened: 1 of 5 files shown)


tests/acceptance/index.js
```diff
@@ -23,11 +23,14 @@ runTests(argv.browsers);
 async function runTests (browsers) {
   const testcafe = await createTestCafe();
   try {
-    await testcafe.createRunner()
+    const result = await testcafe.createRunner()
       .src('tests/acceptance/suites/*.js')
       .browsers(browsers)
       .startApp(`npx serve -p ${PORT} test-site/public`, 4000)
       .run({ quarantineMode: true });
+    if (result > 0) {
```
cea2aj, line 31: nit: Is result the exit code of the runner? If so, I think we could exit with that code so that we don't lose that info. If not, this works!

oshi97: result is the number of failed tests here!

oshi97: ah I should rename this

cea2aj (approved):

PR #780, opened 2021-05-13
Update Theme dependencies according to Dependabot's suggestions.

tmeyer2115:
Dependabot suggested we update the following dependencies:
- hosted-git-info from 2.8.8 to 2.8.9.
- postcss from 8.1.4 to 8.2.10.
- lodash from 4.17.20 to 4.17.21.
- grunt from 1.1.0 to 1.3.0.

TEST=manual

Built a local Jambo site and ensured the served site looked as expected. Also
verified that the Grunt watch task still worked.


oshi97 (approved):
lgtm!

PR #782, opened 2021-05-19
Do not run the `babel` helper in Development Mode.

tmeyer2115:
This PR adds a check to the `babel` helper to see if we are currently in the Development Mode. If
so, the enclosed source code is not run through Babel, it is returned verbatim. This is a preview
optimization, since the Babel-ification is a relatively slow operation. It's not necessary for
preview either, assuming devs are not previewing their changes in IE11.

There will need to be a corresponding Jambo PR for this as well. Jambo still has a `babel` helper
of its own for legacy reasons.

J=SLAP-1312
TEST=manual

Ensured that when in Development mode, the helper was a simple pass-through. In Production mode,
the enclosed code was Babel-ified.

### DIFF
--- a/hbshelpers/babel.js
+++ b/hbshelpers/babel.js
@@ -3,7 +3,7 @@ const { transformSync } = require('@babel/core');
 /**
  * Babel-ifies the enclosed code, using the config below. Note that if the 
  * IS_DEVELOPMENT_PREVIEW environment variable is set to 'true', the enclosed
- * code is return verbatim. This is an optimization for the build preview. 
+ * code is returned verbatim. This is an optimization for the build preview. 
  * 
  * @param {import('handlebars').HelperOptions} options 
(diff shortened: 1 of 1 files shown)


hbshelpers/babel.js
```diff
@@ -1,24 +1,31 @@
 const { transformSync } = require('@babel/core');
 
 /**
- * Babel-ifies the enclosed code, using the config below.
+ * Babel-ifies the enclosed code, using the config below. Note that if the 
+ * IS_DEVELOPMENT_PREVIEW environment variable is set to 'true', the enclosed
+ * code is return verbatim. This is an optimization for the build preview.
```
cea2aj, line 6: nit: I think this should read "is returned verbatim"

cea2aj (approved):

PR #783, opened 2021-05-19
Optimize `bundle.css` with `css-nano`. 

tmeyer2115:
This PR adds the `postcss-loader` and `css-nano` to the Theme's Webpack configuration. This addition allows `css-nano` to perform a number of optimizations and minimizations on the `bundle.css` that we output. I saw the size of this asset decrease by about 10 KB on a local test site.

J=SLAP-1299
TEST=auto, manual

Ensured the Percy tests passed after the change. Smoke tested a local test site and verified that the styling looked as expected.


static/webpack-config.js
```diff
@@ -91,6 +92,16 @@ module.exports = function () {
           use: [
             MiniCssExtractPlugin.loader,
             'css-loader',
+            {
+              loader: "postcss-loader",
```
oshi97, line 96: nit: can we use single quotes here

oshi97 (approved):

PR #786, opened 2021-05-20
Fix locator bundle race condition

cea2aj:
Fix the race condition which occurs when the search query loads before the locator bundle

This fix ensures that the locator bundle executes after Answers loads, but before Answers initializes. The race condition issue was first found on Theme 1.20.1, and it caused the cards to break when the page loaded. I put a fix up in #739, which stopped the page from breaking, but it did not fix the underlying cause of the race condition. This PR fixes the root problem by making the execution order of the javascript consistent.

The race condition lead to an issue where two queries were ran on page load. This became a problem in a techops (Zendesk 405748) because the page would load and return no results while being zoomed into Kansas. This PR prevents the double queries from being sent if the locator bundle takes a while to load. There still may be some situations where two queries are sent on page load, but those issues are out of scope of this particular issue.  I also zoomed out the map by default so that if no default initial search is supplied, the map will show the entire US rather than be zoomed into Kansas.

J=SLAP-1329
TEST=manual

Test both Google Maps and Mapbox. Use fiddler to artificially delay the locator bundle. Test iframe and non-iframe experiences.


script/core.hbs
```diff
@@ -7,6 +7,14 @@
 <script src="https://assets.sitescdn.net/answers/v{{global_config.sdkVersion}}/{{#if params.sdkLocaleOverride}}{{#unless (eq params.sdkLocaleOverride 'en')}}{{params.sdkLocaleOverride}}-{{/unless}}{{else if global_config.locale}}{{#unless (eq global_config.locale 'en')}}{{global_config.locale}}-{{/unless}}{{/if}}answerstemplates.compiled.min.js" defer></script>
 <script>
 {{#babel}}
+  function onAnswersScriptLoad () {
+    const waitBeforeInitPromises = [
+      new Promise(HitchhikerJS.pageSpecificScriptsLoaded)
+    ];
+    Promise.all(waitBeforeInitPromises).then(() => {
```
cea2aj, line 14: I'm using Promsie.all because this could be helpful for the dynamic config/JWT stuff. But if it's premature optimization then I can change this to just wait for the single promise.

script/core.hbs
```diff
@@ -7,6 +7,16 @@
 <script src="https://assets.sitescdn.net/answers/v{{global_config.sdkVersion}}/{{#if params.sdkLocaleOverride}}{{#unless (eq params.sdkLocaleOverride 'en')}}{{params.sdkLocaleOverride}}-{{/unless}}{{else if global_config.locale}}{{#unless (eq global_config.locale 'en')}}{{global_config.locale}}-{{/unless}}{{/if}}answerstemplates.compiled.min.js" defer></script>
 <script>
 {{#babel}}
+  function onAnswersScriptLoad () {
+    const waitBeforeInitPromises = [
```
tmeyer2115, line 11: I think we should come up with a more descriptive name here. Maybe `preAnswersInitTasks` or something like that?

static/js/page-specific-scripts-loaded.js
```diff
@@ -0,0 +1,17 @@
+/**
```
tmeyer2115, line 1: Should we name this file/function more generally? Your plan was that this could be used for more than just script tag loading, right? People could define any kind of async task that needs to happen before the `Answers.init`

script/core.hbs
cea2aj: ya, I like that better

script/core.hbs
```diff
@@ -86,3 +94,5 @@
   locale=(findFirst params.sdkLocaleOverride global_config.locale 'en')
   sdkVersion=global_config.sdkVersion
 }}
+
+{{> script/partials/page-specific-scripts pageTemplate=pageTemplate}}
```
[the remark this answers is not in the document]
cea2aj: For 1 and 2, the locator bundle needs to execute after the SDK loads since the locator components are answers components. That's not necessary for the HitchhikerJS bundle because the cards are are put into a BaseCard object which isn't executed until the Answers onReady.

I like idea 3. I'll try that out. I think it might work well with the runtime config stuff too.

static/js/page-specific-scripts-loaded.js
cea2aj: To use this functionality more generally for async tasks, new promises would be added to the `preAnswersInitTasks` array. This function could still be specific to page specific scripts.

cea2aj: this is outdated due to switching to use DOMContentLoaded

script/core.hbs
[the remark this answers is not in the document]
cea2aj: Updated! DOMContentLoaded work as well and is much simpler!

script/core.hbs
```diff
@@ -86,3 +86,7 @@
   locale=(findFirst params.sdkLocaleOverride global_config.locale 'en')
   sdkVersion=global_config.sdkVersion
 }}
+
+{{#ifeq pageTemplate 'vertical-full-page-map'}}
```
cea2aj, line 90: What do you all think about putting this logic here (rather than a page specific scripts partial)?

oshi97: I like being able to see all the scripts tags in one file, but it's also a little annoying for core.hbs to have to "know" the page template and manage that, when we can just put that in the page.html.hbs. Maybe we can come up with some kind of solution in a different pr?

cea2aj: Now that we're using DOMContentLoaded, I think I can move the script back to the page template. Does that sound good for this PR?

oshi97: yeah sgtm! I think I like the idea of centralizing the scripts but not sure how we'd want to do it

tmeyer2115 (approved):

templates/vertical-full-page-map/page.html.hbs
```diff
@@ -1,6 +1,8 @@
 {{#> layouts/html pageWrapperCss="YxtPage-wrapper--mobileListView" }}
   {{#> script/core }}
     {{> cards/all }}
+    {{> theme-components/theme-map/script}}
```
oshi97, line 4: did we just forget these before?

oshi97: oh I took a quick look and I think they already exist in page-setup?

cea2aj: This PR moves them from page setup to here

oshi97: can we move them back into page-setup? Otherwise preexisting vertical-full-page-map pages that don't have these partials will break on upgrade.

cea2aj: sure

cea2aj: updated!

oshi97 (approved):
lgtm!

PR #790, opened 2021-05-24
Allow any SDK branch to be specified in `global_config.sdkVersion`.

tmeyer2115:
This PR allows any SDK branch, not just official releases, to be specified in `global_config.sdkVersion`. To support this, a new HBS helper was added to the Theme: `sdkAssetUrl`. This helper takes in the branch (or release version), locale, and asset name, and provides the correct URL for it. With the addition of this helper, we were able to remove the `sdk-url` partial. The logic for computing the URL is now sufficiently complicated that we should do it in JS instead of HBS. 

J=SLAP-924
TEST=auto, manual

Added unit tests for the new helper. For the following scenarios, I tested with both an English and French locale, and verified the correct asset URLs were provided:
- Released SDK version.
- Release branch of the SDK.
- Feature branch of the SDK.
- Dev branch of the SDK.


tests/script/partials/sdk-js-script-tags.js
```diff
@@ -24,11 +24,6 @@ const jaOutput = `<script
   defer
 ></script>`;
 
-it('creates correct tags when no locale specified', () => {
-  const partialUsage = '{{> script/partials/sdk-js-script-tags sdkVersion=1.8}}'
```
oshi97, line 28: do we ever have a case where locale is not set (maybe when there's no locale_config.json)?

hbshelpers/sdkAssetUrl.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  const isPreReleaseBranch = 
+    RELEASE_BRANCH_REGEX.test(branch) || HOTFIX_BRANCH_REGEX.test(branch);
+  const isLocalizationSupported = 
+    (isReleasedBranch || isPreReleaseBranch) && 
+    !(locale.startsWith('en') || assetName === 'answers.css') ;
+  
+  const parsedAssetName = isLocalizationSupported ?
+    `${locale}-${assetName}` :
```
oshi97, line 31: would we ever want to just always slap the `{locale}-` prefix if the locale is not en, regardless of whether they're using a release/hotfix branch? Right now, if somebody asks for a french SDK when it doesn't exist, we'll give them an english SDK, which I think is better but curious if you had any thoughts?

hbshelpers/sdkAssetUrl.js
```diff
@@ -0,0 +1,35 @@
+const RELEASE_BRANCH_REGEX = /^release\/v[0-9.]+$/;
+const HOTFIX_BRANCH_REGEX = /^hotfix\/v[0-9.]+$/;
+const SEM_VER_REGEX = /^[1-9]+$|^[1-9]+\.[0-9]+$|^[1-9]+\.[0-9]+\.[0-9]+$/;
```
cea2aj, line 3: are we intentionally leaving out SDK versions before 1.0 because nobody should be using them anyway?

tests/script/partials/sdk-js-script-tags.js
tmeyer2115: Right now, whenever we invoke this partial, the `locale` is always set. If it's not in the `locale_config` or `global_config`, we simply use `en`.

hbshelpers/sdkAssetUrl.js
tmeyer2115: We don't support pre-1.0 versions any more.  don't think they would work with the Theme.

oshi97 (approved):

PR #792, opened 2021-05-25
transition node-sass to dart-sass

yen-tt:
Update package.json to use sass packae and remove node-sass in static folder
sass-loader will resolve the implementation based on depedencies and use sass package

J=SLAP-1310
TEST=compile


oshi97 (approved):
lgtm! did you notice any compilation speed changes before and after the change?

PR #796, opened 2021-05-26
Resolved deprecation warnings from inlineAssetHtmlPlugin and webpack

yen-tt:
update packages

- Updated html-webpack-plugin, used by inlineAssetHtlmPlugin, to be more integrated with webpack v5, which got rid of the deprecated warnings from older v4 with asset and public path.
- Since inlineAssetHTMLPlugin only access compilation.assets without modification, no changes are needed in the js file. Instead, updated webpack version to remove compilation asset warning.

J=SLAP-1307
TEST=manual

Compiled successfully without deprecation warnings. Launched test site and performed searches and navigations without error.

### DIFF
--- a/static/package.json
+++ b/static/package.json
@@ -29,7 +29,7 @@
     "grunt-contrib-watch": "^1.1.0",
     "grunt-webpack": "^4.0.0",
     "html-loader": "^1.1.0",
-    "html-webpack-plugin": "^4.3.0",
+    "html-webpack-plugin": "^5.3.1",
     "jambo": "^1.11.0",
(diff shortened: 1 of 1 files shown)


tmeyer2115 (commented):
Looks good! Could we do a bit more testing than just compilation though? Since we upgrade the `html-webpack-plugin` to a new major version, I think we should test the generated site as well. It's possible compilation could succeed but something is off with the generated pages. Just do a simple smoke test: run a few searches, navigate around, etc.

tmeyer2115 (approved):

oshi97 (approved):

PR #797, opened 2021-05-27
add static placeholders for the searchbar and nav components

oshi97:
This component adds static html placeholders for the searchbar
and nav components, which will be displayed prior to the SDK
loading and components being added to the page.

J=SLAP-1336
TEST=manual

see that on each of the 5 page templates, the searchbar and nav
placeholders will display for a brief second before the page finishes
loading


tmeyer2115 (approved):

cea2aj:
this is really cool

PR #798, opened 2021-05-27
remove font-awesome and bootstrap-sass deps

oshi97:
These two dependencies provide sass variables
and mixins that are not used anywhere in the code.
In addition to that, after upgrading to dart-sass,
we now get deprecation warnings for them because
they still use the division operator `/` instead of
math.div

see https://sass-lang.com/documentation/breaking-changes/slash-div

I believe the package-lock was out of date due to some specific way
previous commits were merged.

J=SLAP-1348
TEST=manual

test that dart-sass will throw an error if you try to use a mixin
or variable that is not defined (tested with the bootstrap and
font-awesome vars/mixins) since no error is thrown, that further
verifies that we are not using them anywhere in our code

ran a build


cea2aj (approved):
🧼

PR #799, opened 2021-05-27
upgrade deprecated scss code (slash division and elseif)

oshi97:
Slash division and elseif are deprecated operations in sass.
Right now we're getting some warnings that might scare
hitchhikers. These deprecation warnings popped up 
after upgrading from node-sass to dart-sass.

see: https://sass-lang.com/documentation/breaking-changes/slash-div

J=SLAP-1348
TEST=manual

ran a build, saw reduction in warnings


cea2aj (approved):
looks good

PR #802, opened 2021-05-28
Update runtime config public interface

cea2aj:
Update the public interface of the runtime config in accordance with the product spec

- All new functionality on the window is namespaced to AnswersExperience or AnswersExperienceFrame for iframe integrations
- These namespaces each have an init function and a runtime config object
- The runtime config object has public set get, and a getAll methods
- The 'initAnswersExperience' boolean has been moved out the runtime config. This change lead me to create the iframe message queue to prevent messages from being sent and lost before the iframe has initialized
- I also removed the lastUpdated logic. It was created for an older design where the entire runtime config object was being updated at once, but it's no longer necessary since the runtime config is now updated one key value pair at a time.
- Remove the data attributes feature since we currently can't guarantee the run config will be set by the time answers initializes

J=SLAP-1362
TEST=manual

On both iframe and non-iframe experiences test initializing the experience with custom config. Test using the runtime config data models. Test that calling set on the parent runtime config will update the runtime config within the iframe. Test initialization with JWT.


static/entry.js
```diff
@@ -35,4 +35,9 @@ import ManualInitializer from './js/manual-initializer';
 export { ManualInitializer };
 
 import RuntimeConfigReceiver from './js/runtime-config-receiver';
-export { RuntimeConfigReceiver };
\ No newline at end of file
+export { RuntimeConfigReceiver };
+
+import RuntimeConfig from './js/runtime-config';
+window.AnswersExperience = {
```
tmeyer2115, line 41: If we have this here, do we need the `window.AnswersExperience = {};` in `layouts/html.hbs`?

static/js/iframe-common.js
```diff
@@ -91,8 +94,12 @@ export function generateIFrame(domain, runtimeConfig) {
   iFrameResize({
     checkOrigin: false,
     onInit: function() {
+      iframeInitialized = true;
+      iframeMessageQueue.forEach(message => {
+        sendToIframe(message);
+      })
       runtimeConfig && sendToIframe({
```
tmeyer2115, line 101: If there's a `runtimeConfig`, could we added it do the `iframeMessageQueue`? Then we wouldn't need this special case here, we could use the message queue loop above.

static/js/iframe-staging.js
```diff
@@ -1,15 +1,17 @@
...
-}
+const answersDataAttributes = document.querySelector('#answers-container')?.dataset;
+const runtimeConfig = new RuntimeConfig(answersDataAttributes);
+runtimeConfig._onUpdate(updatedConfig => {
+  sendToIframe({ runtimeConfig: updatedConfig });
+});
+window.AnswersExperienceFrame = {
+  runtimeConfig: runtimeConfig,
+  init: generateInitAnswersExperienceFrameFunction(runtimeConfig)
+};
```
tmeyer2115, line 14: This logic block is the same between iframe-staging and iframe-prod. Could we refactor it such that it lives in iframe-common?

static/js/utils/generateInitAnswersExperienceFrameFunction.js
```diff
@@ -1,8 +1,7 @@
 import { sendToIframe } from '../iframe-common';
 
 /**
- * Returns a function which sends a message to the iframe to initialize Answers
- * with the provided runtimeConfig
+ * Returns a function which updates the runtime config and initializes answers
```
tmeyer2115, line 4: Small nit, but could we have a more compact name for this function? It's a bit of a mouth full

static/js/iframe-staging.js
cea2aj: what do you think about making AnswersExperienceFrame a simple class?

static/entry.js
cea2aj: nope. removing

static/js/iframe-common.js
cea2aj: sure, I like that approach

static/js/iframe-staging.js
cea2aj: Updated to put this logic into AnswersExperienceFrame

static/js/utils/generateInitAnswersExperienceFrameFunction.js
cea2aj: Due to the refactor which added the AnswersExperienceFrame class, this method is no longer needed

tmeyer2115 (approved):
LGTM!

PR #803, opened 2021-05-28
Dynamic link targets

cea2aj:
Use link targets from runtime config, and default to "_top" if it is not specified

Update all targets which include card title links, CTAS, phone links, RTF links, and direct answer links

J=SLAP-1349
TEST=manual

Load the test site, change the runtime config link target and see different link target behavior. Test title links, CTAs, phone links, RTF links, and direct answer links

### DIFF
--- a/cards/event-standard/component.js
+++ b/cards/event-standard/component.js
@@ -12,10 +12,12 @@ class event_standardCardComponent extends BaseCard['event-standard'] {
    * @param profile profile of the entity in the card
    */
   dataForRender(profile) {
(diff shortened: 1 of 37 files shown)


tmeyer2115 (approved):

oshi97:
it looks like the percy snapshots broke?

cea2aj:
> it looks like the percy snapshots broke?

ya, taking a look now

tmeyer2115 (approved):

PR #804, opened 2021-06-01
Fix iframe page navigation when initializeManually=true

cea2aj:
Ensure that the page initializes upon page navigation when initializeManually=true

This functionality broke during a recent re-factor when I moved the initAnswersExperience boolean out of the runtimeConfig and into the top-level of the message. Since this data was previously inside the runtimeConfig, it would persist and the experience would receive the message to re-initialize during page navigation. After moving that data outside of the runtimeConfig, the page would no longer receive that message and would fail to re-initialize. To fix this, I ensure that the parent indicates to the iframe that it needs to re-initialize.

J=none
TEST=manual

Load an iframe test site with manuallyInitialize=true, and call the init function. See that the experience now initializes when navigating verticals within the iframe.

### DIFF
--- a/static/js/answers-experience-frame.js
+++ b/static/js/answers-experience-frame.js
@@ -3,6 +3,7 @@ import { sendToIframe } from './iframe-common';
 export default class AnswersExperienceFrame {
   constructor (runtimeConfig) {
     this.runtimeConfig = runtimeConfig;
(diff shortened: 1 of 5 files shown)


cea2aj:
I've created SLAP-1371 to add unit tests for the feature branch as a whole

oshi97 (approved):

PR #805, opened 2021-06-01
fix google streetview pegman on universal

oshi97:
A previous commit fixed this on the vertical-map
page template. We also need a fix for the map
on universal, which also uses the SDK's map component.

T=409037
TEST=manual, auto

see mr.pegman on universal
update the test-site's universal config to display a google
maps component

### DIFF
--- a/test-site/config/index.json
+++ b/test-site/config/index.json
@@ -36,7 +36,10 @@
     },
     "KM": {
       "universalSectionTemplate": "grid-two-columns",
-      "cardType": "multilang-financial-professional-location"
+      "cardType": "multilang-financial-professional-location",
+      "mapConfig": {
+        "mapProvider": "Google"
+      }
     }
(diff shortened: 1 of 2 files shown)


test-site/config/index.json
```diff
@@ -36,7 +36,10 @@
     },
     "KM": {
       "universalSectionTemplate": "grid-two-columns",
-      "cardType": "multilang-financial-professional-location"
+      "cardType": "multilang-financial-professional-location",
+      "mapConfig": {
+        "mapProvider": "Google"
```
cea2aj, line 41: nit: Is switching this to google intentional or just for testing?

oshi97: just for testing, before this we didn't display a map on universal at all

cea2aj (approved):

PR #806, opened 2021-06-01
Filter component and footer collision on Desktop View

yen-tt:
Update template layouts to have filter with relative position

- Previously, filter in resultsWrapper has position as absolute, which remove the component the normal flow of the document. This caused overlap issue with any component placed towards the bottom of the document.
- Revised layout of vertical-grid and vertical-standard template to organize resultsWrapper into two section (sidebar and mainContent). Placed filter component on sidebar div and answersResults component on maincontent div. Updated scss files to center results on the page.
- For mobile view (when collapsibleFilters setup is used), to maintain the same UX as before, the FiltersWrapper div must be moved from the sidebar div to maincontent div right under ResultsHeader div. To do so, onCreate function for FiltersLink component will move FiltersWrapper element to maincontent div if the viewport pixel is less than `$breakpoint-expanded-filters`. a media query listener is also added on `$breakpoint-expanded-filters` to make sure FiltersWrapper element is in the correct div column when the page viewport changes.
  - `cssBreakpointExpandedFilters: $breakpoint-expanded-filters;` is exported to use in media query listener
  - [MediaQueryListener.addListener()](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener) is deprecated. If possible, MDN recommend to use .addEventListener instead

J=SLAP-1221
TEST=manual

Launched test site and smoke test filters on /people page in collapsible and non-collapsible mode (desktop/mobile view) and in vertical-grid and vertical-standard layout in chrome, safari, firefox, and IE.
See that filters and footer no longer overlap


templates/vertical-standard/page.html.hbs
```diff
@@ -1,54 +1,63 @@
...
-    {{> templates/vertical-standard/script/navigation }}
-    {{> templates/vertical-standard/script/verticalresults }}
-    {{> templates/vertical-standard/script/pagination }}
-    {{> templates/vertical-standard/script/locationbias }}
-    {{!-- {{> templates/vertical-standard/script/sortoptions }} --}}
-    {{!-- {{> templates/vertical-standard/script/facets }} --}}
-    {{!-- {{> templates/vertical-standard/script/filterbox }} --}}
-    {{!-- {{> templates/vertical-standard/script/qasubmission }} --}}
+    {{!-- {{> templates/vertical-grid/collapsible-filters/page-setup }} --}}
+    {{> templates/vertical-grid/script/appliedfilters }}
```
tmeyer2115, line 5: Was the change to use the `vertical-grid` templates here intentional? Each page template has its own set of partials in the `templates/{pageTemplate}` directory. The page template should use those partials, as well as some shared, common ones (such as `script/core`). If page templates refer to partials from other templates, that could make things difficult for the HHs. If a HH were to override one of the `templates/vertical-grid` partials, it would now impact both `vertical-grid` and `vertical-standard` pages. This would surprise them, as they'd expect just the pages built with the `vertical-grid` layout to change.

templates/vertical-grid/page.html.hbs
```diff
@@ -26,25 +26,33 @@
...
+        </div>
+        
+        <div class="Answers-results js-answersResults">
+          {{> templates/vertical-grid/markup/spellcheck }}
+          {{> templates/vertical-grid/markup/verticalresults }}
+          {{> templates/vertical-grid/markup/pagination }}
+          {{!-- {{> templates/vertical-grid/markup/qasubmission }} --}}
+        </div>
       </div>
+      <div class="Answers-resultsRightContainer">
```
tmeyer2115, line 53: Is the `resultsRightContainer` used or a future optimization?

static/scss/answers/templates/vertical-standard.scss
```diff
@@ -23,6 +30,7 @@
 
   &-filtersWrapper {
     margin-bottom: 20px;
+    float: right;
```
cea2aj, line 33: what are your thoughts on using flexbox for this as well? I think justify-content flex-end might work

yen-tt: hmm, I used float since it only deal with positioning a small section without much big layout changes. But instead of just doing this for the filter, I think it will be better to have the left container use flexbox and justify-content flex-end for more consistent scss layout. Updated in the new commit. Thanks for the suggestion!

templates/vertical-standard/page.html.hbs
yen-tt: sorry, it was a mishap switching between files for the templates. I updated and retested the changes!

static/scss/answers/templates/vertical-grid.scss
```diff
@@ -57,12 +57,25 @@
       position: relative;
     }
 
+    &-resultsLeftContainer {
+      flex: 1;
```
oshi97, line 61: if we're just setting one flex property would it make sense to just set it directly (or to specify all 3 values)? I vaguely remember reading something against using the flex shorthand like this https://mastery.games/post/flex-shorthand/ (not this article but similar) also some weirdness with ie11? (https://stackoverflow.com/questions/21600345/flexbox-and-internet-explorer-11-displayflex-in-html)

templates/vertical-grid/page.html.hbs
yen-tt: It helps with centering the mid container that displays answersResults. Currently it's just empty but I think organizing it this way would also be useful if we decide to add anything to the right in the future

static/scss/answers/collapsible-filters/collapsible-filters-templates.scss
```diff
@@ -45,6 +45,11 @@
 .AnswersVerticalGrid.CollapsibleFilters {
   @include CollapsibleFiltersOverrides;
 
+  .Answers-resultsWrapper {
+    width: 100%;
+    max-width: none;
```
oshi97, line 50: do we need to specify max-width: none here (which I think is the default value?)

oshi97:
oh can we also double check that things look good in ie11? Percy doesn't cover us there unfortunately

static/scss/answers/collapsible-filters/sdk-filter-overrides.scss
```diff
@@ -3,9 +3,11 @@
  $container-width-at-collapsed-filters-breakpoint: $breakpoint-collapsed-filters - (2 * $filters-width);
 
 @mixin CollapsibleFiltersOverrides {
-  .Answers-container {
-    @media (min-width: #{$breakpoint-expanded-filters}) {
-      width: math.div($container-width-at-collapsed-filters-breakpoint, $breakpoint-collapsed-filters) * 100%;
+  .Answers {
+    &-container, &-resultsMidContainer {
+      @media (min-width: #{$breakpoint-expanded-filters}) {
+        width: $container-width-at-collapsed-filters-breakpoint / $breakpoint-collapsed-filters * 100%;
```
cea2aj, line 9: can we use math.div instead so that we don't get a webpack warning about the slash?

static/scss/answers/collapsible-filters/collapsible-filters-templates.scss
yen-tt: The max-width is set in theme.scss for all answers-container so I would need to override it here

static/scss/answers/templates/vertical-grid.scss
yen-tt: That make sense. Since the default value and the interpretation of flex one may be different per browser, I will set it explicitly in the flex property as 1 1 0.

templates/vertical-grid/page.html.hbs
yen-tt: after some revision, the right div is no longer necessary. It's removed.

templates/vertical-grid/collapsible-filters/script/filterlink.js
```diff
@@ -11,5 +11,22 @@ ANSWERS.addComponent('FilterLink', {
     CollapsibleFilters.Helpers.clearSearch();
     collapsibleFiltersInteractions.focusSearchBar();
   },
+  onCreate: () => {
+    const moveFiltersWrapper = (containerClassName) => {
+      document.getElementsByClassName(containerClassName)[0].appendChild(
```
cea2aj, line 16: Can we handle the case when `getElementsByClassName` returns an empty array?

templates/vertical-standard/collapsible-filters/script/filterlink.js
```diff
@@ -11,5 +11,22 @@ ANSWERS.addComponent('FilterLink', {
...
+    if (collapsibleFiltersInteractions.isCollapsibleFiltersView()) {
+      moveFiltersWrapper('Answers-resultsMidContainer');
+    }
+    collapsibleFiltersInteractions.setListenerOnExpandedFiltersBreakpoint((e) => {
+      if (e.matches) {
+        moveFiltersWrapper('Answers-resultsLeftContainer');
+      } else {
+        moveFiltersWrapper('Answers-resultsMidContainer');
+      }
+    });
```
cea2aj, line 29: what do you think of moving all of this logic into `collapsibleFitlersInteractions` so that it doesn't have to be duplicated for each template?

static/js/collapsible-filters/interactions.js
```diff
@@ -240,4 +241,18 @@ export default class Interactions {
       yxtFooter.classList.add(this.templateName);
     }
   }
+
+  isCollapsibleFiltersView() {
+    const mediaQuery = window.matchMedia(`(min-width: ${variables.cssBreakpointExpandedFilters})`);
+    return !mediaQuery.matches;
+  }
+
+  setListenerOnExpandedFiltersBreakpoint(handleChange) {
```
cea2aj, line 250: What do you think of `addExpandedFiltersListener`? Also, can we abstract away the event so that the code which calls this function reads like this:
```
collapsibleFiltersInteractions.addExpandedFiltersListener(isExpanded => console.log(isExpanded));
```

templates/vertical-standard/collapsible-filters/script/filterlink.js
yen-tt: This logic is quite specific to verticalGrid/Standard with its left and mid div, so I'm not sure if it's generic enough to live in `collapsibleFiltersInteractions` since it's not applicable to the rest of the templates

static/js/collapsible-filters/interactions.js
yen-tt: sure thing

templates/vertical-standard/collapsible-filters/script/filterlink.js
yen-tt: eh, after some thoughts, I decided to move it to `collapsibleFiltersInteractions` anyway

static/js/collapsible-filters/interactions.js
```diff
@@ -240,4 +241,56 @@ export default class Interactions {
       yxtFooter.classList.add(this.templateName);
     }
   }
+
+  /**
+   *
```
tmeyer2115, line 246: Nit: extra line here.

static/js/collapsible-filters/interactions.js
```diff
@@ -240,4 +241,56 @@ export default class Interactions {
...
+  /**
+   * Add a media query event listener when there's a width view port change
+   * relative to $breakpoint-expanded-filters.
+   *
+   * @param {Function} handleViewportChange callback function to invoke when there's a media query event
+   */
+  addExpandedFiltersListener(handleViewportChange) {
+    const mediaQuery = window.matchMedia(`(min-width: ${variables.cssBreakpointExpandedFilters})`);
+    if ('addEventListener' in mediaQuery) {
+      mediaQuery.addEventListener('change', (e) => handleViewportChange(e.matches))
```
tmeyer2115, line 266: Nit: Can remove parens around `(e)`

static/js/collapsible-filters/interactions.js
```diff
@@ -240,4 +241,56 @@ export default class Interactions {
...
+     }
+  }
+
+  /**
+   * Update FiltersWrapper HTML element's location when switching between
+   * expanded and collapsible filters view in page templates that follows
+   * the left-mid-right div structure (e.g. VerticalStandard and VerticalGrid).
+   */
+  setupFiltersWrapper() {
+    const moveFiltersWrapper = (containerClassName) => {
```
tmeyer2115, line 278: Again, can remove extra parens

static/js/collapsible-filters/interactions.js
```diff
@@ -240,4 +241,56 @@ export default class Interactions {
...
+      mediaQuery.addListener(e => handleViewportChange(e.matches))
+     }
+  }
+
+  /**
+   * Update FiltersWrapper HTML element's location when switching between
+   * expanded and collapsible filters view in page templates that follows
+   * the left-mid-right div structure (e.g. VerticalStandard and VerticalGrid).
+   */
+  setupFiltersWrapper() {
```
tmeyer2115, line 277: Could we make this name a bit more descriptive? Maybe use the word "responsive" somewhere

templates/vertical-grid/page.html.hbs
```diff
@@ -26,24 +26,29 @@
...
-    <div class="Answers-container Answers-resultsWrapper Hitchhiker-3-columns">
-      {{> templates/vertical-grid/markup/spellcheck }}
-      <div class="Answers-resultsHeader">
-        {{> templates/vertical-grid/markup/verticalresultscount }}
-        {{> templates/vertical-grid/markup/appliedfilters }}
-        {{!-- {{> templates/vertical-grid/collapsible-filters/markup/filterlink }} --}}
-        {{!-- {{> templates/vertical-grid/collapsible-filters/markup/viewresultsbutton }} --}}
+    <div class="Answers-container Answers-resultsWrapper">
+      <div class="Answers-resultsLeftContainer">
+        <!-- Uncomment the following div if you want to include filters, facets, or sort options  -->
```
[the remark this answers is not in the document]
yen-tt: In desktop version, yes the the filters come before applied filters. In mobile view (with collapsible filters), applied filters comes before the filters.

tmeyer2115 (approved):

static/scss/answers/templates/vertical-standard.scss
```diff
@@ -21,6 +21,12 @@
     }
   }
 
+  &-resultsMidContainer {
+    flex-direction: column;
+    flex-grow: 1;
+    width: 100%;
```
oshi97, line 27: is this width 100% necessary? same question above

static/js/collapsible-filters/interactions.js
```diff
@@ -240,4 +241,55 @@ export default class Interactions {
...
+   * relative to $breakpoint-expanded-filters.
+   *
+   * @param {Function} handleViewportChange callback function to invoke when there's a media query event
+   */
+  addExpandedFiltersListener(handleViewportChange) {
+    const mediaQuery = window.matchMedia(`(min-width: ${ScssVariables.cssBreakpointExpandedFilters})`);
+    if ('addEventListener' in mediaQuery) {
+      mediaQuery.addEventListener('change', e => handleViewportChange(e.matches));
+     } else if ('addListener' in mediaQuery) {
+      mediaQuery.addListener(e => handleViewportChange(e.matches));
```
oshi97, line 267: I guess we don't need to worry about when addListener doesn't exist since it's compatible with basically all browsers?

templates/vertical-standard/page.html.hbs
```diff
@@ -29,24 +29,28 @@
...
     {{!-- {{> layouts/overlay-suggestions }} --}}
     <div class="Answers-container Answers-resultsWrapper">
-      {{!-- {{> templates/vertical-standard/markup/directanswer }} --}}
-      {{> templates/vertical-standard/markup/spellcheck }}
-      <div class="Answers-resultsHeader">
-        {{> templates/vertical-standard/markup/verticalresultscount }}
-        {{> templates/vertical-standard/markup/appliedfilters }}
-        {{!-- {{> templates/vertical-standard/collapsible-filters/markup/filterlink }} --}}
-        {{!-- {{> templates/vertical-standard/collapsible-filters/markup/viewresultsbutton }} --}}
+      <div class="Answers-resultsLeftContainer">
```
oshi97, line 32: maybe `Answers-leftSidebar` or something like that? I think calling it results container is a little confusing since it doesn't contain any results

static/scss/answers/templates/vertical-standard.scss
yen-tt: it's needed for IE11. the flexbox behavior for ie is different compare to other browsers. Without this, the card size will spill over to the right of the page

static/js/collapsible-filters/interactions.js
yen-tt: Yea, addListener works for all browser but the more modern ones will use addEventListener (e.g. ie only works for addListener, but chrome have support for addEventListener)

templates/vertical-standard/page.html.hbs
yen-tt: Now that I think about it, we should probably avoid left/right naming since we support rtl language as well. What do you think about `Answers-sidebar` and `Answers-maincontent`

oshi97: That makes sense! I like those names (probably camel case for mainContent?)

PR #807, opened 2021-06-02
fix PinImages and ClusterPinImages not changing pin size

oshi97:
One of the requirements of the vertical-full-page-map is
that users can jambo override the PinImages and ClusterPinImages
files in order to change the pin size. @rosiegrant  says that this
was tested at some point, and since none of this code has been
changed since the initial 8000+ line full page map commit,
it's likely this was dropped somewhere during prototyping.

Talked with @rosiegrant about the interface change.

J=TECHOPS-603
TEST=manual

see that I can resize both the regular pin and the cluster pin,
on both chrome and ie11, for both google and mapbox

check that the default pin/cluster pin sizes did not change before and after this commit

check that pinCount is passed to propertiesForStatus correctly

### DIFF
--- a/static/js/theme-map/ClusterPinImages.js
+++ b/static/js/theme-map/ClusterPinImages.js
@@ -20,8 +20,8 @@ class ClusterPinImages {
    * @param {string} pin.backgroundColor Background color for the pin
    * @param {string} pin.strokeColor Stroke (border) color for the pin
    * @param {string} pin.labelColor Label (text) color for the pin
-   * @param {string} pin.width The width of the pin
-   * @param {string} pin.height The height of the pin
+   * @param {number} pin.width The width of the pin
+   * @param {number} pin.height The height of the pin
    * @param {string} pin.labelText The label text for the cluster pin (normally size of cluster)
    * @return string The SVG of the pin
    */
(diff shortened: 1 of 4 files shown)


static/js/theme-map/ClusterPinImages.js
```diff
@@ -20,29 +20,29 @@ class ClusterPinImages {
    * @param {string} pin.backgroundColor Background color for the pin
    * @param {string} pin.strokeColor Stroke (border) color for the pin
    * @param {string} pin.labelColor Label (text) color for the pin
-   * @param {string} pin.width The width of the pin
-   * @param {string} pin.height The height of the pin
+   * @param {number} pin.width The width of the pin
+   * @param {number} pin.height The height of the pin
    * @param {string} pin.labelText The label text for the cluster pin (normally size of cluster)
    * @return string The SVG of the pin
```
cea2aj, line 26: I believe return type needs to be updated

static/js/theme-map/ThemeMap.js
```diff
@@ -294,23 +294,23 @@ class ThemeMap extends ANSWERS.Component {
...
       .withIconTemplate('hovered', (pinDetails) => {
-        return getEncodedSvg(this.config.pinClusterImages.getHoveredPin(pinDetails.pinCount));
+        return getEncodedSvg(this.config.pinClusterImages.getHoveredPin(pinDetails.pinCount).svg);
       })
       .withIconTemplate('selected', (pinDetails) => {
-        return getEncodedSvg(this.config.pinClusterImages.getSelectedPin(pinDetails.pinCount));
+        return getEncodedSvg(this.config.pinClusterImages.getSelectedPin(pinDetails.pinCount).svg);
       })
       .withPropertiesForStatus(status => {
+        const defaultPinDimensions = this.config.pinClusterImages.getDefaultPin();
```
cea2aj, line 306: nit: `defaultPinDimensions` is a little misleading since it also contains the SVG. Thoughts on just using `defaultPin`?

oshi97: I wanted to make it clear you should only use this variable for the dimensions, since the SVG depends on pinCount (which I wasn't passing in). Anyways it's passed in now

cea2aj (approved):

PR #808, opened 2021-06-02
Filter and footer collision (css based changes)

yen-tt:
Update scss to have filter with relative position

- changed flexbox properties and width of components when filter is expanded or when window is resized

J=SLAP-1221
TEST=manual

Launched test site (Chrome, Firefox, IE11) and verified layout changes on event, faqs, and people page


static/scss/answers/collapsible-filters/collapsible-filters-templates.scss
```diff
@@ -45,6 +45,20 @@
 .AnswersVerticalGrid.CollapsibleFilters {
   @include CollapsibleFiltersOverrides;
 
+  .Answers-resultsHeader, .Answers-results {
+    @media (min-width: #{$breakpoint-expanded-filters}) {
+      width: math.div($container-width-at-collapsed-filters-breakpoint, $breakpoint-collapsed-filters) * 100%;
```
cea2aj, line 50: What are your thoughts on making a sass variable for this calculation such as $answers-container-width-percentage and putting it in sdk-filter-overrides.scss? That way we won't have the formula duplicated twice

templates/vertical-grid/page.html.hbs
```diff
@@ -33,18 +33,18 @@
...
-        {{!-- {{> templates/vertical-grid/markup/filterbox }} --}}
-        {{!-- {{> templates/vertical-grid/markup/facets }} --}}
-      {{!-- </div> --}}
       <div class="Answers-results js-answersResults">
         {{> templates/vertical-grid/markup/spellcheck }}
         {{> templates/vertical-grid/markup/verticalresults }}
         {{> templates/vertical-grid/markup/pagination }}
         {{!-- {{> templates/vertical-grid/markup/qasubmission }} --}}
       </div>
+      <!-- Uncomment the following div if you want to include filters, facets, or sort options  -->
```
oshi97, line 42: Is there some way we could avoid switching the order of these divs and also the flex-direction: row-reverse? I was thinking it might be possible to add a container div for Answers-results and Answers-filtersWrapper, that had display: flex and justify-content: center. The row-reverse + flex-wrap is a little counter intuitive to me (though definitely cool!) and we might be able to get away with something a little simpler

cea2aj: Would the we be able to avoid swapping the order of the divs by setting the flex order?

yen-tt: I just had the same thought! Testing the change right now

static/scss/answers/collapsible-filters/collapsible-filters-templates.scss
yen-tt: make sense, I will make that change

cea2aj:
When the page is wider than 2015px, the resultsHeader will move to the right of the results since there is enough room to do so. Do you have any ideas on how to fix that?

static/scss/answers/templates/vertical-standard.scss
```diff
@@ -23,6 +24,7 @@
 
   &-filtersWrapper {
     margin-bottom: 20px;
+    order : 1;
```
oshi97, line 27: I actually tried to use the order property in the past, but we want to avoid using it if possible because it makes the html order not match the page order, which makes it harder to add more html in the future, and also makes it more difficult for screen readers/tab navigation users

oshi97: Right now it feels like we're trying to squeeze a 2d layout (results header on the top, filters and results on the bottom) with a 1D DOM structure. We achieved this in the past by using absolute positioning. I think we should consider adding   an extra div for Answers-filtersWrapper and Answers-results, so we can have flex-direction column for Answers-container, and flex-direction row for this new div

templates/vertical-grid/page.html.hbs
```diff
@@ -51,4 +51,4 @@
       {{> templates/vertical-grid/markup/locationbias }}
     </div>
   </div>
-{{/layouts/html }}
+{{/layouts/html }}
```
oshi97, line 54: nit but can we add the new lines back

static/scss/answers/templates/vertical-standard.scss
yen-tt: That is true. Any changes to the template structure (i.e. new div with styling changes) would break experience layout if user don't recreate their pages with the update. Using order here let us bypass dom changes with just css, but harder for user to make add things.

cea2aj: We are trying to avoid any DOM changes for this item since we want it to work with upgrading. I think DOM changes are the right solution, however we are a little constrained for this item. We spoke to Tom about this yesterday and we may need to hold off on DOM changes until we make some larger UI changes so that we can justify a more complicated upgrade process. Since most sites aren't using custom footers, it would make their upgrade process more difficult without providing any noticeable improvements

static/scss/answers/collapsible-filters/collapsible-filters-templates.scss
```diff
@@ -45,6 +45,26 @@
...
+    @media (min-width: #{$breakpoint-expanded-filters}) {
+      width: $answers-container-width-percentage;
+    }
+    width: var(--hh-answers-container-width);
+    max-width: var(--hh-answers-container-width);
+  }
+
+  .Answers-resultsHeader {
+    @media (min-width: #{$breakpoint-expanded-filters}) {
+      margin: 0 math.div(100% - $answers-container-width-percentage,2);
```
cea2aj, line 58: This does improve the issue, however the header will still snap right above 3233px which could happen while fullscreen on a 4k monitor. The issue is that this calculated percentage doesn't take into account the results having a max width

While we could possibly resolve this issue, I feel that we should probably hold of on this PR because of the complexity that this change would introduce

PR #812, opened 2021-06-04
Product snapshots

cea2aj:
Snapshot the the product video and image cards

J=SLAP-1361
TEST=visual

Check updated percy snapshots

### DIFF
--- a/test-site/config/index.json
+++ b/test-site/config/index.json
@@ -40,6 +40,10 @@
       "mapConfig": {
         "mapProvider": "Google"
       }
+    },
+    "products": {
+      "universalSectionTemplate": "grid-three-columns",
+      "cardType": "product-prominentimage"
     }
   }
 }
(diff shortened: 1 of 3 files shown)


test-site/config/index.json
```diff
@@ -40,6 +40,10 @@
       "mapConfig": {
         "mapProvider": "Google"
       }
+    },
+    "products": {
+      "universalSectionTemplate": "grid-three-columns",
```
oshi97, line 45: do we have to provide this override in the index.json?

oshi97 (approved):

cea2aj:
Screen Shot 2021-06-04 at 3 46 18 PM
Since we have javascript disabled in percy, the extra vertical will overflow the sidebar. We'll just have to keep that in mind for the vertical full page map snapshots

PR #824, opened 2021-06-08
Added common partials for templates

cjiang2000:
Created a common partial for each template.hbs that has duplicates in other templates. Linked all of those individual handlebars files to the common partial.
Common Partials:

- appliedfilters.hbs
- facets.hbs
- filterbox.hbs
- locationbias.hbs
- navigation.hbs
- pagination.hbs
- qasubmission.hbs
- searchbar.hbs
- sortoptions.hbs
- spellcheck.hbs
- verticalresults.hbs
- verticalresultscount.hbs

J = SLAP-1372
TEST = manual

Built website and made sure that each page was working.


templates/common-partials/script/pagination.hbs
```diff
@@ -0,0 +1,19 @@
...
+  container: "#js-answersPagination",
+  onPaginate: (newPageNumber, oldPageNumber, totalPages) => {
+    if (window.parentIFrame) {
+      const paginateMessage = { action: 'paginate' };
+      window.parentIFrame.sendMessage(JSON.stringify(paginateMessage));
+    } else {
+      document.documentElement.scrollTop = 0;
+      document.body.scrollTop = 0; // Safari
+    }
+    const resultsColumn = document.querySelector(".js-locator-resultsWrapper");
```
oshi97, line 11: I did a little more thinking and I think it might be better to have the vertical-full-page-map not use the common partial, at least for now, and we can remove lines 11-14 from the common partial. It's a little spaghetti to have the shared code have logic that is specific to a single page template. In the future we probably want some way to insert page template specific code into the common partial, but I think that's out of scope for this PR.

oshi97 (approved):
lgtm!

PR #825, opened 2021-06-08
Use the esbuild-loader for production minification

cea2aj:
Use the esbuild-loader in order to speed up our production build javascript and css minification

This changes speeds up the production webpack build by approximately 30 to 40%. The trade-off with esbuild is that the resulting minimized file is not quite as small. For index.html of the test site, the page size is about 2% larger.

We can remove cssnano and let esbuild handle css minification because the esbuild css minimizer leads to 8% faster webpack builds than CSS nano, and the resulting pages are 0.5% larger.

esbuild-loader does not support sourcemaps for css minification, however we currently don't provide css sourcemaps so this is not a regression

J=SLAP-1374
TEST=manual

Use the time program to measure the performance difference. Smoke test the site on chrome and on IE11.


yen-tt:
Can esbuild-loader JS transpilation support replace babel-loader as well?

cea2aj:
> Can esbuild-loader JS transpilation support replace babel-loader as well?

No, I tried it but esbuild does not have full support for transpiling to ES5 so we still need to use babel for that

static/package-lock.json
```diff
@@ -1060,32 +1060,6 @@
       "resolved": "https://registry.npmjs.org/@types/anymatch/-/anymatch-1.3.1.tgz",
       "integrity": "sha512-/+CRPXpBDpo2RK9C68N3b2cOvO0Cf5B9aPijHsoDQTHivnGSObdOF2BRQOYjojWTDy6nQvMjmqRXIxH55VjxxA=="
     },
-    "@types/eslint": {
```
oshi97, line 1063: just wanted to check we're ok removing these even though all we did package.json wise was install a new package?

static/webpack/webpack.prod.js
```diff
@@ -29,6 +31,14 @@ module.exports = () => {
           }
         }
       ]
-    }
+    },
+    optimization: {
+      minimizer: [
+        new ESBuildMinifyPlugin({
+          target: 'es5',
```
oshi97, line 38: would we be able to shrink the bundle just a little bit if we specify specific browser versions? Or maybe, since we're already running babel and importing polyfills, can we just not specify a target?

static/webpack/webpack.prod.js
```diff
@@ -29,6 +31,14 @@ module.exports = () => {
...
+          css: true
```
oshi97, line 39: does using this on top of cssnano shrink the css bundle even more? If not maybe we can get away with just using one of these?

static/package-lock.json
cea2aj: I ran another fresh install and now less packages are being removed. I think it might have been my node version that caused it to change the package lock so much

static/webpack/webpack.prod.js
cea2aj: I can check if it's possible to target ie11. I tried not specifying a target and it didn't work since esbuild will introduce some newer javascript during minification if the target is not es5

static/webpack/webpack.prod.js
cea2aj: good point! I will try this with and without cssnano

cea2aj: In my testing the esbuild css minimizer is 8% faster than CSS nano, and the resulting pages are 0.5% larger. Based on that, I think we can remove cssnano

static/webpack/webpack.prod.js
oshi97: ah dang got it, I guess it makes sense it would undo the babel stuff in order to make the code smaller

cea2aj: It is not possible to target ie11. ES5 is the only target that will work for us

oshi97 (approved):

PR #826, opened 2021-06-09
add multilang testing to theme

yen-tt:
Set up test site to support multilanguage and generate Percy snapshots

- Updated locale config to support spanish (es)
- Added index.es.json and config-override *.es.json for pages to have spanish label and multilang cards, if exists
- Updated standardpagenavigator and iframenavigator to support navigating to non-default locale urls
- Added multilangphotographer to capture snapshots in alternate languages

J=1363
TEST=manual and percy

Launched test site and search/navigate through Spanish and English pages to see that messages are translated accordingly.
New snapshots are generated on percy for alternate language (es).


test-site/scripts/create-verticals.js
```diff
@@ -18,27 +18,27 @@ const verticalConfiguration = {
   locations: {
     verticalKey: 'KM',
     template: 'vertical-map',
-    cardName: 'location-standard'
```
tmeyer2115, line 21: Will we still have Percy coverage for all the single-language variants if we replace them with multi-lang here? We want to ensure total coverage for both the single and multi-language variants. Even though the multi-lang and single-lang are supposed to appear the same, they are technically different files. This means someone could bork one and not the other.

yen-tt: you are right, this will only cover multilang variants. I can revert this to the default configuration for en and have some config override for es files to take multilang cards instead.

tests/percy/multilangphotographer.js
```diff
@@ -0,0 +1,170 @@
+/**
+ * @typedef {import('./pagenavigator.js')} PageNavigator
+ * @typedef {import('./Camera.js')} Camera
+ */
+
+/**
+ * Responsible for determining which snapshots to take for specific locale
+ */
+class MultilangPhotographer {
```
tmeyer2115, line 9: Would it make sense to consolidate down to a single `Photographer` class? I guess it'd effectively be what we have here, since it's more general. What do you think @cea2aj ?

yen-tt: Right now `MultilangPhotographer` covers majority of the snapshot tests from `Photographer`(beside directanswers document search) - I'm not sure if we actually want all those snapshots, and if we want to separate it this way for when we may want to include for specific test scenarios only for single/multilang variants in the future.

cea2aj: Currently the responsibility of the `Photographer` is to define the test suite itself. I suggested that we make a `MultilangPhotographer` as a way to break up the test suite so that we don't have to have all of the snapshots in a single file. I have another idea for allowing us to split up the tests into multiple files. I think this change would be out of scope of this PR, but i'll include in here for reference:

Rather than having the photographer define the test suite, the test suite could instead be defined by files in particular folder (similar to how jest and test cafe work). The files would contain exported functions which define the snapshots to take, and they would each be passed a page navigator and a camera. For multilang we could pass an additional object which contains language-specific strings for use in the tests. At this point, the photographer might not be necessary anymore

tests/percy/iframepagenavigator.js
```diff
@@ -9,24 +9,50 @@ class IframePageNavigator extends PageNavigator {
...
+    this._localeVerticalUrlPath = '&verticalUrl=';
+  }
+  
+  /**
+   * Sets locale param for the site's url based on given locale
+   * 
+   * @param {string} locale 
+   */
+  setCurrentLocale(locale) {
+    this._localeUniversalUrlPath = locale === this._defaultLocale? '' : '&verticalUrl=' + locale;
```
cea2aj, line 42: nit: we usually put a space before the question mark when using the ternary operator

tests/percy/iframepagenavigator.js
```diff
@@ -9,24 +9,50 @@ class IframePageNavigator extends PageNavigator {
...
-    const url = `${this._siteUrl}/${this._iframePage}.html?${queryParamsString}`;
+    const url = `${this._siteUrl}/${this._iframePage}.html?${queryParamsString}${this._localeUniversalUrlPath}`;
     await this._page.goto(url);
     await waitTillHTMLRendered(this._page);
   }
 
   async gotoVerticalPage(vertical, queryParams = {}) {
     const queryParamsString = getQueryParamsString(queryParams);
-    const url = `${this._siteUrl}/${this._iframePage}.html?verticalUrl=${vertical}.html&${queryParamsString}`;
+    const url = `${this._siteUrl}/${this._iframePage}.html?${this._localeVerticalUrlPath}${vertical}.html&${queryParamsString}`;
```
cea2aj, line 55: Rather than having a `_localeVerticalUrlPath`, would it be simpler to instead have a `_localePrefix` which would be '' for the default locale or something like 'es/' for Spanish?

cea2aj: Alternatively we could check if a locale is present, and if so we have one way of defining the url, and if the url is not present, we have another way of defining the URL

tests/percy/index.js
```diff
@@ -17,11 +18,18 @@ PercyScript.run(async (page, percySnapshot) => {
...
   const standardPageNavigator = new StandardPageNavigator(page, `http://localhost:${PORT}`);
   const iframePageNavigator = new IframePageNavigator(page, `http://localhost:${PORT}`, 'iframe_test');
+
   const standardCamera = new Camera(percySnapshot);
   const iframeCamera = new Camera(percySnapshot, true);
 
   await (new Photographer(standardPageNavigator, standardCamera).captureSnapshots());
   await (new Photographer(iframePageNavigator, iframeCamera).captureSnapshots());
 
+  standardPageNavigator.setCurrentLocale('es');
```
cea2aj, line 28: what do you think about just using `setLocale`?

tests/percy/multilangphotographer.js
```diff
@@ -0,0 +1,170 @@
...
+    await this._captureVerticalSearch();
+    await this._captureVerticalGridSearch();
+    await this._captureVerticalMapSearch();
+    await this._captureVerticalFullPageMapSearch();
+    await this._captureDirectAnswers();
+  }
+
+  async _captureUniversalSearch () {
+    await this._pageNavigator.gotoUniversalPage();
+    await this._camera.snapshot(this.localeSnapshotTag + 'universal-search');
```
cea2aj, line 82: Rather than a localeSnapshotTag, what do you think of having a locale property in camera, and if it is set, the camera will prepend the locale to the snapshot?

tests/percy/multilangphotographer.js
```diff
@@ -0,0 +1,170 @@
...
+  setLocale(locale) {
+    this._locale = locale;
+    this.localeSnapshotTag = this._locale === 'en'? '' : this._locale + '--';
+    this._setLocaleQueries();
+  }
+
+  /**
+   * Sets custom queries based on locale
+   */
+  _setLocaleQueries() {
```
cea2aj, line 51: It seems like now the multilang phtographer has two repsonsibilites: defining the test suite for multilang tests and defining the multilang translations to be used in the test suit. What do you think about instead passing in the translated queries to the MultilangPhotographer as an object rather than defining them here? We could have a class or a json file that contains the translations for each locale, and then we load the translations for the specific locales and pass it to the multilang photographer

tests/percy/index.js
yen-tt: Since navigator has a defaultLocale field to help out with the url logic, and I thought using `setLocale` may be less clear whether it's setting the default or current locale for the site.

cea2aj: oh, I think `setCurrentLocale` makes sense then

tests/percy/multilangphotographer.js
yen-tt: I like the idea of using a json file to hold the translated queries. If we do pass the translation of specific locale to `multilangphotographer` (and have camera handles the addition locale name tag), then the photographer no longer need to have knowledge of locale. Although, I'm thinking we could still pass locale in multilangphotographer (along with the queries) and have it update locale property for its navigator and camera as well rather than updating each component separately in index.js. What do you think?

cea2aj: Either way sounds good to me

tests/percy/index.js
```diff
@@ -17,11 +19,22 @@ PercyScript.run(async (page, percySnapshot) => {
...
+  iframePageNavigator.setCurrentLocale('es');
+
+  standardCamera.setLocale('es');
+  iframeCamera.setLocale('es');
+  
+  await (new MultilangPhotographer(standardPageNavigator, standardCamera, queryConfig.es).captureSnapshots());
+  await (new MultilangPhotographer(iframePageNavigator, iframeCamera, queryConfig.es).captureSnapshots());
```
oshi97, line 36: do people think it would be possible to not include the "multilang + iframe" combo? I'm imagining nearly all bugs in the iframe experience would be caught by the "en + iframe" snapshots, outside of maybe the multilang variant having longer text that causes weird wrapping/overflow bugs?

yen-tt: since iframe experience utilize the same pages - no changes to translation logic and multilang template cards, it does seems redundant to repeat all multilang variant tests for iframe. Yeah, I can remove this test combo

oshi97 (approved):
this lgtm!

PR #828, opened 2021-06-10
separate bundle into legacy and modern bundle

oshi97:
This commit separates bundle.js into the polyfill-free
bundle.js and polyfilled bundle-legacy.js. Both still
run through the babel loader, because the esbuild minifier
is set to target es5, and does not let you minify some
bundles to es5 and some bundles to newer javascript.

The type="module" and nomodule script loading pattern is used
to dynamically load the correct bundle. This is the pattern we
use for determining whether to use answers-modern or answers
assets. This shrinks bundle.js from 443KB to 313KB in prod mode
(and from 1.3MB to 705KB in dev mode). In prod mode the page speed
went from 50 -> 56, and in dev mode the page speed lighthouse scores
went from 36 -> 48.

J=SLAP-1373
TEST=manual

see ~0.7s build speed improvements on yanswers and answers-atlanticsearch
when I remove the core-js import from entry.js
this build speedup doesn't show up for the test-site for some reason, though
investigated for a while but could not figure out why
I figured that yanswers and another random repo would be better benchmarks, though
the test-site's build speed was also not hurt at all by this change

see that the production modern bundle shrinks from 443KB to 313KB after removing
polyfills

smoke test that full page map on ie11 (browserstack), firefox, safari, chrome, and
galaxy s6 chrome (browserstack)

see that bundle.js is inlined properly, and only the modern one is inlined

### DIFF
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -85,7 +85,8 @@
       });
       {{/babel}}
     </script>
-    <script src="{{relativePath}}/bundle.js" data-webpack-inline></script>
+    <script src="{{relativePath}}/bundle.js" type="module" data-webpack-inline></script>
+    <script src="{{relativePath}}/bundle-legacy.js" nomodule></script>
     {{#if global_config.googleAnalyticsId}}
       <!-- Global site tag (gtag.js) - Google Analytics -->
       <script async src="https://www.googletagmanager.com/gtag/js?id={{global_config.googleAnalyticsId}}"></script>
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -85,8 +85,7 @@
       });
       {{/babel}}
     </script>
-    <script src="{{relativePath}}/bundle.js" type="module" data-webpack-inline></script>
-    <script src="{{relativePath}}/bundle-legacy.js" nomodule></script>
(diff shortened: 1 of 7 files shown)


layouts/html.hbs
```diff
@@ -85,7 +85,8 @@
       });
       {{/babel}}
     </script>
-    <script src="{{relativePath}}/bundle.js" data-webpack-inline></script>
+    <script src="{{relativePath}}/bundle.js" type="module" data-webpack-inline></script>
+    <script src="{{relativePath}}/bundle-legacy.js" nomodule></script>
```
tmeyer2115, line 89: I think we might need to chat with Product about this change. It has some significant upgrade implications. If somebody has forked `html.hbs`, their site will stop working in IE11 if they're not careful during the upgrade. The `bundle.js` asset no longer points to the same thing post-upgrade.

oshi97: talked to Rose and she gave the ok! people who overrode html.hbs will have to refresh their html.hbs anyways for the dev sdk branch change

tmeyer2115 (approved):
You might mention to Rose that people who forked `static/entry.js` will not see the performance benefit unless they update their fork.

PR #830, opened 2021-06-11
Additional locator acceptance tests

cea2aj:
Add additional locator acceptance tests

These tests cover:
- Pagination
- Default initial search
- Interacting with the results list

This PR also includes some additional updates
- Set `yxt-Pin` and `yxt-PinCluster` CSS classes so that I can safely target them from the tests. This should only affect Mapbox since google does not use HTML pins
- Updated the functions on the blocks to all be async. I think this consistency makes it easier to write acceptance tests. It also separates concerns by keeping the lower-level `Selector` object within the blocks.

J=SLAP-1362
TEST=acceptance

### DIFF
--- a/.github/run_browserstack_acceptance.sh
+++ b/.github/run_browserstack_acceptance.sh
@@ -5,6 +5,5 @@ GITHUB_BRANCH=${GITHUB_REF#refs/heads/}
 export BROWSERSTACK_BUILD_ID="${GITHUB_BRANCH} - ${GITHUB_RUN_ID}"
 COMMIT_MSG_TITLE=$(git log -n 1 --pretty=format:%s)
 export BROWSERSTACK_TEST_RUN_NAME=$COMMIT_MSG_TITLE
-export IS_DEVELOPMENT_PREVIEW="false"
 
 npm run acceptance -- --browsers browserstack:ie@11.0 browserstack:safari browserstack:firefox
--- a/.github/run_browserstack_acceptance.sh
+++ b/.github/run_browserstack_acceptance.sh
@@ -5,5 +5,6 @@ GITHUB_BRANCH=${GITHUB_REF#refs/heads/}
 export BROWSERSTACK_BUILD_ID="${GITHUB_BRANCH} - ${GITHUB_RUN_ID}"
 COMMIT_MSG_TITLE=$(git log -n 1 --pretty=format:%s)
 export BROWSERSTACK_TEST_RUN_NAME=$COMMIT_MSG_TITLE
+export IS_DEVELOPMENT_PREVIEW="false"
 
 npm run acceptance -- --browsers browserstack:ie@11.0 browserstack:safari browserstack:firefox
--- a/tests/acceptance/blocks/page.js
+++ b/tests/acceptance/blocks/page.js
@@ -1,0 +1,16 @@
+import { Selector, t } from 'testcafe';
+
+/**
+ * Models the user interaction with the page as a whole
+ */
+class Page {
+  constructor () {
+    this._document = Selector('.YxtPage-doc');
+  }
+
+  async scrollToBottom () {
+    await t.scroll(this._document, 'bottom');
+  }
+}
+
+export default new Page();
--- a/tests/acceptance/blocks/pagination.js
+++ b/tests/acceptance/blocks/pagination.js
@@ -9,6 +9,9 @@ class Pagination {
     this._nextResultsButton = Selector('.js-yxt-Pagination-next');
   }
 
+  /**
+   * Clicks the next results page button
+   */
   async nextResults () {
     await Page.scrollToBottom(); // We must scroll to the bottom of the page to be able to click the next results page button
     await t.click(this._nextResultsButton);
--- a/tests/acceptance/blocks/pagination.js
+++ b/tests/acceptance/blocks/pagination.js
@@ -1,0 +1,18 @@
+import { Selector, t } from 'testcafe';
+import Page from './page';
+
+/**
+ * Models the user interaction with a {import('@yext/answers-search-ui').PaginationComponent}.
+ */
+class Pagination {
+  constructor () {
+    this._nextResultsButton = Selector('.js-yxt-Pagination-next');
+  }
+
+  async nextResults () {
+    await Page.scrollToBottom(); // We must scroll to the bottom of the page to be able to click the next results page button
+    await t.click(this._nextResultsButton);
+  }
+}
+
+export default new Pagination();
--- a/tests/acceptance/blocks/pagination.js
+++ b/tests/acceptance/blocks/pagination.js
@@ -1,5 +1,6 @@
 import { Selector, t } from 'testcafe';
 import Page from './page';
+import VerticalResults from './verticalresults';
 
 /**
  * Models the user interaction with a {import('@yext/answers-search-ui').PaginationComponent}.
@@ -14,6 +15,7 @@ class Pagination {
    */
   async nextResults () {
     await Page.scrollToBottom(); // We must scroll to the bottom of the page to be able to click the next results page button
+    await VerticalResults.scrollToBottom();
     await t.click(this._nextResultsButton);
   }
 }
--- a/tests/acceptance/blocks/thememap.js
+++ b/tests/acceptance/blocks/thememap.js
@@ -60,21 +60,27 @@ class ThemeMap {
   async waitForZoomStabilization () {
     const pollingIntervalMsecs = 500;
     const minNumStableIntervals = 3;
+    const maxNumIntervals= 10;
     let previousZoom = await this.getZoom();
     let numStableIntervals = 0;
+    let numIntervals = 0;
     let isZoomStabilized = false;
+    let isMaxIntervalsReached = false;
 
-    while (!isZoomStabilized) {
+    while (!isZoomStabilized && !isMaxIntervalsReached) {
       await t.wait(pollingIntervalMsecs);
       const currentZoom = await this.getZoom();
 
-      if (currentZoomSize === previousZoom) {
+      if (currentZoom === previousZoom) {
         numStableIntervals++;
       } else {
         numStableIntervals = 0;
       }
 
+      numIntervals++;
+
       isZoomStabilized = (numStableIntervals >= minNumStableIntervals);
+      isMaxIntervalsReached = (numIntervals >= maxNumIntervals);
       previousZoom = currentZoom;
     }
   }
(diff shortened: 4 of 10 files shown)


tmeyer2115:
> Add additional locator acceptance tests
> 
> J=SLAP-1362
> TEST=acceptance

Can you provide high level descriptions of the tests added?

tests/acceptance/blocks/page.js
```diff
@@ -0,0 +1,16 @@
+import { Selector, t } from 'testcafe';
+
+/**
+ * Models the user interaction with the page as a whole
+ */
+class Page {
```
tmeyer2115, line 6: If we have a base `Page` class like this, what do you think about having all the other, specific `Pages` extend it?

.github/run_browserstack_acceptance.sh
```diff
@@ -5,5 +5,6 @@ GITHUB_BRANCH=${GITHUB_REF#refs/heads/}
 export BROWSERSTACK_BUILD_ID="${GITHUB_BRANCH} - ${GITHUB_RUN_ID}"
 COMMIT_MSG_TITLE=$(git log -n 1 --pretty=format:%s)
 export BROWSERSTACK_TEST_RUN_NAME=$COMMIT_MSG_TITLE
+export IS_DEVELOPMENT_PREVIEW="false"
```
tmeyer2115, line 8: Did we ever figure out why the tests were running in dev mode? By default, it should be Prod

tests/acceptance/blocks/thememap.js
```diff
@@ -29,6 +38,46 @@ class ThemeMap {
...
+   * Waits for the map zoom to sabilize
+   */
+  async waitForZoomStabilization () {
+    const pollingIntervalMsecs = 500;
+    const minNumStableIntervals = 3;
+    let previousZoom = await this.getZoom();
+    let numStableIntervals = 0;
+    let isZoomStabilized = false;
+
+    while (!isZoomStabilized) {
```
tmeyer2115, line 67: Do we need a timeout here? I'm wondering if there's a situation in which we could get stuck waiting indefinitely.

tests/acceptance/blocks/verticalresults.js
```diff
@@ -32,19 +35,47 @@ class VerticalResults {
...
-  getResults () {
-    return this._selector;
+  async isCardFocused () {
+    return this._focusedCard.exists;
+  }
+
+  /**
+   * Clicks the first result card
+   */
+  async clickFirstCard () {
```
tmeyer2115, line 57: Could we make a general `clickCard(cardIndex)` method? This would then be a special case of that. I could see us wanting the more general method for future tests.

tests/acceptance/blocks/page.js
cea2aj: Oliver and I opted to not create specific pages like we do in the SDK for the theme acceptance tests. When writing the new theme acceptance tests, Oliver asked me to make them simpler, and my solution was to remove the Page abstraction from the tests. By removing that abstraction the tests are simpler because they no longer require getters and setter to access the blocks from each page. I looked into testcafe's guides for writing tests with a pages model, and I think that makes a lot of sense for websites with completely different functionality on each page, but since the search ui is component based, organizing the code into blocks which correspond with those components is simpler. The acceptance tests in the SDK have an additional layer of abstraction that not even the testcafe page model examples have, and I think that these tests are actually more similar to testcafe's examples in terms of the levels of abstraction used

tests/acceptance/blocks/thememap.js
cea2aj: I don't think we need one. The zoom value is actually very stable. But I can add a `maxNumPolls` just to be safe

tests/acceptance/blocks/verticalresults.js
cea2aj: sure

.github/run_browserstack_acceptance.sh
cea2aj: I mostly added this here for clarity. Why did we believe they were running in dev mode again? When I look at the acceptance tests for the PR where I removed postcss, they successfully ran in IE11 in browserstack which isn't possible if it were running in dev mode

tests/acceptance/blocks/thememap.js
```diff
@@ -1,23 +1,32 @@
...
     this._searchThisAreaToggle = Selector('.Answers-searchThisAreaToggleLabel');
     this._searchThisAreaButton = Selector('.js-searchThisAreaButton');
   }
 
   /**
-   * Selects a pin from the Mapbox map.
+   * Selects a pin from the map.
    */
-  async selectMapboxPin () {
+  async selectPin () {
```
tmeyer2115, line 18: For this and `selectPinCluster` is there any way to target a specific pin or cluster? Just curious. Doesn't seem like we need that for this PR, but we may want that for future tests.

tests/acceptance/blocks/verticalresults.js
```diff
@@ -32,19 +35,47 @@ class VerticalResults {
...
 
   /**
-   * Gets a selector for the card associated with the currently focused map pin.
-   * @returns {Selector}
+   * Returns the number of results from the VerticalResultsCount-start.
+   * @returns {Promise<number>}
    */
-  getFocusedCard () {
-    return this._focusedCard;
+   async getResultsCountStart () {
```
tmeyer2115, line 41: Small nit: what do you think of these naming changes
- `getResultsCountTotal` -> `getResultsCount`
- `getResultsCountStart` -> `getResultsOffset`

The phrase `ResultsCountTotal` sounds a bit redundant to me and `ResultsCountStart` is ambiguous.

tests/acceptance/blocks/page.js
tmeyer2115: Got it, thanks for the explanation! I was assuming a similar abstraction layer to the SDK. Let's see how this level of abstraction works for us in the Theme. So far it does seem to make things simpler! If all goes well, maybe we change the SDK to use this scheme as well.

tests/acceptance/blocks/thememap.js
cea2aj: Yes, it should be possible to select specific pins or clusters since the different clusters and pins have different aria labels that we can use for targeting.

tests/acceptance/blocks/verticalresults.js
cea2aj: I will update. I think those are clearer

tests/acceptance/index.js
```diff
@@ -31,8 +31,10 @@ async function runTests (browsers) {
     if (numberTestsFailed > 0) {
       process.exit(1);
     }
-  }
-  finally {
+  } catch (e) {
+    console.error(e);
+    process.exit(1);
+  } finally {
     await testcafe.close();
```
oshi97, line 38: does anything weird happen if we exit the process without closing testcafe?

tests/acceptance/blocks/thememap.js
```diff
@@ -29,6 +38,58 @@ class ThemeMap {
...
+  }
+
+  /**
+   * Gets the current zoom of the map
+   * @returns {Promise<number>}
+   */
+  async getZoom () {
+    const zoom = await ClientFunction(() => {
+      return ANSWERS
+        .components
```
oshi97, line 55: does `ANSWERS.components.getActiveComponent()` work here?

tests/acceptance/index.js
cea2aj: not that I know of but, thinking about it further, I think I should close it before forcefully exiting

tests/acceptance/blocks/thememap.js
cea2aj: would that be less robust? If we get the component by name we know for sure that it is the VerticalFullPageMapOrchestrator. If we use getActiveComponent(), would that change baseed on which components were used before calling `getZoom()`?

oshi97: getActiveComponent() gets the first component by type, and here we get the first component by name, so I think it should be roughly the same robustness (unless I'm misunderstanding something)?

cea2aj: oh, that makes sense. I hadn't used `getActiveComponent()` before and I didn't realize it accepted a parameter

tests/acceptance/suites/vertical-full-page-map.js
```diff
@@ -7,44 +7,44 @@ import Pagination from '../blocks/pagination';
 fixture`Vertical Full Page Map`
   .page(`http://localhost:${PORT}/locations_full_page_map`)
 
-test('Can search and get results', async t => {
-  await SearchBar.submitQuery('virginia');
-  const isResultsPresent = await VerticalResults.isResultsPresent();
-  await t.expect(isResultsPresent).ok();
-});
+// test('Can search and get results', async t => {
```
tmeyer2115, line 10: Are these tests all commented out as a result of some local testing changes?

cea2aj: I commented out most of the tests so that I could more quickly test the one that is failing on IE11. I will comment them back in when i am done testing

tests/acceptance/blocks/pagination.js
```diff
@@ -14,6 +15,7 @@ class Pagination {
    */
   async nextResults () {
     await Page.scrollToBottom(); // We must scroll to the bottom of the page to be able to click the next results page button
```
tmeyer2115, line 17: For my own edification, we first need to scroll to the bottom of the page, then scroll to the bottom of the VR to click the button?

cea2aj: Correct. It's needed primarily for screen with smaller heights. I'm actually making an item for this because when the height of the browser is small, the vertical results scroll bar can get hidden, and scrolling the page as well is the only way to get to the bottom of the vertical results

.github/run_browserstack_acceptance.sh
oshi97, line 8: I don't think setting the environment variable here does anything. At this point `npm run build-test-site` is already done right?

cea2aj: oh, you're right. I would need to set it in the github actions for it to take effect in the build. Honestly though, prod mode is the default, so I don't really think it's necessary to set it

tests/acceptance/blocks/thememap.js
```diff
@@ -29,6 +38,57 @@ class ThemeMap {
   async clickSearchThisAreaButton () {
     await t.click(this._searchThisAreaButton);
   }
+
+  async clickMap () {
+    await t.wait(10000);
```
oshi97, line 43: were these 10s waits for testing?

cea2aj: ah, my bad. I forget to take these out after testing

oshi97 (approved):
lgtm!

PR #831, opened 2021-06-11
Filter and Cluster acceptance tests for the locator

cea2aj:
Add additional filter and cluster tests

Enable clustering on the filter page and add some test coverage for filters and clusters.

J=SLAP-1362
TEST=acceptance


tests/acceptance/blocks/filters.js
```diff
@@ -0,0 +1,36 @@
+import { Selector, t } from 'testcafe';
+
+/**
+ * Models the user interaction with filters
+ */
+class Filters {
```
tmeyer2115, line 6: Would it make sense to call this `CollapsibleFilters`? The class seems specific to that use-case. We might want to create a `Filters` class later on for the static, non-collapsible filters.

cea2aj: ya, that makes more sense

tmeyer2115 (approved):

PR #832, opened 2021-06-14
add static/assets loader usage to snapshots

oshi97:
This commit adds a usage of the html-asset-loader
to our Percy snapshots. The universal page's KM
vertical now uses a custom static icon.

J=NONE
T=manual

test that I can load up the test site and the icon
appears


cea2aj (approved):
lgtm

PR #834, opened 2021-06-14
Wrap 'Opens At' text in a call to `_localizer.getTranslation`.

tmeyer2115:
In the `open-status/message-factory` we forgot to wrap one of the static strings in a `_localizer.getTranslation` call. This caused the 'Opens At' string to always appear in English.

J=SLAP-1380
TEST=manual, auto

This fix was already verified manually by a HH. I am just adding it to the Theme. I did add a unit test suite for the `message-factory`. The test suite and test I added would have caught this issue and so will ensure there aren't future regressions on this front.

### DIFF
--- a/tests/static/js/hours/open-status/messagefactory.js
+++ b/tests/static/js/hours/open-status/messagefactory.js
@@ -1,0 +1,50 @@
+import { OpenStatusStrings, OpenStatusTypes } from '../../../../../static/js/hours/open-status/constants';
+import OpenStatusMessageFactory from '../../../../../static/js/hours/open-status/messagefactory';
+
+const localizer = {
+  getLocalizedTime: jest.fn(yextTime => {
+    let time = new Date();
+    time.setHours(Math.floor(yextTime / 100));
+    time.setMinutes(yextTime % 100);
+
+    return time.toLocaleString('en', {
+      hour: 'numeric',
+      minute: 'numeric'
+    });
+  }),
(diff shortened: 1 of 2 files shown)


tests/static/js/hours/open-status/messagefactory.js
```diff
@@ -0,0 +1,50 @@
...
+    time.setMinutes(yextTime % 100);
+
+    return time.toLocaleString('en', {
+      hour: 'numeric',
+      minute: 'numeric'
+    });
+  }),
+  getTranslation: jest.fn(str => {
+    const translations = {
+      [OpenStatusStrings.CLOSED]: 'Closed',
```
oshi97, line 17: Can we use strings different than the values of the `OpenStatusStrings` object? Right now I think this technically does not test the regression being fixed

tmeyer2115: So because it verifies the expected calls to `getTranslation` are being made, it does cover this specific regression. We did not wrap one of the strings in a `getTranslation` call. The assertions on line 47 and 48 make sure we have two calls to that method, meaning all strings are wrapped.

oshi97: Oh I see I missed that, I feel like it would be a little simpler to just change the mocked value of getTranslation to be an actualy translation instead of adding extra expects but that also works!

oshi97 (approved):

PR #836, opened 2021-06-16
Refresh error due to double encoding in verticalUrl

yen-tt:
- URLSearchParams set() encoded verticalUrl even though it's already encoded, extracted from window's URL pathname. Changed to appending to the param string

J=SLAP-1376
TEST=manual

Launched multilang site with iframe data-path directed to a page with spaces in its names. Confirmed refresh and back/forward nav works accordingly.

### DIFF
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -115,11 +115,11 @@
     <script>
       {{#babel}}
       function iframeGetSearchParams() {
-        let params = new URLSearchParams(window.location.search);
+        let params = window.location.search.substr(1);
         let verticalUrl = window.location.pathname.substr(1);
         return verticalUrl 
-          ? params.toString() + '&verticalUrl=' + verticalUrl
-          : params.toString();
+          ? params + '&verticalUrl=' + verticalUrl
+          : params;
       }
       let iframeLoadedResolve;
       window.iframeLoaded = new Promise(resolve => {
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -116,11 +116,10 @@
       {{#babel}}
       function iframeGetSearchParams() {
         let params = new URLSearchParams(window.location.search);
-        let verticalUrl = window.location.pathname.substr(1)
-        if (verticalUrl) {
-          params.set('verticalUrl', verticalUrl);
-        }
-        return params.toString();
+        let verticalUrl = window.location.pathname.substr(1);
+        return verticalUrl 
+          ? params.toString() + '&verticalUrl=' + verticalUrl
+          : params.toString();
       }
       let iframeLoadedResolve;
       window.iframeLoaded = new Promise(resolve => {
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -115,8 +115,8 @@
     <script>
       {{#babel}}
       function iframeGetSearchParams() {
-        let params = window.location.search.substr(1);
-        let verticalUrl = window.location.pathname.substr(1);
+        const params = window.location.search.substr(1);
+        const verticalUrl = window.location.pathname.substr(1);
         return verticalUrl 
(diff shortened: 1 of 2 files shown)


layouts/html.hbs
```diff
@@ -116,11 +116,10 @@
       {{#babel}}
       function iframeGetSearchParams() {
         let params = new URLSearchParams(window.location.search);
-        let verticalUrl = window.location.pathname.substr(1)
-        if (verticalUrl) {
-          params.set('verticalUrl', verticalUrl);
-        }
-        return params.toString();
+        let verticalUrl = window.location.pathname.substr(1);
```
tmeyer2115, line 119: If we're not using `params.set`, do we need to wrap the `window.location.search` in `URLSearchParams` any more? Could we instead just have:
```
let params = window.location.search;
return verticalUrl  ? params + '&verticalUrl=' ...
```

yen-tt: ah yup, that's not needed any more. Updated

static/js/iframe-common.js
```diff
@@ -15,9 +15,6 @@ export function generateIFrame(domain, queryParam, urlParam) {
     var paramString = window.location.search;
     paramString = paramString.substr(1, paramString.length);
 
-    // Decode ASCII forward slash to avod repeat encodings on page refreshes
-    paramString = paramString.replace("%2F", "/");
```
tmeyer2115, line 19: Are we able to get rid of this because `window.location.search` is already encoded?

layouts/html.hbs
```diff
@@ -115,12 +115,11 @@
     <script>
       {{#babel}}
       function iframeGetSearchParams() {
-        let params = new URLSearchParams(window.location.search);
-        let verticalUrl = window.location.pathname.substr(1)
-        if (verticalUrl) {
-          params.set('verticalUrl', verticalUrl);
-        }
-        return params.toString();
+        let params = window.location.search.substr(1);
```
oshi97, line 118: nit but we can use const here!

static/js/iframe-common.js
yen-tt: yeah, it's already encoded pulling from browser. I believe this was added to avoid the same double encoding problem but with multilang when the verticalUrl is something like 'es/help_articles'.

layouts/html.hbs
yen-tt: got it!

layouts/html.hbs
```diff
@@ -115,12 +115,11 @@
...
       {{#babel}}
       function iframeGetSearchParams() {
-        let params = new URLSearchParams(window.location.search);
-        let verticalUrl = window.location.pathname.substr(1)
-        if (verticalUrl) {
-          params.set('verticalUrl', verticalUrl);
-        }
-        return params.toString();
+        const params = window.location.search.substr(1);
+        let verticalUrl = window.location.pathname.substr(1);
```
oshi97, line 119: this one too :eyes:

yen-tt: ahhh yes yes. my bad

layouts/html.hbs
```diff
@@ -115,12 +115,11 @@
...
-        let params = new URLSearchParams(window.location.search);
-        let verticalUrl = window.location.pathname.substr(1)
-        if (verticalUrl) {
-          params.set('verticalUrl', verticalUrl);
-        }
-        return params.toString();
+        const params = window.location.search.substr(1);
+        const verticalUrl = window.location.pathname.substr(1);
+        return verticalUrl 
+          ? params + '&verticalUrl=' + verticalUrl
```
oshi97, line 121: do we need to add a question mark in front of the params?

oshi97 (approved):
lgtm!

cea2aj (approved):

PR #837, opened 2021-06-16
add analytics to the video card (vimeo + youtube)

oshi97:
This commit adds onPlay analytics to the video card.
The vimeo and youtube iframe video apis are loaded
using code splitting, to not add to the bundle size
for non video card users, and also not block rendering.

J=SLAP-1381
TEST=manual,auto

test that the video analytics are fired on play for
both vimeo and youtube

test clicking play multiple times fires an event EVERY time

test in dev build mode and prod build mode, for ie11 and chrome

tried to add acceptance tests, but the video iframes do not play
well with testcafe. tried even a dummy page with just a single
hardcoded video iframe and I cannot click it in testcafe (either
programmatically or manually through Live mode

check that the vimeo and youtube apis are not loaded multiple
times even with multiple searches/cards)

test that does not send extra analytics events when the video buffers
test that for both vimeo and youtube, going into the video timeline and clicking another point
will trigger an analytics event


static/js/video-apis/youtube/YoutubePlayer.js
```diff
@@ -0,0 +1,28 @@
...
+    this.player = new YT.Player(el, {
+      events: {
+        onStateChange: event => this.onStateChange(event)
+      }
+    });
+    this.playerState = YT.PlayerState.UNSTARTED;
+    this.onPlay = onPlay;
+  }
+
+  onStateChange({ data }) {
```
yen-tt, line 21: If a video switched between states like buffering and playing, wouldn't this keep calling analyticsReporter to send a report that a video is played? Not sure if this would be an issue or something we want

oshi97: yup that's definitely something we need to handle! I missed this

cards/multilang-product-prominentvideo/component.js
```diff
@@ -54,6 +54,35 @@ class multilang_product_prominentvideoCardComponent extends BaseCard['multilang-
...
+    } else if (this.vimeoUrl) {
+      HitchhikerJS.requireVimeoAPI().then(addPlayer);
+    }
+  }
+
+  onPlay() {
+    const event = new ANSWERS.AnalyticsEvent('CTA_CLICK')
+      .addOptions({
+        verticalKey: this.verticalKey,
+        entityId: this.result._raw.id,
```
cea2aj, line 79: nit: I'm wondering if we want to do some null checking here just to be extra careful. Maybe something like `this.result?._raw?.id`

cards/multilang-product-prominentvideo/component.js
```diff
@@ -54,6 +54,35 @@ class multilang_product_prominentvideoCardComponent extends BaseCard['multilang-
     };
   }
 
+  onMount() {
+    const videoSelector = '.js-HitchhikerProductProminentVideo-video';
+    const videoEl = this._container.querySelector(videoSelector);
+    if (!videoEl) {
+      return;
+    }
+    const addPlayer = api => {
```
cea2aj, line 63: what do you think about `videoApi` rather than `api`?

static/js/video-apis/VimeoAPI.js
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Interacts with the Vimeo Player API.
+ */
+class VimeoAPI {
+  /**
+   * Adds a new Vimeo iframe player using the given element.
+   *
+   * @param {HTMLElement} el 
+   * @param {Function} onPlay
```
cea2aj, line 11: nit: I think we should indicate that the second param is an object like:
```
@param {HTMLElement} el 
@param {Object} options
@param {Function) options.onPlay
```
Thoughts?

static/js/video-apis/youtube/loadYoutubeAPI.js
```diff
@@ -0,0 +1,18 @@
+import YoutubeAPI from './YoutubeAPI';
+
+/**
+ * Loads the youtube iframe_api.
+ */
+function loadYoutubeIframeAPI() {
+  return new Promise(resolve => {
+    window.onYouTubeIframeAPIReady = () => resolve(window.YT);
```
cea2aj, line 8: for my own edification, the youtube api calls this function when it loads?

static/js/video-apis/youtube/loadYoutubeAPI.js
```diff
@@ -0,0 +1,18 @@
...
+    const tag = document.createElement('script');
+    tag.src = 'https://www.youtube.com/iframe_api';
+    const firstScriptTag = document.getElementsByTagName('script')[0];
+    firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
+  });
+}
+
+const youtubeApiPromise =
```
cea2aj, line 16: Is it possible to simplify this file by making it simply return a promsise? Or do we need the `loadYoutubeIframeAPI` function? If so, maybe we could rename this file to `loadYoutubeApiPromise`

cards/multilang-product-prominentvideo/component.js
oshi97: shouldn't hurt!

static/js/video-apis/VimeoAPI.js
oshi97: will call the param param1 instead of options since that's what my vscode spit out

static/js/video-apis/youtube/loadYoutubeAPI.js
oshi97: yes
https://developers.google.com/youtube/iframe_api_reference#Requirements

oshi97: unfortunately it has to live on the window

static/js/video-apis/youtube/loadYoutubeAPI.js
oshi97: if we make it a function that returns a promise, we'll generate a new promise each time instead of returning the same one
rename makes sense though!

oshi97: Can simplify the promise quite a bit. Having the function that just loads the iframe_api is from when that was a separate file's export

static/js/video-apis/youtube/YoutubePlayer.js
```diff
@@ -0,0 +1,47 @@
...
+    this.onPlay = onPlay;
+  }
+
+  /**
+   * @param {Object} param0 
+   * @param {PlayerState} param0.data
+   */
+  onStateChange({ data }) {
+    const isPlayingVideo = this.videoIsPlayingForState(data);
+    const wasAlreadyPlayingVideo = this.videoIsPlayingForState(this.playerState);
```
cea2aj, line 28: would it be simpler to have a boolean value which keeps track of whether or not the video has started?

oshi97: we need to handle pausing the video, ending the video, cueing a new video

cea2aj: oh, we want to send a new analytics event every time someone pauses and hits play again?

cea2aj (approved):

static/js/video-apis/youtube/loadYoutubeAPI.js
cea2aj: love the refactor!

PR #839, opened 2021-06-17
Consistent Breakpoints in Theme

cjiang2000:
Changed most @media to @include breakpoints instead. Also moved answers-variable imports into common/variables. Changed $screen-xs-max value to $breakpoint-mobile-max and $screen-sm-min value to $breakpoint-mobile-min.

J=SLAP-1277
TEST=manual
Built the test site and checked on Chrome and IE11 for any unusual activity with varying window sizes.

### DIFF
--- a/static/scss/answers/common/variables.scss
+++ b/static/scss/answers/common/variables.scss
@@ -1,3 +1,6 @@
+@import '../answers-variables-default';
+@import '../../answers-variables';
+
 $grid-gutter-width: 32px !default;
 $grid-gutter-width-xs: $grid-gutter-width !default;
 $grid-gutter-width-sm: $grid-gutter-width !default;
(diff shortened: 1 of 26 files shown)


static/scss/answers/common/variables.scss
```diff
@@ -9,8 +9,8 @@ $vertical-margin-top: $grid-gutter-width !default;
 $vertical-margin-bottom: $grid-gutter-width !default;
 
 $screen-xs-min: 480px;
-$screen-xs-max: 767px;
-$screen-sm-min: 768px;
+$screen-xs-max: 768px;
```
oshi97, line 12: did we mean to change screen-xs-max from 767 to 768?

cjiang2000: no, but it doesn't make a difference because we are no longer using those values in utilitymixins. I'll change it back though just for more clarity.

cjiang2000: no, but it doesn't make a difference because we are no longer using those values in utilitymixins. I'll change it back though just for more clarity.

oshi97 (approved):
lgtm! 🥼

PR #844, opened 2021-06-22
RuntimeConfig Setters for analyticsEventsEnabled

yen-tt:
Theme's runtime config setters for analyticsEventsEnabled update SDK's config accordingly after Answers is initialized

- update runtimeConfig to support general and key-specific listener with event type
- changed AnswersExperience to a class
- provide runtimeConfig a key-specific listener to handle updates on 'analyticsEventsEnabled' from AnswersExperience - a deferred promise is placed in the window (from html.hbs), resolve after Answers is initialized (wait duration of 5 secs) and call SDK's analytics setter, reject if Answers failed to initialize.
- config updates from AnswersExperienceFrame send message to iframe and handle by RuntimeConfigReceiver. Since this invokes runtimeConfig.set from AnswersExperience, the same key-specific listener will be used.

J=SLAP-1393
TEST=manual & auto

Launched hitchhiker theme test site, called runtimeConfig.set for analyticsEventsEnabled in iframe and non-iframe page. Set analyticsEventsEnabled to false/true, and see that console.log from SDK's analytics setter printed the correct result. Click thumbs up on direct answer card, and confirm a network call is made to realtimeanalytics when set to true and no feedback got send when set to false.
Update and test general/key-specific listener in jest

### DIFF
--- a/static/js/HitchhikerJS.js
+++ b/static/js/HitchhikerJS.js
@@ -34,12 +34,6 @@ export { ManualInitializer };
 import RuntimeConfigReceiver from './runtime-config-receiver';
 export { RuntimeConfigReceiver };
 
-import DeferredPromise from './deferred-promise';
-window.AnswersInitialized = new DeferredPromise((resolve, reject) => {
-  setTimeout(reject, 5000, 
-    'Timed out waiting on Answers initialization. Unable to update Answer\'s configuration(s).');
-});
-
 import RuntimeConfig from './runtime-config';
 import AnswersExperience from './answers-experience';
 const runtimeConfig = new RuntimeConfig();
--- a/static/js/HitchhikerJS.js
+++ b/static/js/HitchhikerJS.js
@@ -34,9 +34,7 @@ export { ManualInitializer };
 import RuntimeConfigReceiver from './runtime-config-receiver';
 export { RuntimeConfigReceiver };
 
-import RuntimeConfig from './runtime-config';
-window.AnswersExperience = {
-  runtimeConfig: new RuntimeConfig()
-};
+import AnswersExperience from './answers-experience';
+window.AnswersExperience = AnswersExperience;
 
 export * from './video-apis';
--- a/static/js/HitchhikerJS.js
+++ b/static/js/HitchhikerJS.js
@@ -37,7 +37,6 @@ export { RuntimeConfigReceiver };
 import RuntimeConfig from './runtime-config';
 import AnswersExperience from './answers-experience';
 const runtimeConfig = new RuntimeConfig();
-const answersExperience = new AnswersExperience(runtimeConfig);
-window.AnswersExperience = answersExperience;
+window.AnswersExperience = new AnswersExperience(runtimeConfig);
 
 export * from './video-apis';
--- a/static/js/HitchhikerJS.js
+++ b/static/js/HitchhikerJS.js
@@ -34,7 +34,16 @@ export { ManualInitializer };
 import RuntimeConfigReceiver from './runtime-config-receiver';
 export { RuntimeConfigReceiver };
 
+import DeferredPromise from './deferred-promise';
+window.AnswersInitialized = new DeferredPromise((resolve, reject) => {
+  setTimeout(reject, 5000, 
+    'Timed out waiting on Answers initialization. Unable to update Answer\'s configuration(s).');
+});
+
+import RuntimeConfig from './runtime-config';
 import AnswersExperience from './answers-experience';
-window.AnswersExperience = AnswersExperience;
+const runtimeConfig = new RuntimeConfig();
+const answersExperience = new AnswersExperience(runtimeConfig);
+window.AnswersExperience = answersExperience;
 
 export * from './video-apis';
--- a/static/js/answers-experience-frame.js
+++ b/static/js/answers-experience-frame.js
@@ -4,9 +4,11 @@ export default class AnswersExperienceFrame {
   constructor (runtimeConfig) {
     this.runtimeConfig = runtimeConfig;
     this._hasManuallyInitialized = false;
-
-    runtimeConfig._onUpdate(updatedConfig => {
-      sendToIframe({ runtimeConfig: updatedConfig });
+    
+    Object.entries(runtimeConfig.getAll()).forEach(([key, value]) => {
+      runtimeConfig._onUpdate(key, configOption => {
+        sendToIframe({ runtimeConfig: {key: configOption} });
+      });
     });
   }
 
--- a/static/js/answers-experience-frame.js
+++ b/static/js/answers-experience-frame.js
@@ -6,7 +6,6 @@ export default class AnswersExperienceFrame {
     this._hasManuallyInitialized = false;
 
     runtimeConfig.registerListener({
-      eventType: 'update',
       callback: updatedConfig => {
         sendToIframe({ runtimeConfig: updatedConfig });
       }
--- a/static/js/answers-experience-frame.js
+++ b/static/js/answers-experience-frame.js
@@ -4,11 +4,12 @@ export default class AnswersExperienceFrame {
   constructor (runtimeConfig) {
     this.runtimeConfig = runtimeConfig;
     this._hasManuallyInitialized = false;
-    
-    Object.entries(runtimeConfig.getAll()).forEach(([key, value]) => {
-      runtimeConfig._onUpdate(key, configOption => {
-        sendToIframe({ runtimeConfig: {key: configOption} });
-      });
+
+    runtimeConfig.registerListener({
+      eventType: 'update',
+      callback: updatedConfig => {
+        sendToIframe({ runtimeConfig: updatedConfig });
+      }
     });
   }
 
--- a/static/js/answers-experience.js
+++ b/static/js/answers-experience.js
@@ -6,7 +6,6 @@ export default class AnswersExperience {
     this.AnswersInitializedPromise = new DeferredPromise();
 
     runtimeConfig.registerListener({
-      eventType: 'update',
       key: 'analyticsEventsEnabled',
       callback: updatedConfigOption => {
         this.AnswersInitializedPromise
--- a/static/js/answers-experience.js
+++ b/static/js/answers-experience.js
@@ -1,22 +1,36 @@
+import DeferredPromise from './deferred-promise';
+
 export default class AnswersExperience {
   constructor (runtimeConfig) {
     this.runtimeConfig = runtimeConfig;
 
-    runtimeConfig._onUpdate(updatedConfig => {
-      window.AnswersInitialized
-        .then(() => updateAnswersConfig(updatedConfig))
-        .catch(err => console.warn(err));
+    this.AnswersInitializedPromise = new DeferredPromise((resolve, reject) => {
+      setTimeout(reject, 5000, 
+        'Timed out waiting on Answers initialization. Unable to update Answer\'s configuration(s).');
     });
 
-		/**
-     * Update Answer's runtime configurations
-     * 
-     * @param {Object} updatedConfig An object representation of the runtime config
-     */
-    function updateAnswersConfig(updatedConfig) {
-      if (updatedConfig['analyticsEventsEnabled']) {
-        ANSWERS.setAnalyticsOptIn(updatedConfig['analyticsEventsEnabled'] === 'true');
-      }
+    runtimeConfig._onUpdate('analyticsEventsEnabled', updatedConfigOption => {
+      this.AnswersInitializedPromise
+        .then(() => this.updateAnswersAnalyticsEventsConfig(updatedConfigOption))
+        .catch(err => console.warn(err));
+    });
+  }
+
+  /**
+   * Update Answer's analytics events config based on new value 
+   * of 'analyticsEventsEnabled' key in runtimeConfig
+   * 
+   * @param {string|boolean} updatedConfigOption
+   */
+  updateAnswersAnalyticsEventsConfig (updatedConfigOption) {
+    let option = null;
+    if (typeof(updatedConfigOption) === 'string') {
+      option = updatedConfigOption.toLowerCase() === 'true';
+    } else if (typeof(updatedConfigOption) === 'boolean') {
+      option = updatedConfigOption;
+    }
+    if (option != null) {
+      ANSWERS.setAnalyticsOptIn(option);
     }
   }
 }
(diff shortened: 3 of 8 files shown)


static/js/answers-experience.js
```diff
@@ -0,0 +1,51 @@
+import RuntimeConfig from './runtime-config';
+
+/**
+ * Check if Answers is initialized in window in a timed loop
+ * 
+ * @returns {Promise} When resolved, Answers exists in window
+ */
+function waitForAnswersInit() {
```
cea2aj, line 8: I think we need to do more than just wait for the Answers instance to appear on the window because the setAnalyticsOptIn function depends on the `_analyticsReporterService` which isn't set until Answers initializes. One idea is that we could put a promise on the window such as `AnswersInitialized` in layouts.hbs, and in core.hbs, we can resolve that promise inside of the Answers onReady. So rather than having a `waitForAnswersInit` function, we could use `window.AnswersInitialized`

static/js/answers-experience.js
```diff
@@ -0,0 +1,51 @@
...
+
+const runtimeConfig = new RuntimeConfig();
+
+runtimeConfig._onUpdate(updatedConfig => {
+  waitForAnswersInit()
+    .then(() => updateAnswersConfig(updatedConfig))
+    .catch(err => console.warn(err));
+});
+
+const AnswersExperience = {
```
cea2aj, line 47: If we switch this to a class, we can pass in a RuntimeConfig object into the constructor. What do you think of that approach?

static/js/answers-experience.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+function updateAnswersConfig(updatedConfig) {
+  if (updatedConfig['analyticsEventsEnabled']) {
+    ANSWERS.setAnalyticsOptIn(updatedConfig['analyticsEventsEnabled'] === 'true');
+  }
+}
+
+const runtimeConfig = new RuntimeConfig();
+
+runtimeConfig._onUpdate(updatedConfig => {
```
cea2aj, line 41: what do you think about putting this function in its own file such as handleRuntimeConfigUpdates.js?

static/js/answers-experience.js
oshi97: I like this suggestion! we can even use a fancy "Answers init returns a promise" that somebody added a while back and set that as the promise's value :^)

static/js/answers-experience.js
```diff
@@ -0,0 +1,51 @@
...
+    }, intervalLength);
+  });
+}
+
+/**
+ * Update Answer's runtime configurations
+ * 
+ * @param {Object} updatedConfig An object representation of the runtime config
+ */
+function updateAnswersConfig(updatedConfig) {
```
oshi97, line 33: It's a little bit strange to call this "updateAnswersConfig" when this doesn't actually update the internal state of the RuntimeConfig. In my head what this is really doing is listening for updates to a specific key, 'analyticsEventsEnabled', and adding a listener to update the SDK when such a change occurs. Would it be overkill to add the concept of listeners to RuntimeConfig?

oshi97: Ah it looks like there's already an _onUpdate() method, so maybe it's not so crazy to let you listen to a specific key?

static/js/answers-experience.js
```diff
@@ -0,0 +1,51 @@
...
+  if (updatedConfig['analyticsEventsEnabled']) {
+    ANSWERS.setAnalyticsOptIn(updatedConfig['analyticsEventsEnabled'] === 'true');
```
oshi97, line 35: will the value always be the string `'true'` and not the boolean value `true`?

static/js/answers-experience.js
yen-tt: I didn't realize that Answers is put on window before it's initialized but yeah connor's idea sounds good! I just updated following his suggestion. I didn't see your comment beforehand oliver, what do you mean by answers init returns a promise? like in ANSWERS.init pr initAnswers()?

static/js/answers-experience.js
yen-tt: Yup! converted to AnswersExperience class

static/js/answers-experience.js
yen-tt: I configured this _onUpdate method inside AnswersExperience class (similar structure to AnswersExperienceFrame). How does that sound?

static/js/answers-experience.js
yen-tt: From the ticket description, it seems like it will be in string form, but I'm actually not sure if that's set in stone.

static/js/answers-experience.js
yen-tt: hmm i was thinking the naming implies updating configurations of answers object, not that it changes the runtimeconfig object of the site, if I'm understanding your comment right? But yeah we could change runtimeconfig to have individual listeners for keys so it doesn't need to run through all options on a single key/value change.

static/js/answers-experience.js
oshi97: Yup! ANSWERS.init() actually returns a promise. I think this was Connor's first item https://github.com/yext/answers-search-ui/pull/996

static/js/deferred-promise.js
```diff
@@ -0,0 +1,15 @@
+export default class DeferredPromise {
```
cea2aj, line 1: This is a really cool! Can you add a jsdoc to give an overview of the class?

static/js/answers-experience.js
```diff
@@ -0,0 +1,22 @@
...
+  constructor (runtimeConfig) {
+    this.runtimeConfig = runtimeConfig;
+
+    runtimeConfig._onUpdate(updatedConfig => {
+      window.AnswersInitialized
+        .then(() => updateAnswersConfig(updatedConfig))
+        .catch(err => console.warn(err));
+    });
+
+		/**
```
cea2aj, line 11: nit: spacing

static/js/answers-experience.js
cea2aj: that makes sense!

static/js/HitchhikerJS.js
```diff
@@ -34,9 +34,16 @@ export { ManualInitializer };
 import RuntimeConfigReceiver from './runtime-config-receiver';
 export { RuntimeConfigReceiver };
 
+import DeferredPromise from './deferred-promise';
+window.AnswersInitialized = new DeferredPromise((resolve, reject) => {
```
cea2aj, line 38: What do you think of putting this AnswersInitialized promise inside of `AnswersExperience` rather than on the window?

cea2aj: Do you think it would be clearer if we renamed `AnswersInitialized` to `AnswersInitializedPromise`?

static/js/answers-experience.js
yen-tt: updated to individual key listeners instead of all!

static/js/answers-experience-frame.js
```diff
@@ -4,9 +4,11 @@ export default class AnswersExperienceFrame {
...
     this.runtimeConfig = runtimeConfig;
     this._hasManuallyInitialized = false;
-
-    runtimeConfig._onUpdate(updatedConfig => {
-      sendToIframe({ runtimeConfig: updatedConfig });
+    
+    Object.entries(runtimeConfig.getAll()).forEach(([key, value]) => {
+      runtimeConfig._onUpdate(key, configOption => {
+        sendToIframe({ runtimeConfig: {key: configOption} });
+      });
```
cea2aj, line 11: When someone sets a new `runtimeConfig` options that wasn't set when `AnswersExperienceFrame` was constructed, I don't think it will be sent to the iframe because it was never registered to listen to that key. So I think it would be best if we make the key-specific listeners an additive change.

static/js/runtime-config.js
```diff
@@ -32,15 +32,19 @@ export default class RuntimeConfig {
...
 
   /**
-   * A function which is called any time the RuntimeConfig is updated
+   * A function which is called any time a specific key in RuntimeConfig is updated
+   * @param {string} key
    * @param {Function} cb 
    */
-  _onUpdate (cb) {
-    this._onUpdateListener = cb;
+  _onUpdate (key, cb) {
```
cea2aj, line 46: If we do refactor such that` _onUpdate` stays, the same, I think we should use the function name `registerListener` for registering single listeners so that we are more consistent with storage in Answers: https://github.com/yext/answers-search-ui/blob/master/src/core/storage/storage.js.

static/js/runtime-config.js
```diff
@@ -32,15 +32,19 @@ export default class RuntimeConfig {
   set (key, value) {
     this._validateSet(key, value);
     this._data[key] = value;
-    this._onUpdateListener(this.getAll());
+    if (!this._onUpdateListeners[key]) {
+      this._onUpdate(key, () => {});
+    }
+    this._onUpdateListeners[key](value);
```
cea2aj, line 38: Is there a reason for setting the listener if it doesn't exist rather than having the logic call the listener only if it exists?

static/js/HitchhikerJS.js
```diff
@@ -35,8 +35,9 @@ import RuntimeConfigReceiver from './runtime-config-receiver';
 export { RuntimeConfigReceiver };
 
 import RuntimeConfig from './runtime-config';
-window.AnswersExperience = {
-  runtimeConfig: new RuntimeConfig()
-};
+import AnswersExperience from './answers-experience';
+const runtimeConfig = new RuntimeConfig();
+const answersExperience = new AnswersExperience(runtimeConfig);
+window.AnswersExperience = answersExperience;
```
cea2aj, line 41: nit: we could just have `window.AnswersExperience = new AnswersExperience(runtimeConfig);`

static/js/answers-experience.js
```diff
@@ -0,0 +1,36 @@
+import DeferredPromise from './deferred-promise';
+
+export default class AnswersExperience {
+  constructor (runtimeConfig) {
+    this.runtimeConfig = runtimeConfig;
+
+    this.AnswersInitializedPromise = new DeferredPromise((resolve, reject) => {
+      setTimeout(reject, 5000,
```
cea2aj, line 8: Do we need to reject it after 5 seconds?

static/js/answers-experience-frame.js
yen-tt: right. I should update runtime-config.js to support both `generalUpdateListener` and `keySpecificListeners`. AnswersExperienceFrame would need a generalUpdateListener to catch any types of key update to send to iframe, and AnswersExperience will need keySpecificListeners for handling specific key set from runtime config receiver. That sounds ok?

static/js/answers-experience.js
yen-tt: I guess we could leave it as pending forever until resolve? I put it there to clean up the promise and give some sort of log for user if answers is not initialize.

static/js/runtime-config.js
yen-tt: ah yea, good point..I will update that!

static/js/answers-experience-frame.js
cea2aj: Ya, that sounds good!

static/js/answers-experience.js
cea2aj: What are your thoughts on having the reject called if the Answers init fails, rather than rejecting after 5 seconds? I think that should be possible by adding a `.catch` to the `Answer.init()` call

yen-tt: oh yeah, that works!

static/js/runtime-config.js
yen-tt: What do you think about using `_onUpdate (cb, key = null)` for both general and specific key listeners (if key is specified). I could keep `_onUpdate` for the general listener and `registerListener` for the specific one, but would that be confusing or easy to mix up?

cea2aj: What do you think about `registerListener({ event: 'update', key: '_linkTarget', cb: () => {} });` for key specific listeners and `registerListener({ event: 'update', cb: () => {} });` for generic update listeners?

static/js/runtime-config.js
```diff
@@ -4,7 +4,7 @@
 export default class RuntimeConfig {
   constructor (initialConfig = {}) {
     this._data = { ...initialConfig };
-    this._onUpdateListener = () => {};
+    this._listeners = [];
```
oshi97, line 7: would it make sense for this to be a map of keys to listeners? I think the SDK just has an array of objects but the listener lookup might be slightly faster if we use a map (though would probably be uglier to read and also pretty negligible). The keyless listeners would probably have to go into their own variable if we use a map instead (or be set to like the blank string key).

static/js/answers-experience.js
```diff
@@ -0,0 +1,36 @@
+import DeferredPromise from './deferred-promise';
+
+export default class AnswersExperience {
+  constructor (runtimeConfig) {
+    this.runtimeConfig = runtimeConfig;
+    this.AnswersInitializedPromise = new DeferredPromise();
+
+    runtimeConfig.registerListener({
+      eventType: 'update',
```
oshi97, line 9: what do you think about not including eventType for now? I think it might make the code slightly simpler, and it's not something we'll be prevented from adding in the future (we can just set the default to 'update' or just refactor).

static/js/runtime-config.js
```diff
@@ -32,19 +32,41 @@ export default class RuntimeConfig {
...
-    if (!this._onUpdateListeners[key]) {
-      this._onUpdate(key, () => {});
+    this._callListeners('update', key);
+  }
+
+  /**
+   * Adds a listener for a given key, or any time RuntimeConfig is 
+   * updated if there's no key provided.
+   *
+   * @param {*} listener the listener to add
```
cea2aj, line 42: I think we typically but Object here instead of `*`

static/js/runtime-config.js
```diff
@@ -4,7 +4,18 @@
...
-    this._onUpdateListener = () => {};
+    
+    /**
+     * Hold listeners that applies to all keys
+     * @type {Object[]}
+     */
+    this._generalListeners = [];
+    
+    /**
+     * Hold listeners that applies to specific keys
```
oshi97, line 15: super nit but maybe we remove the word Hold? (unrelated but I think it should just be apply not applies)
so maybe `Listeners that apply to specific keys`, and same with generalListeners?

oshi97 (commented):
one jsdoc nit besides that lgtm!

oshi97 (approved):
🌐 🌐 🌐

cea2aj (approved):

PR #845, opened 2021-06-22
Remove _fireOnDomLoad

cea2aj:
We can remove the `_fireOnDomLoad` from the manual initializer because the setup function itself is called during `DOMContentLoaded`

J=none
TEST=manual

Smoke test runtime config and manual initialization

### DIFF
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -6,15 +6,6 @@
       if (window.name == 'overlay') {
         window.isOverlay = true;
       }
-      function onDocumentReady(cb) {
-        if (document.readyState === 'complete' 
-          || document.readyState === 'loaded'
-          || document.readyState === 'interactive') {
-          cb.bind(this)();
-        } else {
-          document.addEventListener('DOMContentLoaded', cb.bind(this));
-        }
-      }
       {{/babel}}
     </script>
     {{> layouts/headincludes }}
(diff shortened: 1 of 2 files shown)


layouts/html.hbs
```diff
@@ -6,6 +6,14 @@
       if (window.name == 'overlay') {
         window.isOverlay = true;
       }
+      function onDocumentReady(cb) {
+        if (document.readyState === "complete"
```
tmeyer2115, line 10: Is it possible we will ever invoke `onDocumentReady` when the document is already loaded? I'm wondering if we need this condition or if the `addEventListener` will suffice.

cea2aj: I'm using it inside the iframeResizer onMessage, so we do need to check the ready state. Also I think it makes it a little more robust overall

tmeyer2115 (approved):

layouts/html.hbs
```diff
@@ -6,6 +6,14 @@
...
+          || document.readyState === "loaded") {
```
oshi97, line 11: nit can we do single quotes

cea2aj: sure. Updated!

oshi97 (approved):

tmeyer2115 (approved):

oshi97 (approved):

PR #853, opened 2021-06-28
Runtime config acceptance test

cea2aj:
An acceptance test for testing runtime config behavior

This test confirms that runtime config set in a parent site is passed to the child runtime config (inside the iframe)

J=SLAP-1371
TEST=acceptance


tests/acceptance/index.js
```diff
@@ -28,7 +28,7 @@ async function runTests (browsers) {
     const numberTestsFailed = await testcafe.createRunner()
       .src('tests/acceptance/suites/*.js')
       .browsers(browsers)
-      .startApp(`npx serve -p ${PORT} test-site/public`, 4000)
+      .startApp(`npx serve -l tcp://0.0.0.0:${PORT} test-site/public`, 4000)
```
oshi97, line 31: was this to fix acceptance tests not running locally?

tests/acceptance/blocks/childruntimeconfig.js
```diff
@@ -0,0 +1,30 @@
...
+    })(key, value);
+  }
+
+  /**
+   * Gets a value of the runtime config
+   * @param {string} key 
+   * @returns {*}
+   */
+  async get(key) {
+    return await ClientFunction((key) => {
```
oshi97, line 24: I think we teeechnically don't need this await

tests/acceptance/blocks/parentruntimeconfig.js
```diff
@@ -0,0 +1,30 @@
+import { ClientFunction } from 'testcafe';
+
+/**
+ * Provides interaction with the runtime config located in the AnswersExperienceFrame
+ */
+class ParentRuntimeConfig {
```
tmeyer2115, line 6: This may not be possible, but I'm wondering if we could consolidate down to something like:

``` 
class RuntimeConfigWrapper {
   constructor(runtimeConfig) {
   }
   
   async set(key, value) {
      await ClientFunction((key, value) => this.runtimeConfig(key, value));
    }
}
```

Then, instead of having two very similar classes in `ParentRuntimeConfig` and `ChildRuntimeConfig`, we'd have: 
```
const parentRuntimeConfig = new RuntimeConfigWrapper(AnswersExperienceFrame.runtimeConfig);
const childRuntimeConfig = new RuntimeConfigWrapper(AnswersExperience.runtimeConfig);
```

tests/acceptance/index.js
cea2aj: Ya, this helps my networking problem locally, but it doesn't affect anything else

tests/acceptance/blocks/parentruntimeconfig.js
cea2aj: Because this is a ClientFunction, I don't think that will work as written because I don't think ClientFunction can be passed a non-serializable parameters such as a function. However, I could pass a boolean flag and toggle the behavior that way to consolidate the RuntimeConfig classes

tests/acceptance/blocks/childruntimeconfig.js
cea2aj: you're right! removing

tests/acceptance/blocks/parentruntimeconfig.js
cea2aj: updated to include an `isParent` parameter in the constructor

tmeyer2115 (approved):

PR #855, opened 2021-06-28
AnswersExperienceFrame Unit tests

cea2aj:
Add tests for AnswersExperienceFrame

I created a jest mock for iframe-common so that we could check the interaction with the `sendToIframe` function

J=SLAP-1371
TEST=auto


tmeyer2115 (approved):

oshi97 (approved):
I like the mocks!

PR #856, opened 2021-06-28
Add RTF link click analytics for Direct Answer cards.

tmeyer2115:
This PR ensures that all Direct Answer cards in the Theme support RTF link
click analytics. The code was taken largely from the SDK. The main difference
is the source of some of the event options.

J=SLAP-1405
TEST=manual

Built a local test site with the new base `card_component`. Saw the correct
`AnalyticsEvent`s getting fired when I clicked on a RTF DA's links. Verified
that clicking plain text in the RTF value did not cause an event to be
fired.

### DIFF
--- a/directanswercards/card_component.js
+++ b/directanswercards/card_component.js
@@ -67,6 +67,9 @@ BaseDirectAnswerCard["{{componentName}}"] = class extends ANSWERS.Component {
         });
       }
     }
+
+    const rtfElement = this._container.querySelector('.js-yxt-rtfValue');
+    rtfElement && rtfElement.addEventListener('click', e => this._handleRtfClickAnalytics(e));
   }
 
   /**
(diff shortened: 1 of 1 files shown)


directanswercards/card_component.js
```diff
@@ -67,6 +67,9 @@ BaseDirectAnswerCard["{{componentName}}"] = class extends ANSWERS.Component {
         });
       }
     }
+
+    const rtfElement = this._container.querySelector('.js-yxt-rtfValue');
```
oshi97, line 71: the SDK does the same thing, but just for my edification are we only adding an event listener to the first instance of .js-yxt-rtfValue?

tmeyer2115: Yes, I believe so. This is how things work in the SDK as well. IIRC, we were told at the time not to worry about multiple RTF fields on a card.

oshi97 (approved):

PR #857, opened 2021-06-28
add translation test

oshi97:
Adds a translation test github action, that creates a messages.pot using jambo extract-translations
and then checks that every .po file matches messages.pot.
This is the same logic that the SDK's translation test uses.
Also adds an Italian translation that was missing, so that tests pass.
Previously, the translation for `vertical-full-page-map-alternative-verticals/alternativeverticals.hbs:47` was used 
for `vertical-full-page-map-alternative-verticals/alternativeverticals.hbs:5` instead of the actual line 5 translation,
causing the translation for line 47 to be doubled and the translation for line 5 to be missing.

J=SLAP-1378
TEST=auto

checked that the test will run on master and release branches


.github/workflows/translation-test.yml
```diff
@@ -0,0 +1,30 @@
+name: Translation Test
+
+on: push
```
tmeyer2115, line 3: Is this action currently running on any push to a branch? I'm wondering if we can be more selective about when this action is run. We definitely want to make sure we don't merge anything into master that's missing translations. But, it's possible that we merge some code into `develop`, say, that doesn't have translations yet. We may have finished the feature but are waiting on Smartling for the translations.

oshi97: updated to only run on release branches and master!

.github/workflows/translation-test.yml
```diff
@@ -0,0 +1,35 @@
+name: Translation Test
+
+
```
cea2aj, line 3: nit: is there an extra space here?

cea2aj (approved):

.github/workflows/translation-test.yml
oshi97: yikes, updated!

PR #858, opened 2021-06-28
Manual initializer tests

cea2aj:
Unit tests for the manual initializer

J=SLAP-1371
TEST=auto

### DIFF
--- a/tests/static/js/manual-initializer.js
+++ b/tests/static/js/manual-initializer.js
@@ -4,7 +4,7 @@ describe('The manual initializer works', () => {
   let initAnswers;
 
   beforeEach(() => {
-    global.window.AnswersExperience = {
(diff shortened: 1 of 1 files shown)


tests/static/js/manual-initializer.js
```diff
@@ -0,0 +1,32 @@
+import ManualInitializer from '../../../static/js/manual-initializer';
+
+describe('The manual initializer works', () => {
+  let initAnswers;
+
+  beforeEach(() => {
+    global.window.AnswersExperience = {
```
oshi97, line 7: is the global necessary?

cea2aj: based on the is-staging tests, I thought it was but it turns out it's not. Updated

oshi97 (approved):

PR #861, opened 2021-06-29
Locator time debouncing

cea2aj:
Debounce the locator based on time

This PR adds a time-based debouncer so that `searchThisArea` will not execute until 250ms have passed since the last time the debounce function was called. This prevents unnecessary searches from being ran when someone quickly zooms in or out of the map.

For this change, I renamed the `SearchDebouncer` to the `SearchPreventer`since it is a more accurate name. The distinction is that if a search doesn't pass the criteria of the `SearchPreventer`, it will never be ran. Debouncing is different because when a debounced function is called, it will eventually execute after sufficient time has passed without another function call.

I was hoping that this change would also reduce the number of searches made when dragging the map, however the debounce time necessary to make that work is too large and would lead to a sluggish user experience. It is possible that we would be able to accomplish it in a non-sluggish manner by reseting the debouncer whenever the user clicks on the map canvas, however further experimentation would be necessary.

I opted not to debounce `searchThisArea()` calls after a user clicks on the 'Search This Area` button or on a cluster since there is little need for a debouncer there.

J=SLAP-1301
TEST=manual

Test that zooming in the map quickly doesn't trigger a new search until the zooming is done

### DIFF
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -1,7 +1,7 @@
 import { Coordinate } from './Geo/Coordinate.js';
 import { smoothScroll } from './Util/SmoothScroll.js';
-import { getLanguageForProvider, isViewableWithinContainer, removeElement } from './Util/helpers.js';
-import { SearchDebouncer } from './SearchDebouncer';
+import { isViewableWithinContainer, removeElement, debounce } from './Util/helpers.js';
+import { SearchPreventer } from './SearchPreventer';
 import { defaultCenterCoordinate } from './constants.js';
 
 import ZoomTriggers from './Maps/ZoomTriggers.js';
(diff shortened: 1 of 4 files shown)


static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -1,7 +1,7 @@
 import { Coordinate } from './Geo/Coordinate.js';
 import { smoothScroll } from './Util/SmoothScroll.js';
-import { getLanguageForProvider, isViewableWithinContainer, removeElement } from './Util/helpers.js';
```
cea2aj, line 3: getLanguageForProvider was not being used in this file so I removed it

tmeyer2115 (approved):

oshi97:
percy went splat [here](https://percy.io/cd93112c/answers-hitchhiker-theme/builds/11210427/changed/629453896?browser=chrome&browser_ids=17%2C19%2C20&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=375&widths=375%2C1280) probably not our fault though?

cea2aj:
> percy went splat [here](https://percy.io/cd93112c/answers-hitchhiker-theme/builds/11210427/changed/629453896?browser=chrome&browser_ids=17%2C19%2C20&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=375&widths=375%2C1280) probably not our fault though?

ya, that was just percy going splat. I'll keep an eye on that in the future

tmeyer2115 (approved):

PR #865, opened 2021-06-29
Overlay button legacy bundle

cea2aj:
Improve development build time and web performance by adding a button legacy build

Create a legacy bundle build similar to #828 so that we can avoid the ponyfills and polyfills for modern web browsers. This yeids the following benefits:
- Improved webpack build time by about 13% from 4.5 seconds to 4 seconds when IS_DEVELOPMENT_PREVIEW=true
- Modern browsers are served a much smaller overlay-button bundle (32KB rather than the 211KB of the legacy bundle)

Since specifying type="module" on the  modern bundle makes it deferred, we must wait for DOMContentLoaded before trying to access `window.OverlayButtonJS`

J=SLAP-1417
TEST=manual

Smoke test the overlay on chrome and on IE11 on browserstack.

### DIFF
--- a/static/js/overlay/button-frame/button.html
+++ b/static/js/overlay/button-frame/button.html
@@ -1,12 +1,15 @@
 <!DOCTYPE html>
 <html>
 <head data-iframe-height="true">
-  <script src="overlay-button.js"></script>
+  <script src="overlay-button.js" type="module"></script>
+  <script src="overlay-button-legacy.js" nomodule></script>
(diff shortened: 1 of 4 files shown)


tmeyer2115 (approved):
We'll need to give Rose the heads up here so she can call out the potential upgrade pitfall. If someone forked the `button.html`, they could be in trouble

oshi97 (approved):

PR #866, opened 2021-06-29
Webpack Bundle Analyzer

cea2aj:
Add the Webpack Bundle Analyzer which can be enabled by setting the environment variable ANALYZE_BUNDLE=true

J=none
TEST=manual

### DIFF
--- a/static/webpack-config.js
+++ b/static/webpack-config.js
@@ -6,6 +6,7 @@ const HtmlPlugin = require('html-webpack-plugin');
 const RemovePlugin = require('remove-files-webpack-plugin');
 const { merge } = require('webpack-merge');
 const { parse } = require('comment-json');
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 
 module.exports = function () {
   const jamboConfig = require('./jambo.json');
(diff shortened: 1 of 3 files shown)


static/webpack-config.js
```diff
@@ -6,6 +6,7 @@ const HtmlPlugin = require('html-webpack-plugin');
 const RemovePlugin = require('remove-files-webpack-plugin');
 const { merge } = require('webpack-merge');
 const { parse } = require('comment-json');
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
```
oshi97, line 9: this is probably not going to change much, but if you move the require into the if statement does the build speed for regular builds change at all?

cea2aj: that probably would help out a little

static/webpack-config.js
```diff
@@ -66,6 +67,14 @@ module.exports = function () {
     })
   ];
 
+  const analyzeBundle = (process.env || {}).ANALYZE_BUNDLE === 'true';
+
+  if (analyzeBundle) {
+    plugins.push(new BundleAnalyzerPlugin({
+      analyzerMode: 'static',
```
oshi97, line 74: oh shoot I missed this the first time around, but did you say that the server mode is better? (I've only tried the static mode)

cea2aj: I'm pretty sure the server mode just serves the same HTML page that you get with the static. I was having some issues with the server mode but static worked for me

static/webpack-config.js
oshi97: ok I'm down to do that then!

cea2aj:
Closing because we decided not to include this functionality with the theme out of the box

PR #869, opened 2021-06-30
Runtime Config Session Tracking Support

cea2aj:
Add support for 'sessionTrackingEnabled' in the runtime config

This PR moves the canonicalization of the runtime config listeners into RuntimeConfig itself. The waiting for Answers to initialize is also moved out of the listeners and into `AnswersExperience`.

Note: Currently when sessionTrackingEnabled is set, the ordering of the verticals on the universal results page changes. I suspect this is an issue with the answers-search-ui and therefore the issue is not related to this PR.

J=SLAP-1410
TEST=manual

Test both setting both 'analyticsEventsEnabled' and 'sessionTrackingEnabled' from an iframe integration and a standard integration


static/js/runtime-config-listeners/analytics.js
```diff
@@ -0,0 +1,13 @@
+import { canonicalizeBoolean } from '../utils';
+
+export default {
```
tmeyer2115, line 3: If we ever add TypeScript to the Theme, it'd be great to have a `RuntimeConfigListener` type. I could see Consulting or a more advanced client wanting to right their own listeners.

static/js/runtime-config-listener-registrant.js
```diff
@@ -0,0 +1,16 @@
+import analyticsListener from './runtime-config-listeners/analytics';
+import sessionTrackingListener from './runtime-config-listeners/session-tracking';
+
+/**
+ * Responsible for registering listeners to a runtime config object
+ */
+export default class RuntimeConfigListenerRegistrant {
```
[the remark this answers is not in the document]
cea2aj: You're right, the complexity is not justified here. I think I got a little carried away with the SRP and trying to keep HitchhikerJS minimal

static/js/runtime-config-listeners/analytics.js
cea2aj: ya, that would be great

static/js/utils.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Returns true if the value is either the boolean value
+ * 'true' or a string representation of 'true'
+ * 
+ * @param {string|boolean} value
+ * @returns {boolean}
+ */
+ export function canonicalizeBoolean (value) {
```
oshi97, line 8: nit but extra space at the start of the line

tests/static/js/utils.js
```diff
@@ -0,0 +1,28 @@
+import { canonicalizeBoolean } from '../../../static/js/utils';
+ 
+ describe('canonicalizeBoolean works properly', () => {
```
oshi97, line 3: extra space

static/js/runtime-config-listeners/analytics.js
```diff
@@ -0,0 +1,13 @@
...
+  key: 'analyticsEventsEnabled',
+  callback: value => {
+    window.AnswersExperience.AnswersInitializedPromise
+      .then(() => {
+        const canonicalizedValue = canonicalizeBoolean(value);
+        ANSWERS.setAnalyticsOptIn(canonicalizedValue);
+      })
+      .catch(err => console.warn(err));
```
oshi97, line 11: if an error is caught, is the console.warn duplicated between the two runtime config listeners, since they both have this catch?

static/js/utils.js
```diff
@@ -0,0 +1,16 @@
+/**
```
oshi97, line 1: would it make sense to do this kind of data cleaning before the listener is called? not sure how possible this would be, but it would be nice if the listeners didn't have to worry about this (and perform the same canonicalization multiple times)

static/js/runtime-config-listeners/analytics.js
```diff
@@ -0,0 +1,13 @@
+import { canonicalizeBoolean } from '../utils';
+
+export default {
+  key: 'analyticsEventsEnabled',
+  callback: value => {
+    window.AnswersExperience.AnswersInitializedPromise
```
oshi97, line 6: is there some way we could get rid of having to specify waiting for ANSWERS.init to finish? I was thinking we could add a registerListener method to AnswersExperience that would automatically wait for AnswersInitializedPromise before executing callbacks, but I'm there's probably a myriad of ways to do it
We could even add a .catch() for the AnswersInitializedPromise there, before .then()-ing the callbacks?

static/js/HitchhikerJS.js
```diff
@@ -35,8 +35,15 @@ import RuntimeConfigReceiver from './runtime-config-receiver';
 export { RuntimeConfigReceiver };
 
 import RuntimeConfig from './runtime-config';
-import AnswersExperience from './answers-experience';
 const runtimeConfig = new RuntimeConfig();
+
+import AnswersExperience from './answers-experience';
 window.AnswersExperience = new AnswersExperience(runtimeConfig);
 
+import analyticsListener from './runtime-config-listeners/analytics';
```
oshi97, line 43: would it make more sense to do the registerListener calls inside AnswersExperience?

oshi97: pipe dream but maybe there's a world where we manage to move all the cards into the bundle, and then move the ANSWERS.init into AnswersExperience?

static/js/runtime-config-listeners/analytics.js
cea2aj: ya, good catch (pun intended)

cea2aj: I think it makes sense to remove the catches from here then

static/js/utils.js
cea2aj: ya, I can experiment with this. We could have an args param in the listeners which indicates the data types of the callback args so that we can canonicalize accordingly

static/js/runtime-config-listeners/analytics.js
cea2aj: I'm going to try setting up some processing that can wait for the AnswersInitializedPromise and that canonicalizes the params

static/js/HitchhikerJS.js
cea2aj: ya the, listener registration does make sense to do in the AnswersExperience

cea2aj: Moving our JS out of html.hbs and core.hbs into HitchhikerJS would be great

cea2aj: updated to move registerListener calls to AnswersExperience

static/js/runtime-config-listeners/analytics.js
cea2aj: I moved canonicalization to the runtime config itself

static/js/utils.js
cea2aj: I moved canonicalization to the runtime config itself

static/js/runtime-config-listeners/analytics.js
cea2aj: The waiting for the Answers intit is now setup in the AnswersExperience

static/js/runtime-config-listeners/analytics.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * @type {import('../runtime-config.js').RuntimeConfigListener}
+ */
+export default {
+  key: 'analyticsEventsEnabled',
+  valueType: 'boolean',
```
oshi97, line 6: love the valueType! way better than what I was thinking

static/js/utils.js
```diff
@@ -0,0 +1,16 @@
...
+ * Returns true if the value is either the boolean value
+ * 'true' or a string representation of 'true'
+ * 
+ * @param {string|boolean} value
+ * @returns {boolean}
+ */
+export function canonicalizeBoolean (value) {
+  if (typeof(value) === 'string') {
```
oshi97, line 9: super nit but I normally see typeof without the parenthesis (I didn't even know you could do it with parens lol). Do we have a preference between the two?

static/js/answers-experience.js
```diff
@@ -1,35 +1,51 @@
...
+ */
 
 export default class AnswersExperience {
   constructor (runtimeConfig) {
     this.runtimeConfig = runtimeConfig;
     this.AnswersInitializedPromise = new DeferredPromise();
+    this._runtimeConfigListeners = [
+      analyticsListener,
+      sessionTrackingListener
+    ]
```
oshi97, line 16: I would say "nit semicolon" but if I'm being honest I don't like semicolons lol

static/js/runtime-config.js
```diff
@@ -1,3 +1,16 @@
+import { canonicalizeBoolean } from './utils';
+
+/**
+ * @typedef {Object} RuntimeConfigListener
```
oshi97, line 4: 😮  typescript prep

oshi97 (approved):
curious what other people think but looks great to me!

static/js/utils.js
cea2aj: I think i prefer it without

yen-tt (approved):
Nice! I like the valueType field addition and the mapping step that automatically add the wait for Answers init instead of specifying that manually in each listener object. Makes it simpler to add more runtimeConfig listeners later. lgtm!

PR #871, opened 2021-06-30
import runtime hbs helpers passed to the sdk from the hbshelpers folder

oshi97:
This commit lets us share hbshelpers between jambo and the SDK, instead
of copying them (which is what we were doing before). This is done using
a webpack resolve alias.
Helpers that have handlebars in them cannot be easily shared, though
there is only one case of that.

The helpers were moved above the partial-block, to avoid issues with adding
components before registering helpers. We don't have any components in the theme
 that have this problem, but I have run into this before when messing around with things.

J=SLAP-1297
TEST=manual

smoke test that the page still loads
cards would break instantly if the any and all helpers weren't registered
to the SDK

test custom icons in card components, on both es and en locales (to test relative path
handling)

### DIFF
--- a/cards/event-standard/component.js
+++ b/cards/event-standard/component.js
@@ -33,7 +33,8 @@ class event_standardCardComponent extends BaseCard['event-standard'] {
       CTA1: {
         label: 'RSVP', // The CTA's label
         iconName: 'calendar', // The icon to use for the CTA
-        url: profile.ticketUrl || profile.website, // The URL a user will be directed to when clicking
+        iconUrl: 'static/assets/slapshot.png',
+        url: '123', // The URL a user will be directed to when clicking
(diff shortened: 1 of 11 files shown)


oshi97:
testing prod mode

oshi97:
Webpack doesn't like it when you mix ES6 and commonjs, and unfortunately jambo only understands hbs helpers that are commonjs. So putting it into the bundle is probably out of the window for now.
I tried thinking about how to do the inlining method as well with just HBS, but obviously module.exports is not something hbs understands. We could do jank where we create an empty object called "module", then loop over the desired helpers, continually reassigning the value of module.exports and calling ANSWERS.registerHelper on it. 
We probably need to allow jambo to handle hbshelpers using ES6 syntax

PR #873, opened 2021-07-01
full-page-map: fix alternative verticals iconUrl not using relativePath

oshi97:
Noticed this while testing replacing all the IconComponent usages.

Relative path was not being appended to the iconUrl correctly
because the template was referencing just `relativePath`
and not `@root.relativePath` or `../../relativePath` 
(i.e. it was using the wrong handlebars context).

This wouldn't have caused visual issues for most users,
since with the relativePath not being set correctly the iconUrl
would get set to `/<iconUrl>`, instead of something like `../<iconUrl>`
which for most cases will point to the same location.

This only matters if somebody is taking the files output by the
theme's build, and moving them into a directory structure.
In that case things may break, since the two iconUrls above may
no longer point to the same location.

I changed the other two relativePath usages for consistency

J=SLAP-1297
TEST=manual

test that the iconUrl for alternativeverticals on a full page map
now uses the correct relative url, instead of the absolute url
that would result from relativePath being null

added a percy snapshot for alternative verticals on the full page map,
though this doesn't really test what was being fixed

### DIFF
--- a/theme-components/vertical-full-page-map/script.js
+++ b/theme-components/vertical-full-page-map/script.js
@@ -32,9 +32,9 @@ ANSWERS.addComponent('VerticalFullPageMapOrchestrator', Object.assign({},
         {{#with (lookup verticalsToConfig verticalKey)}}
         {{#if isFirst}}isFirst: {{isFirst}},{{/if}}
         {{#if icon}}icon: "{{{icon}}}",{{/if}}
-        {{#if iconUrl}}iconUrl: "{{#unless (isNonRelativeUrl iconUrl)}}{{relativePath}}/{{/unless}}{{{iconUrl}}}",{{/if}}
+        {{#if iconUrl}}iconUrl: "{{{relativePathHandler url=iconUrl relativePath=root.relativePath}}}",{{/if}}
         label: {{> verticalLabel overridedLabel=label verticalKey=../verticalKey fallback=@key}},
-        url: "{{#if url}}{{{url}}}{{else if ../url}}{{../../relativePath}}/{{{../url}}}{{else}}{{{@key}}}.html{{/if}}",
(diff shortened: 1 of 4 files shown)


theme-components/vertical-full-page-map/script.js
```diff
@@ -32,9 +32,9 @@ ANSWERS.addComponent('VerticalFullPageMapOrchestrator', Object.assign({},
         {{#with (lookup verticalsToConfig verticalKey)}}
         {{#if isFirst}}isFirst: {{isFirst}},{{/if}}
         {{#if icon}}icon: "{{{icon}}}",{{/if}}
-        {{#if iconUrl}}iconUrl: "{{#unless (isNonRelativeUrl iconUrl)}}{{relativePath}}/{{/unless}}{{{iconUrl}}}",{{/if}}
+        {{#if iconUrl}}iconUrl: "{{{relativePathHandler url=iconUrl relativePath=@root.relativePath}}}",{{/if}}
```
cea2aj, line 35: this is unrelated to this, but we can probably switch over all of the `isNonRelativeUrl` calls to use the `relativePathHandler`, right?

oshi97: yeah, this one is a little weird because the url is @key concatenated with ".html" but we have a concat helper that can do that.

oshi97: Hmm actually, we already updated these in other components that have verticalPages config. We don't want to use a concat helper here, unlike what I originally thought, because by default we want es/locations.html to have links to "people.html" not "../people.html" (if that makes sense)

tmeyer2115 (approved):

yen-tt (commented):
Just for my own understanding, this is because we use #each loop in handlebars and doing '../' (not the actual file path stuff) let us access the data from parent context instead of the item that is being loop right?

PR #875, opened 2021-07-02
Loading indicator in theme

yen-tt:
- add opacity setting in theme css
- add default loadingIndicator to true in searchbar

Note: loading indicator changes require corresponding loading indicator udpates from SDK

J=SLAP-1335
TEST=manual
- Loading indicator is present when set to true during loading
- During loading, universal result section have opacity set to 0.5 and 1. Confirm visual update on page

### DIFF
--- a/templates/common-partials/script/searchbar.hbs
+++ b/templates/common-partials/script/searchbar.hbs
@@ -26,6 +26,9 @@ ANSWERS.addComponent("SearchBar", Object.assign({}, {
   {{#if verticalKey}}
     verticalKey: "{{{verticalKey}}}",
   {{/if}}
+  loadingIndicator: {
+    display: true
+  },
   ...(shouldAddOverlayConfig ? overlayConfig : {}),
   onCreate: function() {
     this._container.textContent = '';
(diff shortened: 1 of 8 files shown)


templates/common-partials/script/searchbar.hbs
```diff
@@ -26,6 +26,9 @@ ANSWERS.addComponent("SearchBar", Object.assign({}, {
   {{#if verticalKey}}
     verticalKey: "{{{verticalKey}}}",
   {{/if}}
+  loadingIndicator: {
+    display: true
```
tmeyer2115, line 30: The HHs may want the ability to toggle this on and off. So, instead of hard-coding it here, I think we should add the `true` default in the `ComponentSettings.SearchBar` of each template's `page-config`. That way, the HHs can see how to enable/disable the indicator if they want.

yen-tt: oh yeah. I thought I would have it here and HHs would manually add this customization on whatever template they want. But explicitly have it on the template makes more sense. Updated!

tmeyer2115 (approved):

oshi97 (approved):

PR #877, opened 2021-07-07
point the test-site to develop

oshi97:
This way we automatically get any new features we're working on, and
need to use for percy snapshots/acceptance tests in the theme.

J=SLAP-1423
TEST=manual

see that the test-site still loads

### DIFF
--- a/test-site/config/global_config.json
+++ b/test-site/config/global_config.json
@@ -1,5 +1,5 @@
 {
-  "sdkVersion": "feature/develop-i18n", // The version of the Answers SDK to use
+  "sdkVersion": "develop", // The version of the Answers SDK to use
   "apiKey": "2d8c550071a64ea23e263118a2b0680b", // The answers api key found on the experiences page. This will be provided automatically by the Yext CI system
   // "experienceVersion": "<REPLACE ME>", // the Answers Experience version to use for API requests. This will be provided automatically by the Yext CI system
   // "businessId": "<REPLACE ME>", // The business ID of the account. This will be provided automatically by the Yext CI system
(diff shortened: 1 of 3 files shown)


test-site/config/global_config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "sdkVersion": "feature/develop-i18n", // The version of the Answers SDK to use
+  "sdkVersion": "develop", // The version of the Answers SDK to use
```
tmeyer2115, line 2: I'm a bit torn about this. On the one hand, it'd be nice to constantly be testing the latest SDK in the Theme. On the other, it feels odd to have the Theme's `global_config` point to one SDK version and this to another. IMO, we should have the test-site always on the same SDK version as in the `global_config`.

oshi97: I feel like it's okay since this global_config is for the test site, which is what we use to develop on

oshi97: We'll also get to run canary/latest through the theme's test battery which could help catch breaking changes/bugs in the SDK. On the other hand it's possible some change in the SDK covers up a bug in the theme... Do you have any suggestions for how we can handle cases when we're developing some feature hand in hand with the SDK + theme? I know in the past we sometimes would have feature branches, but I would like to stay away from that if possible due to the issues that come with rebasing.

tmeyer2115: It feels unsafe to have the `test-site` pointed one way and the Theme's `global_config` another. I think we need to always ensure that the SDK version used in developing the Theme is the same one used here in the test site. Maybe we need to write some code that augments this `global_config` with the SDK version from the one in the Theme. Allowing the two to drift could cause some confusion.

If the `test-site` inherits the SDK version from the Theme, developing the Theme and SDK hand in hand would work as it does today. On the Theme's feature branch, we point to the corresponding SDK feature branch.

oshi97: yeah you're right, my brain assumed that whenever the SDK is released we will also release a new theme version, but this isn't necessarily true. will keep this manual for now

cea2aj: If we make the CAC Repo Generator (SLAP-1446), the test site would use the same SDK version as the theme's top level global config by default. That might be cleaner than writing some custom code that augments the global config of the test site

PR #878, opened 2021-07-12
TestCafe wait for results

yen-tt:
Update cafeTest acceptance test

- Added `waitOnSearchComplete` in verticalResults.js to wait till query responses return status 200 and searchComplete state appears on page.
- Register requestLogger to each test to track vertical request and response to Answers API endpoint
- Added IE11nocachehook from SDK
- From oliver's acceptance test failure investigation..
&ensp; - first point: the test first function call is `verticalResults.scrollToBottom()`, which wait on `this._resultsWrapper` when perform scroll on page. This resultsWrapper will always exist because it's set in our page template. The test didn't account for waiting on the actual search result div. Calling `waitOnSearchComplete` before scrolling would prevent this issue.
&ensp; - second point: removed the arbitrary wait time 1s in `dragLeft()`. Replaced with waiting for the response to be returned first using searchComplete state.
- Updated tests to wait for results before calling `getNumResults()` which check if results exists after a search. (From testCafe documentation, Selector wait/timeouts functionality have no effect on Selector.exists and Selector.count properties. These properties are evaluated immediately.)

J=SLAP-1425
TEST= auto

ran testCafe locally to verify tests and wait time in selectors


tests/acceptance/blocks/verticalresults.js
```diff
@@ -66,6 +69,19 @@ class VerticalResults {
     return this._selector.exists;
   }
 
+  /**
+   * wait for results to return based on search complete state
```
tmeyer2115, line 73: Proper capitalization

tests/acceptance/blocks/verticalresults.js
```diff
@@ -66,6 +69,19 @@ class VerticalResults {
     return this._selector.exists;
   }
 
+  /**
+   * wait for results to return based on search complete state 
+   * (default selector timeout of 10 seconds)
+   */
+   async waitOnSearchComplete () {
+    if(!this._preSearch.exists) {
```
tmeyer2115, line 77: For my own edification, what is `preSearch`?

tests/acceptance/blocks/verticalresults.js
```diff
@@ -66,6 +69,19 @@ class VerticalResults {
...
+      await this._searchLoading();
+      await t.expect(this._searchLoading.exists).ok();
+    }
+    await this._searchComplete();
+    await t.expect(this._searchComplete.exists).ok();
```
tmeyer2115, line 82: Does `await this._searchComplete()` being successful imply `t.expect(this._searchComplete.exists).ok()` automatically?

tests/acceptance/blocks/verticalresults.js
```diff
@@ -66,6 +69,19 @@ class VerticalResults {
     return this._selector.exists;
   }
 
+  /**
+   * wait for results to return based on search complete state 
+   * (default selector timeout of 10 seconds)
+   */
+   async waitOnSearchComplete () {
```
tmeyer2115, line 76: If someone calls this method, but no search has actually been executed, will it timeout? I'm wondering if we should handle that situation gracefully, like it returns immediately. Maybe not something we need to worry about, just a thought.

tests/acceptance/blocks/verticalresults.js
yen-tt: test would still consider passed if `await this._searchComplete()` doesn't find any matching element

tests/acceptance/blocks/verticalresults.js
yen-tt: it's one of the search states from SDK (presearch, loading, completed). I originally thought initial search on page may go from presearch to completed, skipping loading state. Turns out it's just testCafe unable to catch the loading transition. Updated to track query request/responses instead!

tests/acceptance/blocks/verticalresults.js
yen-tt: in this method, and in the updated one, it will timeout eventually. I think if no search is executed and someone uses this method, meaning they are expecting a search or result returned of some sort, it make sense to indicate a failure here to notify that the expected behavior is not met

tests/acceptance/blocks/verticalresults.js
```diff
@@ -66,6 +68,42 @@ class VerticalResults {
...
+  /**
+   * Wait for results to load on page by checking query response status and searchComplete state
+   * (timeout is set to 10 seconds)
+   */
+  async waitOnSearchComplete() {
+    const responseWaitTimeout = 10000;
+    const waitTimeInterval = 200;
+    let totalWaitTime = 0;
+    while (totalWaitTime < responseWaitTimeout 
+      && this._queryRequestLogger.requests.find(r => !r.response || r.response.statusCode !== 200)) {
```
cea2aj, line 80: Is this code waiting while an invalid response exists? is there a reason for that approach rather than waiting until a successful response has been logged?

yen-tt: Logger will hold all request/response throughout a test so just checking for a successful response might accidentally check on previous requests. But, after some thoughts, I updated this method check for a successful response in the loop, and clear out the logger at the end of this method. This make sure we are always verifying the latest response.

tests/acceptance/suites/vertical-full-page-map-with-filters.js
```diff
@@ -3,12 +3,18 @@ import SearchBar from '../blocks/searchbar';
 import VerticalResults from '../blocks/verticalresults';
 import ThemeMap from '../blocks/thememap';
 import CollapsibleFilters from '../blocks/collapsiblefilters';
+import { registerIE11NoCacheHook } from '../../test-utils/testcafe';
 
 fixture`Vertical Full Page Map with Filters and Clusters`
   .page(`http://localhost:${PORT}/locations_full_page_map_with_filters`)
+  .beforeEach(async t => {
+    await VerticalResults.registerLogger(t);
+    await registerIE11NoCacheHook(t);
```
oshi97, line 12: should the ie11 hook go inside registerLogger? like, is there ever a time we would call registerLogger without also registering the ie11 hook?

yen-tt: yeah, that would make sense. Updated!

oshi97 (approved):

tests/acceptance/blocks/verticalresults.js
```diff
@@ -66,6 +69,44 @@ class VerticalResults {
     return this._selector.exists;
   }
 
+  /**
+   * Wait for results to load on page by checking query response status and searchComplete state
+   * (timeout is set to 10 seconds)
+   */
+  async waitOnSearchComplete() {
```
oshi97, line 76: Sorry I know I just recently approved this, but after looking at the SDK half of this item maybe this waitOnSearchComplete should go into its own request logger class instead of inside the VerticalResults component's class?

cea2aj: I agree, if we can move this to it's own class we could easily expand it to universal search.  Yen, if we have a new class such as `SearchRequestLogger`, we could have a `registerVerticalSearchLogger` and a `registerUniversalSearchLogger` function so that we could set it up for either type of search

oshi97 (approved):

PR #880, opened 2021-07-13
Send QueryId to the Parent for iframe integrations

cea2aj:
Add a iframe-messaging script partial which can be used to send messages to the parent site in iframe integrations

The iframe-common.js script must be overridden in order to listen to these queryId messages.

J=SLAP-1439
TEST=manual

Override the iframe-common script and print all messages received by the parent frame. Observe a message being received on every search which includes the queryId.

### DIFF
--- a/script/iframe-messaging.js
+++ b/script/iframe-messaging.js
@@ -1,0 +1,15 @@
+/**
+ * Add code here to send messages to the parent site for iframe integrations.
+ * To listen to these messages, static/js/iframe-common.js must be overriden
+ * and the iFrameResize onMessage function must be modified.
+ */
+
+ANSWERS.core.storage.registerListener({
+  eventType: 'update',
+  storageKey: 'query-id',
+  callback: id => {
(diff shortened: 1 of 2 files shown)


script/iframe-messaging.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Add code here to send messages to the parent site for iframe integrations.
+ * To listen to these messages, static/js/iframe-common.js must be overriden
+ * and the iFrameResize onMessage function must be modified.
+ */
+
+ANSWERS.core.storage.registerListener({
```
tmeyer2115, line 7: Outside the scope of this PR, but I might make an item to have a top-level `registerListener` on `ANSWERS`. That way, we're not adding references to implementation details like `ANSWERS.core.storage`. The cat's already out of the bag, the Theme is using that other places. But, if we have a top-level `ANSWERS` method, might save us some trouble if we ever change internal storage details.

tmeyer2115 (approved):

PR #882, opened 2021-07-15
Post SessionID to Parent

yen-tt:
send the sessionId in the same message that the queryId is being sent

J=SLAP-1450
TEST=manual

Launched iframe test site, performed a search and verified through console log in iframe-common.js onMessage() that a message is delivered to parent page with queryId and sessionId.

### DIFF
--- a/script/iframe-messaging.js
+++ b/script/iframe-messaging.js
@@ -9,7 +9,8 @@ ANSWERS.core.storage.registerListener({
   storageKey: 'query-id',
   callback: id => {
     window.parentIFrame.sendMessage(JSON.stringify({
-      queryId: id
+      queryId: id,
+      sessionId: ANSWERS.core.getOrSetupSessionId()
     }));
   }
 });


oshi97 (approved):

oshi97:
I know the item says this should go in develop, but I thought this would only be needed in the case deflection branch?

yen-tt:
> I know the item says this should go in develop, but I thought this would only be needed in the case deflection branch?

@cea2aj commits in case deflection feature branch is also in develop so I assume both need to be updated?

cea2aj:
> > I know the item says this should go in develop, but I thought this would only be needed in the case deflection branch?
> 
> @cea2aj commits in case deflection feature branch is also in develop so I assume both need to be updated?

yes, we are putting case deflection work in a feature branch and in develop

cea2aj (approved):

oshi97:
> > > I know the item says this should go in develop, but I thought this would only be needed in the case deflection branch?
> > 
> > 
> > @cea2aj commits in case deflection feature branch is also in develop so I assume both need to be updated?
> 
> yes, we are putting case deflection work in a feature branch and in develop

okay cool! for some reason I was imagining we would cherry-pick this into the feature branch, but it probably makes more sense to just to rebase the feature branch onto develop

oshi97:
> > > > I know the item says this should go in develop, but I thought this would only be needed in the case deflection branch?
> > > 
> > > 
> > > @cea2aj commits in case deflection feature branch is also in develop so I assume both need to be updated?
> > 
> > 
> > yes, we are putting case deflection work in a feature branch and in develop
> 
> okay cool! for some reason I was imagining we would cherry-pick this into the feature branch, but it probably makes more sense to just to rebase the feature branch onto develop

more response pyramids

PR #884, opened 2021-07-15
fix waitOnSearchComplete loop issue

yen-tt:
- update hook registration to use `requestHook` on fixture instead of `beforeEach` for each test
- update waitOnSearchComplete to use current testInstance from test suite instead of the imported testInstance reference from testcafe
- update tests on searchThisArea functionality and drag on search

### DIFF
--- a/tests/acceptance/suites/vertical-full-page-map.js
+++ b/tests/acceptance/suites/vertical-full-page-map.js
@@ -31,24 +31,22 @@ test('Clicking on a pin focuses on a results card', async t => {
 test('Search when map moves works', async t => {
   await SearchBar.submitQuery('virginia');
   await SearchRequestLogger.waitOnSearchComplete(t);
-  const resultsCountBeforeDrag = await VerticalResults.getNumResults();
   await ThemeMap.dragLeft();
   await ThemeMap.dragLeft();
-  await SearchRequestLogger.waitOnSearchComplete(t);
-  const resultsCountAfterDrag = await VerticalResults.getNumResults();
-  await t.expect(resultsCountBeforeDrag !== resultsCountAfterDrag).ok();
+  const isSearchFired = await SearchRequestLogger.waitOnSearchComplete(t);
+  await t.expect(isSearchFired).ok();
 });
 
 test('Search this area button works', async t => {
   await SearchBar.submitQuery('virginia');
   await ThemeMap.toggleSearchThisArea();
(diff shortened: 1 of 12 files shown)


oshi97:
is there something we can do about the search when map moves test? it's still marked as unstable

tests/acceptance/suites/vertical-full-page-map.js
```diff
@@ -30,48 +33,44 @@ test('Clicking on a pin focuses on a results card', async t => {
...
 test('Search when map moves works', async t => {
   await SearchBar.submitQuery('virginia');
-  await SearchRequestLogger.waitOnSearchComplete();
-  const resultsCountBeforeDrag = await VerticalResults.getNumResults();
+  await SearchRequestLogger.waitOnSearchComplete(t);
   await ThemeMap.dragLeft();
   await ThemeMap.dragLeft();
-  await SearchRequestLogger.waitOnSearchComplete();
-  const resultsCountAfterDrag = await VerticalResults.getNumResults();
-  await t.expect(resultsCountBeforeDrag !== resultsCountAfterDrag).ok();
```
oshi97, line 39: do we no longer want this expect?

tests/acceptance/suites/vertical-full-page-map.js
```diff
@@ -4,23 +4,26 @@ import VerticalResults from '../blocks/verticalresults';
 import ThemeMap from '../blocks/thememap';
 import Pagination from '../blocks/pagination';
 import SearchRequestLogger from '../searchrequestlogger';
+import { VERTICAL_SEARCH_URL_REGEX } from '../constants';
+import { registerIE11NoCacheHook } from '../../test-utils/testcafe';
 
 fixture`Vertical Full Page Map`
   .page(`http://localhost:${PORT}/locations_full_page_map`)
+  .requestHooks(SearchRequestLogger.createVerticalSearchLogger())
```
oshi97, line 12: like this!

tests/acceptance/suites/vertical-full-page-map.js
yen-tt: the goal of this test is to see that dragging the map would trigger a search. Previously, we saved the results before and after to compare and check for a search. Now, we can just use the requestLogger to confirm that a search is fired!

oshi97 (approved):

PR #885, opened 2021-07-16
remove html-asset-loader, preserve static/assets folder instead

oshi97:
This commit removes the html-asset-loader, which was used to automagically
perform webpack loads for static assets. As we have continued to add more complexity
to the theme, the scope of the html-asset-loader has grown to a difficult to maintain state.
The loader has to search through not only HTML, but also all javascript strings we provide.

Furthermore, the implementation of this loader is dependent on implementation specific
details of the html-loader.

Finally, removing this loader will make it much easier to migrate to ESBuild, or any other
bundler we choose.

I moved html-loader to the prod build only, and turned on all the default minification settings
instead of our custom ones. I also updated the version to the latest.
Also update browserlist (`npx browserslist@latest --update-db`) because it was asking to be updated.

J=SLAP-1419
TEST=manual

see that the test-site still displays static image assets correctly


static/webpack-config.js
```diff
@@ -60,7 +60,16 @@ module.exports = function () {
     new RemovePlugin({
       after: {
         root: `${jamboConfig.dirs.output}`,
-        include: ['static'],
+        test: [
+          {
+            folder: 'static',
+            method: absoluteFilePath => {
+              const filePathRelativeToOutput = path.relative(jamboConfig.dirs.output, absoluteFilePath);
+              return !filePathRelativeToOutput.startsWith('static/assets');
```
cea2aj, line 68: Can you expand on the reasoning behind this for my own edification?

oshi97: instead of erasing the entire static folder, it's only erasing files that don't start with "static/assets"
the path.relative is because the plugin sends the absolute file path to this hook, e.g
`/Users/oshi/answers-hitchhiker-theme/test-site/public/static/assets/slapshot.png`

I tried just adding exclude: ['static/assets'] but for whatever reason that doesn't work with `include`, they even say in their example for exclude that you have to use `test` instead of `include` if you want to use `exclude`. From my testing it looks like exclude only works with individual files and not folders though, so I had to resort to this weird checking.

https://www.npmjs.com/package/remove-files-webpack-plugin

cea2aj: Ah, got it. It just took me a second to follow the logic here. Would this explanatory value be accurate/helpful?

```suggestion
              const isFileWithinStaticAssetOutputDir = filePathRelativeToOutput.startsWith('static/assets')
              return !isFileWithinStaticAssetOutputDir;
```

cea2aj (approved):

PR #886, opened 2021-07-16
Add a manually written pot file for the open status formatter

oshi97:
We want to be able to get translations from smartling instead of
asking a hitchhiker/native speaker at yext. Eventually we will
probably want some way of injecting translations into the formatter.

Also fix rename open-status-18n to open-status-i18n (year old typo).

J=SLAP-1455
TEST=none


translations/open-status.pot
```diff
@@ -0,0 +1,23 @@
...
+
+msgctxt "Whether a store is open or closed"
+msgid "Open 24 Hours"
+msgstr ""
+
+msgctxt "Whether a store is open or closed"
+msgid "Closed"
+msgstr ""
+
+msgctxt "Whether a store is open or closed"
```
cea2aj, line 13: nit: This context might be a little confusing for this particular translation. How about something like: "Indicates the time that a store will open"

translations/open-status.pot
```diff
@@ -0,0 +1,23 @@
+msgid ""
+msgstr ""
+"Content-Type: text/plain; charset=UTF-8\n"
+
+msgctxt "Whether a store is open or closed"
```
cea2aj, line 5: what do you think about something like: "The hours that a store is open"

oshi97: updated!

translations/open-status.pot
oshi97: yeah makes sense

cea2aj (approved):

PR #887, opened 2021-07-16
Dev/assign

cjiang2000:
Created the assign handlebars helper to fill out some partials using Object.assign() at build time
JIRA=SLAP-1306
TEST=manual

checked output in console


tests/hbshelpers/assign.js
```diff
@@ -0,0 +1,21 @@
+const hbs = require('handlebars');
+import assign from '../../hbshelpers/assign';
+
+hbs.registerHelper('assign', assign);
+
+  it('performs an assign', () => {
```
oshi97, line 6: indentation is weird

tests/templates/script/commonPartials.js
```diff
@@ -0,0 +1,31 @@
...
+        });
+    });
+    */
+    
+
+    it('checks all common partials', () => {
+        files.forEach(function (file) {
+            const filename = file.concat('.js')
+            var tempPath = path.join(__dirname, "fixtures", filename);
+            expect(fs.readFileSync(tempPath, 'utf-8')).toEqual(hbs.compile(fs.readFileSync(path.join(partialPath, file), 'utf-8'))());
```
oshi97, line 28: eh I'd probably break up the readFileSyncs into individual variables

oshi97: instead of passing no data into the templates
`hbs.compile(fs.readFileSync(path.join(partialPath, file), 'utf-8'))()`

We probably want to pass some data into the templates, otherwise we're not really testing any cases here.

```js
hbs.compile(fs.readFileSync(path.join(partialPath, file), 'utf-8'))({
  some: 'data',
  appliedFilters: {
    componentSettings: { /* blah */ }
  }
})
```

tests/templates/script/commonPartials.js
```diff
@@ -0,0 +1,31 @@
+const path = require('path');
+const fs = require('fs');
+const hbs = require('../../test-utils/hbs');
+const directoryPath = 'templates/common-partials/script'
+const partialPath = path.resolve(__dirname, `../../../${directoryPath}`);
+const files = fs.readdirSync(partialPath);
+
+describe('common partials convert from handlebars correctly', () => {
+   /*
```
oshi97, line 9: we probably don't need to keep this around

hbshelpers/assign.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function assign(...options) {
```
oshi97, line 1: when I talked with Tom about this 2 weeks ago he didn't really like the name "assign". Maybe something like "shallowMergeConfig"? 3 words is a little extra, but I think we want some way to signify that this only does a shallow merge as opposed to a deep merge, and that this is only for static config.

templates/common-partials/script/appliedfilters.hbs
```diff
@@ -1,7 +1,13 @@
...
+  {{#assign}}[
+  {
+    "container": "#js-answersAppliedFilters"
+    {{#if verticalKey}}
+    ,
+    "verticalKey": "{{{verticalKey}}}"
+    {{/if}}
+  },
+  {{{json componentSettings.AppliedFilters}}}
+]
```
oshi97, line 11: would it be weird to smoosh lines 11-13 together? and also lines 1-2?

cjiang2000: theres some weird behavior so I'm gonna leave it unchanged for now, but if it matters I can definitely make the change.

oshi97: what is the weird behavior?

tests/templates/script/commonPartials.js
```diff
@@ -0,0 +1,62 @@
+const path = require('path');
+const fs = require('fs');
+const hbs = require('../../test-utils/hbs');
+const directoryPath = 'templates/common-partials/script'
+const partialPath = path.resolve(__dirname, `../../../${directoryPath}`);
+const files = fs.readdirSync(partialPath);
+
+/*
+//Generates common-partials default values
```
oshi97, line 9: this is pretty much the same thing as below but replacing the expect with a writeFile, so we don't need to keep this around.

tests/templates/script/commonPartials.js
```diff
@@ -0,0 +1,62 @@
...
+                "Navigation": {"a":"testN"},
+                "QASubmission": {"a":"testQAS"},
+                "SearchBar": {"a":"testSB"},
+                "SortOptions": {"a":"testSO"},
+                "SpellCheck": {"a":"testSC"},
+                "VerticalResults": {"a":"testVR"},
+                "VerticalResultsCount": {"a":"testVRC"}
+            }
+        }
+        ));
```
oshi97, line 60: smoosh lines 59 and 60

tests/templates/script/commonPartials.js
```diff
@@ -0,0 +1,62 @@
...
+*/
+
+
+it('checks all common partials', () => {
+    files.forEach(function (file) {
+        const filename = file.concat('.js')
+        var tempPath = path.join(__dirname, "fixtures", filename);
+        const fixture = fs.readFileSync(tempPath, 'utf-8')
+        const partial = fs.readFileSync(path.join(partialPath, file), 'utf-8')
+        expect(fixture).toEqual(hbs.compile(partial)( {
```
oshi97, line 44: extra space before the curly brace

templates/common-partials/script/appliedfilters.hbs
cjiang2000: extra new lines, but it doesnt make a difference if i change all of them, bc then its consistent

hbshelpers/shallowMergeConfig.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function shallowMergeConfig(...options) {
+  var parsedData = JSON.parse(options[options.length - 1].fn(this));
```
oshi97, line 2: use const where possible, then if you need to use var use let instead

tests/templates/script/commonPartials.js
```diff
@@ -0,0 +1,62 @@
...
+        }
+    });
+});
+*/
+
+
+it('checks all common partials', () => {
+    files.forEach(function (file) {
+        const filename = file.concat('.js')
+        var tempPath = path.join(__dirname, "fixtures", filename);
```
oshi97, line 41: const instead of var

hbshelpers/shallowMergeConfig.js
```diff
@@ -0,0 +1,11 @@
...
+  var stringFormat;
+  if (parsedData.length) {
```
oshi97, line 4: maybe we can use Array.isArray instead? think it's just a little clearer

tests/templates/script/commonPartials.js
```diff
@@ -0,0 +1,62 @@
...
+        if (err) {
+            console.error(err)
+            return
+        }
+    });
+});
+*/
+
+
+it('checks all common partials', () => {
```
oshi97, line 38: If we do the for loop outside of the `it`, we can get N tests with 1 expect each, instead of 1 test with N expects. The issue with 1 test with N expects is that it will short circuit on the first expect that breaks. With N tests, it's also clearer exactly which partial broke. You can use a template string for the name of the test to insert the name of the partial being tested https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

tests/templates/script/fixtures/pagination.hbs.js
```diff
@@ -0,0 +1,13 @@
...
+    if (window.parentIFrame) {
+      const paginateMessage = { action: 'paginate' };
+      window.parentIFrame.sendMessage(JSON.stringify(paginateMessage));
+    } else {
+      document.documentElement.scrollTop = 0;
+      document.body.scrollTop = 0; // Safari
+    }
+  },
+    verticalKey: "testKey",
+}, {}));
```
oshi97, line 13: There's no data here but that's probably fine. We only really need data for the partials using the new helper

hbshelpers/shallowMergeConfig.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function shallowMergeConfig(...options) {
```
cea2aj, line 1: can you add a jsdoc here explaining a little about what this does?

hbshelpers/shallowMergeConfig.js
```diff
@@ -1,3 +1,9 @@
+/**
+ * This helper merges JSON objects into a single config.
+ * 
+ * @param {spread syntax} Either an array of JSON objects, or a single JSON object.
```
cea2aj, line 4: rather than "spread syntax", can we document this as `...Object`?

cea2aj (approved):

PR #888, opened 2021-07-19
add templateValidator to the Theme

yen-tt:
- Add logger using npmlog package for color in logs
- Add templatedatavalidator hook to check for missing experienceKey, and invalid cardTypes or universalSectionTemplates

J=SLAP-1424
TEST=compile

- Changed a vertical config to use invalid cardType and universalSectionTemplates. Verified errors show up in console during jambo build (using local jambo that point to develop branch)
- Changed a universal config and verified errors show up in console

### DIFF
--- a/commands/helpers/utils/logger.js
+++ b/commands/helpers/utils/logger.js
@@ -1,0 +1,41 @@
+const log = require('npmlog');
+
+log.heading = 'jambo';
+log.headingStyle = { fg: 'magenta' };
+
+// Don't display a log prefix.
+const PREFIX = '';
+
+/**
+ * Sets the global logging level.
+ */
+exports.setLogLevel = function(level) {
+  log.level = level;
+}
+
+/**
+ * Logs an error.
+ *
+ * @param  {...any} args 
+ */
+exports.error = function(...args) {
+  log.error(PREFIX, ...args)
+}
+
+/**
+ * Logs a warning.
+ *
+ * @param  {...any} args 
+ */
+exports.warn = function(...args) {
+  log.warn(PREFIX, ...args)
+}
+
+/**
+ * Logs info.
+ *
+ * @param  {...any} args 
+ */
+exports.info = function(...args) {
+  log.info(PREFIX, ...args)
+}
--- a/hooks/templatedatavalidator.js
+++ b/hooks/templatedatavalidator.js
@@ -51,8 +51,11 @@ function isPageVerticalConfigValid(pageData, jamboConfig) {
     }
     const universalSectionTemplate = pageData.verticalsToConfig[key].universalSectionTemplate;
     const cardType = pageData.verticalsToConfig[key].cardType;
-    if (!isUniversalSectionTemplateValid(key, themeDirectory, universalSectionTemplate)
-      | !isCardTypeValid(key, themeDirectory, cardType)) {
+    const validatorResults = [
+      isUniversalSectionTemplateValid(key, themeDirectory, universalSectionTemplate), 
+      isCardTypeValid(key, themeDirectory, cardType)
+    ];
+    if(validatorResults.some((result) => !result)) {
       isValid = false;
     }
   });
--- a/hooks/templatedatavalidator.js
+++ b/hooks/templatedatavalidator.js
@@ -10,7 +10,12 @@ const { error } = require('../commands/helpers/utils/logger');
  */
 module.exports = function (pageData) {
   const jamboConfig = parseJamboConfig();
-  return isGlobalConfigValid(pageData.global_config) & isPageVerticalConfigValid(pageData, jamboConfig);
+  const validatorResults = [
+    isGlobalConfigValid(pageData.global_config), 
+    isPageVerticalConfigValid(pageData, jamboConfig)
+  ];
+  const isValid = !validatorResults.some((result) => !result);
+  return isValid;
 }
 
 /**
--- a/hooks/templatedatavalidator.js
+++ b/hooks/templatedatavalidator.js
@@ -5,23 +5,23 @@ const { error } = require('../commands/helpers/utils/logger');
 /**
  * Validates page template's data configuration
  * 
- * @param {Object} pageData configuration of a page template
+ * @param {Object} pageData data that gets passed into a page template
  * @returns {boolean} false if validator should throw an error
  */
 module.exports = function (pageData) {
   const jamboConfig = parseJamboConfig();
-  return isGlobalConfigValid(pageData.global_config) && isPageVerticalConfigValid(pageData, jamboConfig);
+  return isGlobalConfigValid(pageData.global_config) & isPageVerticalConfigValid(pageData, jamboConfig);
 }
 
 /**
  * Validates global config for the page template
  * 
- * @param {Object} global_config 
+ * @param {Object} globalConfig 
  * @returns {boolean}
  */
-function isGlobalConfigValid(global_config) {
-  if (!global_config.experienceKey) {
-    error('Missing Info: no experienceKey found in config file.');
+function isGlobalConfigValid(globalConfig) {
+  if (!globalConfig.experienceKey) {
+    error('Missing Info: no experienceKey found.');
     return false;
   }
   return true;
@@ -36,10 +36,9 @@ function isGlobalConfigValid(global_config) {
  */
 function isPageVerticalConfigValid(pageData, jamboConfig) {
   let isValid = true;
-  const themeDirectory = path.resolve(process.cwd(), jamboConfig.dirs.themes, jamboConfig.defaultTheme);
+  const themeDirectory = path.resolve(jamboConfig.dirs.themes, jamboConfig.defaultTheme);
   Object.keys(pageData.verticalsToConfig).forEach(key => {
     if (key === 'Universal') {
-      console.log(pageData);
       if (!isAllVerticalConfigsValid(pageData.verticalConfigs, jamboConfig)) {
         isValid = false;
       }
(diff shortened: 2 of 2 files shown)


hooks/templatedatavalidator.js
```diff
@@ -0,0 +1,115 @@
...
+ * @param {string} themeDir
+ * @param {string} cardType
+ * @returns {boolean}
+ */
+function isCardTypeValid(verticalName, themeDir, cardType) {
+  if (cardType) {
+    const cardTypePath = path.resolve(themeDir, 'cards/', cardType);
+    const customCardTypePath = path.resolve('cards/', cardType);
+    if (!fs.existsSync(cardTypePath) && !fs.existsSync(customCardTypePath)) {
+      error(`Invalid cardType: can't find "${cardType}" at at the expected paths "${cardTypePath}" or "${customCardTypePath}" for vertical "${verticalName}".`);
```
oshi97, line 90: just curious but what does cardTypePath print out in sitesgitstorm console? maybe we would want just the relative path from the repo root with path.join instead of the absolute path

hooks/templatedatavalidator.js
```diff
@@ -0,0 +1,115 @@
...
+ * @returns {boolean}
+ */
+function isAllVerticalConfigsValid(verticalConfigs, jamboConfig) {
+  if (!verticalConfigs) {
+    return true;
+  }
+  let isValid = true;
+  Object.keys(verticalConfigs).forEach(key => {
+    if (!isPageVerticalConfigValid(verticalConfigs[key], jamboConfig)) {
+      isValid = false;
```
oshi97, line 111: I like that this doesn't short circuit!

hooks/templatedatavalidator.js
```diff
@@ -0,0 +1,115 @@
...
+const { error } = require('../commands/helpers/utils/logger');
+/**
+ * Validates page template's data configuration
+ * 
+ * @param {Object} pageData configuration of a page template
+ * @returns {boolean} false if validator should throw an error
+ */
+module.exports = function (pageData) {
+  const jamboConfig = parseJamboConfig();
+  return isGlobalConfigValid(pageData.global_config) && isPageVerticalConfigValid(pageData, jamboConfig);
```
oshi97, line 13: maybe we can also have the non short circuiting logic here too, so we call `isPageVerticalConfigValid` even if the global config is not valid?

hooks/templatedatavalidator.js
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Validates global config for the page template
+ * 
+ * @param {Object} global_config 
+ * @returns {boolean}
+ */
+function isGlobalConfigValid(global_config) {
+  if (!global_config.experienceKey) {
+    error('Missing Info: no experienceKey found in config file.');
```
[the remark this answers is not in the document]
yen-tt: From the test site, the experience key is set in locale_config.json but later gets reformatted into pageData.global_config. Would it be confusing to say that it's for global config? If not, I will update it!

commands/helpers/utils/logger.js
```diff
@@ -0,0 +1,41 @@
+const log = require('npmlog');
+
+log.heading = 'jambo';
```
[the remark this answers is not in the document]
yen-tt: I think it's ok to keep it as 'jambo' heading since it's technically a validation running from jambo? Changing the heading and style with the current setup from logger.js would override the functions (export with different log config) so it wouldn't maintain the jambo npm color and heading. Unless I use the npmlog functions directly `log.error` instead of the exported functions.

hooks/templatedatavalidator.js
```diff
@@ -0,0 +1,115 @@
+var fs = require('fs');
+var path = require('path');
+const { parseJamboConfig } = require('../commands/helpers/utils/jamboconfigutils');
+const { error } = require('../commands/helpers/utils/logger');
+/**
+ * Validates page template's data configuration
+ * 
+ * @param {Object} pageData configuration of a page template
```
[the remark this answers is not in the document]
yen-tt: yup. hmm ok, will update

hooks/templatedatavalidator.js
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Valivates vertical config for current page template
+ * 
+ * @param {Object} pageData 
+ * @param {Object} jamboConfig 
+ * @returns {boolean}
+ */
+function isPageVerticalConfigValid(pageData, jamboConfig) {
+  let isValid = true;
+  const themeDirectory = path.resolve(process.cwd(), jamboConfig.dirs.themes, jamboConfig.defaultTheme);
```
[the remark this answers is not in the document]
yen-tt: ah, didn't realize the current working directory is used by default if resolving all {from} paths doesn't get an absolute path. Will remove it!

hooks/templatedatavalidator.js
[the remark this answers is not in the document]
oshi97: ah no you're right! maybe just "no experienceKey found" then?

hooks/templatedatavalidator.js
```diff
@@ -0,0 +1,114 @@
...
+const { error } = require('../commands/helpers/utils/logger');
+/**
+ * Validates page template's data configuration
+ * 
+ * @param {Object} pageData data that gets passed into a page template
+ * @returns {boolean} false if validator should throw an error
+ */
+module.exports = function (pageData) {
+  const jamboConfig = parseJamboConfig();
+  return isGlobalConfigValid(pageData.global_config) & isPageVerticalConfigValid(pageData, jamboConfig);
```
oshi97, line 13: bitwise and is cool here! I feel like it might get us in trouble if we have future validators that return truthy/falsy values instead of actual booleans. Could we find a different way to write this?

hooks/templatedatavalidator.js
yen-tt: it prints out `/repo-vol/themes/answers-hitchhiker-theme/cards/link-standard`

oshi97: ok I'm fine either way! I think just having it be `themes/answers-hitchhiker-theme/cards/link-standard` is better most of the time, but very occasionally it might be useful to have the full path

hooks/templatedatavalidator.js
yen-tt: updated!

hooks/templatedatavalidator.js
yen-tt: hmm `path.join(hemeDir, 'cards/', cardType);` would still give me full path though

oshi97: oh I guess you would have to change themeDir to also use path.join()

hooks/templatedatavalidator.js
```diff
@@ -0,0 +1,122 @@
...
+ * Valivates vertical config for current page template
+ * 
+ * @param {Object} pageData 
+ * @param {Object} jamboConfig 
+ * @returns {boolean}
+ */
+function isPageVerticalConfigValid(pageData, jamboConfig) {
+  let isValid = true;
+  const themeDirectory = path.resolve(jamboConfig.dirs.themes, jamboConfig.defaultTheme);
+  Object.keys(pageData.verticalsToConfig).forEach(key => {
```
oshi97, line 45: sorry i didn't notice this earlier, but I think it would be a little simpler if we used Array.prototype.every instead of an isValid param and forEach loop (also in isAllVerticalConfigsValid as well)

hooks/templatedatavalidator.js
yen-tt: ah duh, yup updated!

hooks/templatedatavalidator.js
yen-tt: I think Array.prototype.every stops evaluating the array (short circuits) the first time it finds an element that does not satisfy the given requirement. I think we should keep the forEach loop. I should make the changes for where I use Array.prototype.some as well since it also short circuits.

yen-tt: Updated to use Reduce() instead, I think it's a bit cleaner now. What do you think?

oshi97: ahh good catch!

oshi97 (approved):
lgtm!

hooks/templatedatavalidator.js
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * Valivates vertical config for current page template
+ * 
+ * @param {Object} pageData 
+ * @param {Object} jamboConfig 
+ * @returns {boolean}
+ */
+function isPageVerticalConfigValid(pageData, jamboConfig) {
+  const themeDirectory = path.join(jamboConfig.dirs.themes, jamboConfig.defaultTheme);
+  return Object.keys(pageData.verticalsToConfig).reduce((isValid, key) => {
```
oshi97, line 44: Ah shoot ok the isValid is actually kind of tricky too. (Sorry I know I'm really harping on this PR lol...) Maybe we could use `.map` and then call `.every` instead of just `.reduce` ?

hooks/templatedatavalidator.js
```diff
@@ -0,0 +1,111 @@
...
+ * @param {Object} pageData data that gets passed into a page template
+ * @returns {boolean} false if validator should throw an error
+ */
+module.exports = function (pageData) {
+  const jamboConfig = parseJamboConfig();
+  const validatorResults = [
+    isGlobalConfigValid(pageData.global_config), 
+    isPageVerticalConfigValid(pageData, jamboConfig)
+  ];
+  const isValid = !validatorResults.some((result) => !result);
```
oshi97, line 17: nit but could we flip this to use every to avoid the double negation?

hooks/templatedatavalidator.js
yen-tt: no problem. This is a good idea!

oshi97 (approved):
lgtm!

PR #889, opened 2021-07-19
use webpack asset modules instead of file-loader, fix preload

oshi97:
This commit replaces file-loader with webpack asset modules,
which is the webpack 5 way to do it. This is slightly more streamlined
than using loaders that are separate npm packages, which have been
deprecated by asset modules.

This commit also fixes font preloads no longer pointing correctly to font
assets. Originally, I tried to separate that out to another PR, however
it seems like html-loader@2 only works properly with asset modules and
not file-loader. As opposed to downgrading html-loader, then reupgrading
it in this PR, they're in the same PR.

I removed the minimize=true config for html-loader, because by default
html-loader will minimize when mode="production" and won't minimize when
mode="development"

J=SLAP-1084,SLAP-1419
TEST=manual

see that html is minified in prod mode, but not for dev mode

see that the font preloads get their URLs updated correctly,
for example pointing to source-sans.[hash].woff instead of
static/assets/fonts/source-sans.woff

see that bundle.css points to the correct URL, and that the
font shows up properly on the page


static/webpack-config.js
```diff
@@ -127,11 +127,42 @@ module.exports = function () {
...
+                        });
+                        const isFont = attributes.find(attr => {
+                          return attr.name === 'as' && attr.value === 'font';
+                        });
+                        return isPreload && isFont;
+                      }
+                    }
+                  ],
+                  urlFilter: (attribute, value) => {
+                    return /^[./]*static\/assets\//.test(value);
```
cea2aj, line 154: Do we need to escape that forward slash inside the brackets? Regex101 is giving me an error there

static/webpack-config.js
```diff
@@ -127,11 +127,42 @@ module.exports = function () {
...
+                      attribute: 'href',
+                      type: 'src',
+                      filter: (tag, attribute, attributes) => {
+                        const isPreload = attributes.find(attr => {
+                          return attr.name === 'rel' && attr.value === 'preload';
+                        });
+                        const isFont = attributes.find(attr => {
+                          return attr.name === 'as' && attr.value === 'font';
+                        });
+                        return isPreload && isFont;
```
cea2aj, line 149: Does this filter mean that only preload fonts are updated to use the hash in their name? Do we not need this for other assets/

static/webpack-config.js
oshi97: really? it works fine for me on regex101, maybe you weren't setting it to javascript?

static/webpack-config.js
oshi97: Fonts are the only assets that still run through webpack, because we import them in our css using resolve-url-loader. We could update the css to not use resolve-url-loader, though I'm a little worried about that breaking people's custom fonts

static/webpack-config.js
cea2aj: oh, that was it

cea2aj (approved):

PR #891, opened 2021-07-20
fix get language locale from map provider

yen-tt:
- `getLanguageForProvider` was using an undefined variable supportedLocalesforProvider. Updated function to 
get provider instance based on given provider name, and check if the given locale is within the list of provider's supported locales.

J=TECHOPS-1288
TEST=manual

used client experience that uses locale en_GB and see that map is loaded on page
added jest test for getLanguageForProvider()

### DIFF
--- a/static/js/theme-map/Util/helpers.js
+++ b/static/js/theme-map/Util/helpers.js
@@ -40,7 +40,6 @@ const getMapProvider = (mapProvider) => {
   if (mapProvider === 'mapbox') {
     return MapboxMaps;
   }
-  console.warn(`Map provider ${mapProvider} is not supported in the theme. Default to MapBox.`);
   return MapboxMaps;
 }
 
--- a/static/js/theme-map/Util/helpers.js
+++ b/static/js/theme-map/Util/helpers.js
@@ -1,3 +1,6 @@
+import { GoogleMaps } from '../Maps/Providers/Google.js';
+import { MapboxMaps } from '../Maps/Providers/Mapbox.js';
+
 /**
  * Gets the language locale according to specific fallback logic
  * 1. The user-specified locale to the component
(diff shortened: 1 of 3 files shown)


cea2aj:
Looks good. I think we should probably add a little unit test coverage as well

tests/static/js/theme-map/Util/helpers.js
```diff
@@ -52,3 +52,25 @@ describe('getNormalizedLongitude', () => {
     });
   });
 });
+
+describe('getLanguageForProvider', () => {
+  describe('it works with different language/locale pairs', () => {
+    it('works with 0', () => {
+      expect(getNormalizedLongitude(0)).toEqual(0);
```
oshi97, line 59: extra test

yen-tt: ahhh, my bad.

static/js/theme-map/Util/helpers.js
```diff
@@ -1,25 +1,29 @@
...
+const getLanguageForProvider = (localeStr, mapProvider) => {
   if (localeStr.length == 2) {
     return localeStr;
   } 
 
   if (localeStr.length > 2) {
-    if (supportedLocalesForProvider.includes(localeStr)) {
-      return localeStr;
-    } 
+    const provider = (mapProvider === 'google') ? GoogleMaps : MapboxMaps;
```
oshi97, line 20: should we display an error if mapprovider is not google or mapbox? maybe we want to extract this method to a separate helper since it's also used in ThemeMap.js?

yen-tt: yeah, that would help if something is not configure right. I will update this

yen-tt: do you think it would be better to default to google/mapbox, or just log error and return null?

oshi97: maybe log an error and return 'en'? I think returning null would make it blow up later down the road. If we want it to blow up I think we should just throw an error, though I imagine logging an error is good enough to not hide bugs but still handle things gracefully?

oshi97: ooo sorry you mean for the helper, yeah I think that's fine!

oshi97: we probably don't want to default, I think the provider already defaults to mapbox somewhere else in the code?

yen-tt: I think you are referring to the loadAndInitializeMap() in ThemeMap for defaulting to mapbox? which we would replace with our helper call. So I guess to keep the functionality the same, I would default to Mapbox.

oshi97: ah yeah that's my bad

oshi97 (approved):

PR #897, opened 2021-07-27
Update translation verification

cea2aj:
Update translation verification so that it also checks that the messages.pot file in the repo is up to date

Also update translation test to run on every branch

J=none
TEST=manual

Ran the new extract translations script and see the correct behavior. Update the circle ci script and check the exit codes to ensure that code 0 is returned when successful, and 1 is returned when the messages.pot is out of date or when translations are missing

### DIFF
--- a/.github/run_translation_verification.sh
+++ b/.github/run_translation_verification.sh
@@ -1,4 +1,3 @@
-
 #!/bin/bash
 
 set -e
--- a/.github/run_translation_verification.sh
+++ b/.github/run_translation_verification.sh
@@ -4,9 +4,8 @@
 set -e
 
 # Verifiy that the messages.pot file in the repo is up to date
-# If there are any git diffs after translations are extracted, the checked-in file is out of date
+# If there is a git diff after translations are extracted, the checked-in file is out of date
 npm run extract-translations
-
 git diff --exit-code translations/messages.pot > /dev/null # send stdout to /dev/null to reduce clutter in the CI output
 diff_exit_code=$?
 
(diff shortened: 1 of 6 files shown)


.github/run_translation_verification.sh
```diff
@@ -0,0 +1,24 @@
+
+#!/bin/bash
+
+set -e
+
+# Verifiy that the messages.pot file in the repo is up to date
+# If there are any git diffs after translations are extracted, the checked-in file is out of date
+npm run extract-translations
+
+git diff --exit-code translations/messages.pot > /dev/null # send stdout to /dev/null to reduce clutter in the CI output
```
oshi97, line 10: I think we can use something like the $GITHUB_REF env var to always run the messages.pot verification, but only run the [lang].po check for master and release branches

oshi97: Ah actually GITHUB_BASE_REF might be better

cea2aj: In the run_browserstack_acceptance.sh script we use $GITHUB_BRANCH. Does that work?

oshi97: I thiink that might be the branch being merged into the base branch? Not sure though

cea2aj: updated!

oshi97: oh wait I did not realize we were already using it for filtering lol, think I was looking at the version on master

.github/run_translation_verification.sh
```diff
@@ -0,0 +1,32 @@
+
```
oshi97, line 1: besides this extra line pr looks great!

oshi97 (approved):

PR #874, opened 2021-07-01
replace IconComponent with icon partials

oshi97:
This PR updates all IconComponent usages to use the icon partials in the SDK
It also points the test-site to the feature/develop-i18n branch of the SDK,
which is the same commit that develop is currently at, but includes i18n assets
so our spanish snapshots work

The first group of commits do a find and replace for identical (including whitespace like indentation)
usages of IconComponent.

J=SLAP-1297
TEST=manual

test that custom icons can still be set for card CTAs, section titles, and the searchbar
see that built in icons still show up

PR #899, opened 2021-07-27
Add HitchhikerCTA-icon classes back

cea2aj:
Add back the HitchhikerCTA divs and classes that were removed in #874

J=SLAP-1473
TEST=visual

Inspect the DOM and see the div and class present for the standard card. Inspect the updated Percy snapshots and see some of the icons change size slightly due to the CSS targeting the div. Also see the icon appear for the standard card which was previously missing.


oshi97 (approved):
thank you for fixing these!

PR #900, opened 2021-07-28
i18n rtl css support

yen-tt:
use rtl css bundle if locale is a language written from right to left

- use `rtlcss-webpack-plugin` to generate a second CSS bundle that's RTL if there's a locale in `locale_config` file that requires such bundle.
- update `html.hbs`to update href to bundle.rtl.css and include dir='rtl' if it's a rtl locale
- change span styling in verticalResultsCount so its ordering is correct on both ltr and rtl direction

J=SLAP-1458
TEST=manual

launched test site in theme with en, es, and ar locale. See that *.rtl.css bundles are created in public folder. See that en and es pages function normally and ar pages uses rtl format.
launched test site in theme with en and es only. See that no *.rtl.css bundles are created. See that en and es pages function normally.

### DIFF
--- a/hbshelpers/cssrtl.js
+++ b/hbshelpers/cssrtl.js
@@ -1,0 +1,12 @@
+/**
+ * Returns true if the given language is written from right to left (requires rtl css)
+ * 
+ * @param {string} locale 
+ * @returns {boolean}
+ */
+module.exports = function cssrtl(locale) {
+  if(locale === 'ar') {
+    return true;
+  }
+  return false;
+}
--- a/static/js/rtl.js
+++ b/static/js/rtl.js
@@ -1,0 +1,12 @@
+/**
+ * Returns true if the given language is written from right to left (requires rtl css)
+ * 
+ * @param {string} locale 
+ * @returns {boolean}
+ */
+module.exports = function isRTL (locale) {
+  if(locale === 'ar') {
+    return true;
+  }
+  return false;
+}
--- a/static/webpack-config.js
+++ b/static/webpack-config.js
@@ -7,6 +7,7 @@ const RemovePlugin = require('remove-files-webpack-plugin');
 const { merge } = require('webpack-merge');
 const { parse } = require('comment-json');
 const RtlCssPlugin = require('rtlcss-webpack-plugin');
+const isRTL = require('./js/rtl');
 
 module.exports = function () {
   const jamboConfig = require('./jambo.json');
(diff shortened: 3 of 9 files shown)


hbshelpers/cssrtl.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Returns true if the given language is written from right to left (requires rtl css)
+ * 
+ * @param {string} locale 
+ * @returns {boolean}
+ */
+module.exports = function cssrtl(locale) {
```
cea2aj, line 7: nit: what do you think of the name: isRTL?

cea2aj: Are we able to put this function in static and refer to it from this helper? Ideally we have a list of RTL languages in one place so that we don't have to update them here and also in static/webpack-config.js

static/webpack-config.js
```diff
@@ -26,6 +27,19 @@ module.exports = function () {
     });
   }
 
+  const cssrtlPlugins = [];
```
cea2aj, line 30: nit: I believe this pattern with the array that gets pushed to was created because there can be multiple htmlPlugins. I don't think it's necessary for cssRTL, however it's pretty simple so I'm okay with keeping it as-is if you'd like. Another option would be to make the `hasRtlLocale` boolean available in this scope, and then after the plugins array is defined we can add the plugin if `hasRtrlLocale` is true

hbshelpers/cssrtl.js
yen-tt: that make sense. Updated!

static/webpack-config.js
yen-tt: I see what you mean. I think we should keep this pattern so it's easier to see everything that is included in `plugins` instead of adding on to it in another chunk of code later in the file. I renamed it to `cssRtlPlugin` to indicate there's only one if that's clearer.

static/js/rtl.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Returns true if the given language is written from right to left (requires rtl css)
+ * 
+ * @param {string} locale 
+ * @returns {boolean}
+ */
+module.exports = function isRTL (locale) {
```
oshi97, line 7: maybe this can go into static/webpack since it's only used as part of the build? right now static/js is for stuff that goes on the browser

static/js/rtl.js
```diff
@@ -0,0 +1,12 @@
...
+  if(locale === 'ar') {
```
oshi97, line 8: space after if

static/webpack-config.js
```diff
@@ -6,6 +6,8 @@ const HtmlPlugin = require('html-webpack-plugin');
 const RemovePlugin = require('remove-files-webpack-plugin');
 const { merge } = require('webpack-merge');
 const { parse } = require('comment-json');
+const RtlCssPlugin = require('rtlcss-webpack-plugin');
```
oshi97, line 9: just for my edification what kinds of transformations does this do?
does this just auto fix a decent chunk of the sprint items we have? 
Maybe a dumb question but I assuuume this is meant to work well with dir="rtl"? Like we won't like double "flip" certain styling properties

static/webpack-config.js
```diff
@@ -33,6 +35,19 @@ module.exports = function () {
     globalConfig = parse(globalConfigRaw);
   }
 
+  const cssRtlPlugin = [];
```
oshi97, line 38: I agree with Connor using a list for a single item is strange, can't we just have it be
```js
let cssRtlPlugin;
//...
[
  cssRtlPlugin,
  ...htmlPlugins
]
```

static/webpack-config.js
yen-tt: it's use, in addition to mini-css-extract-plugin we already have, to create a second rtl css bundle. It uses the rtlcss library under the hood that tom suggested on slack. I'm not sure if it fix most of the sprint items, this might have to apply to SDK css as well. The map error and the chevron change still needs to be fix.

yen-tt: From what I understand, dir="rtl" affects the direction property in CSS for text content flow within a block-level element. But it doesn't change the other styling like margin left/right and whatnots, which is what this plugin would take care of. so I don't think we will double flip things

static/webpack-config.js
yen-tt: in the case that we don't need a cssRtlPlugin, it would have an undefined element in there and would throw an error when running webpack. I could move this chunk of code after `plugins` is first set and add to `plugins` directly if cssRtlPlugin is required, but then it might not be organize(?) as seeing everything that's suppose to be in plugins to be already there when it's first set.

oshi97: okay that's fair!

oshi97: Would renaming `htmlPlugins` to just `plugins` simplify things a little? then we could just push all plugins onto this array. We already have an array called `plugins` anyways.

static/webpack-config.js
oshi97: awesome!

oshi97: I guess we can look into specifically what css changes it makes later

oshi97 (approved):

PR #901, opened 2021-07-29
Map issues fixed for RTL locale/language

yen-tt:
Fix Mapbox issues when using RTL styling and language

- Map not showing up in rtl format issue is from Mapbox only (Google works on all map pages) due to[ css issues](https://github.com/mapbox/mapbox-gl-js/issues/9329) from their api. This is fixed in [v1.9.0](https://github.com/mapbox/mapbox-gl-js/blob/main/CHANGELOG.md#190). This pr updated the version to v1.13.1 only since they stop supporting ie11 in higher version.
- Use [MapBoxLanguage](https://github.com/mapbox/mapbox-gl-language) plugin to update language on displayed on map. Previously, map will display in english only.
- MapBox by default doesn't support RTL languages. Updated to use [mapbox-rtl-text](https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-rtl-text/) plugin to resolve the issue.
- Update zoom control box position based on if it's ltr or rtl language
- Move rtl.js function to static/common folder since it's use in browser and in webpack

J=SLAP-1459
TEST=manual

Launched multilang test site
-  see that map appears in en, es, and ar for all map pages (locations, locations_google, locations_full_page_map, locations_full_page_map_with_filters) and providers (Google, Mapbox)
- see that the language displayed in the map matched with the given locale in url
- see that arabic is displayed property on map in rtl styling
- see that zoom control box is in top right for en and es, and is in top left for ar.


static/js/theme-map/Maps/Providers/Google.js
```diff
@@ -6,6 +6,7 @@ import { MapProviderOptions } from '../MapProvider.js';
 import { ProviderMap } from '../ProviderMap.js';
 import { ProviderPin } from '../ProviderPin.js';
 import { debounce } from '../../Util/helpers';
+import isRTL from '../../../../webpack/rtl';
```
oshi97, line 9: hmm ok this probably shouldn't go in webpack if we want to use it in static js. Maybe this can go in a static/common folder, or do you have any other ideas?

static/js/theme-map/Maps/Providers/Mapbox.js
```diff
@@ -175,22 +183,38 @@ const yextAPIKey = 'pk.eyJ1IjoieWV4dCIsImEiOiJqNzVybUhnIn0.hTOO5A1yqfpN42-_z_GuL
  * @see {MapProvider~loadFunction}
  */
 function load(resolve, reject, apiKey, {
-  version = 'v1.6.1'
+  version = 'v1.13.0'
```
oshi97, line 186: Is there a 1.13.1 we can use? I noticed one exists here https://github.com/mapbox/mapbox-gl-js/releases?after=v2.2.0 but not sure if that's the same thing

static/js/theme-map/Maps/Providers/Mapbox.js
```diff
@@ -175,22 +183,38 @@ const yextAPIKey = 'pk.eyJ1IjoieWV4dCIsImEiOiJqNzVybUhnIn0.hTOO5A1yqfpN42-_z_GuL
...
 }) {
   const baseUrl = `https://api.mapbox.com/mapbox-gl-js/${version}/mapbox-gl`;
 
   const mapStyle = document.createElement('link');
   mapStyle.rel = 'stylesheet';
   mapStyle.href = baseUrl + '.css';
 
+  const mapLanguageScript = document.createElement('script');
+  mapLanguageScript.src = 'https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-language/v0.10.1/mapbox-gl-language.js';
```
oshi97, line 195: would it be possible to import this as an npm module and import it with a dynamic import?

static/js/theme-map/Maps/ProviderMap.js
```diff
@@ -24,6 +24,15 @@ class ProviderMapOptions {
     this.zoomEndHandler = () => {};
     this.canvasClickHandler = () => {};
     this.providerOptions = {};
+    this.language = 'en';
+  }
+
+  /**
+   * @param {string} language The language displayed on map
+   */
+  withLanguage(language) {
```
oshi97, line 33: For some reason I thought the full page map already knew about your locale to some degree. Could we just use the locale to determine your language?

static/js/theme-map/Maps/Providers/Mapbox.js
oshi97: actually ignore this, I think if they're hosting it already there's not much point using a dynamic import

static/js/theme-map/Maps/ProviderMap.js
yen-tt: MapBoxMap class doesn't have access to locale or language from what I can see. I console log out its content and params given from constructor and couldn't find anything either. So I had to pass the language it. User wouldn't need to add anything in their configuration, `ThemeMap` would take the user provided locale and pass that accordingly to the actual Map provider instance.

static/js/theme-map/Maps/Providers/Mapbox.js
yen-tt: oh I was looking at here: https://github.com/mapbox/mapbox-gl-js/blob/main/CHANGELOG.md. looks different because the bug fix from your link is mentioned in v2.0.1 from my link instead

static/js/theme-map/Maps/Providers/Google.js
yen-tt: mm I think putting it in static/common is good. will update!

static/js/theme-map/Maps/Providers/Mapbox.js
oshi97: Ah okay would it be okay for you to try to see if 1.13.1 works? I imagine they just forgot to update their changelog since their npm also has a 1.13.1

yen-tt: yea, will do!

static/js/theme-map/Maps/Providers/Mapbox.js
```diff
@@ -26,10 +28,16 @@ class MapboxMap extends ProviderMap {
...
+    this.map.addControl(new MapboxLanguage({
+      defaultLanguage: options.language
+    }));
+
     // Add the zoom control
     if (options.controlEnabled) {
       const zoomControl = new mapboxgl.NavigationControl({showCompass: false})
-      this.map.addControl(zoomControl);
+      isRTL(options.language)
+      ? this.map.addControl(zoomControl, 'top-left')
```
cea2aj, line 39: nit: usually we put a tab before the "?" or the ":" of the ternary operator

static/js/theme-map/Maps/Map.js
```diff
@@ -53,6 +53,15 @@ class MapOptions {
...
     this.singlePinZoom = 14;
     this.wrapper = null;
+    this.language = 'en';
+  }
+
+  /**
+   * @param {string} language The language displayed on map
+   */
+  withLanguage(language) {
+    this.language = language;
```
cea2aj, line 63: I would just like to check is this language always a 2 character language like "es", or can it be a locale such as "es_US"? If it's just 2 chars then maybe we should take a substring? See my pr here: https://github.com/yext/answers-search-ui/pull/1070/files

yen-tt: this language would be from `getLanguageForProvider` function (from this [techop fix](https://github.com/yext/answers-hitchhiker-theme/pull/891/files)). It will not always be just 2 chars but it should be appropriate for the corresponding provider. In this case with Mapbox, it will always be 2 chars since we don't have any specific supported language locale specified for Mapbox

cea2aj: Okay, in that case what do you think of calling this `withLocale` instead since we generally use "language" to refer to just the two char language, and a "locale" to include additional info such as a dialect or region

yen-tt: sure thing. Updated!

cea2aj (approved):

PR #902, opened 2021-07-30
RTL Answers CSS and chevrons

cea2aj:
Update to use the answers.rtl.css if the locale is an rtl locale, and update the chevron_icons in the link components so that they will flip for RTL locales

J=SLAP-1465
TEST=manual

See that on an RTL locale, the answers.rtl.css bundle is used and the link title chevron icon flips. Test that the overlay chevron is flipped when using the answers.rtl.css bundle.

### DIFF
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html class="YxtPage-doc">
+<html class="YxtPage-doc" {{#if (isRTL global_config.locale) }}dir="rtl"{{/if}}>
   <head>
     <script>
       {{#babel}}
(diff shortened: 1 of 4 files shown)


layouts/html.hbs
```diff
@@ -143,7 +143,11 @@
     {{/babel}}
     </script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/iframe-resizer/4.2.10/iframeResizer.contentWindow.min.js"></script>
-    <link rel="stylesheet" type="text/css" href="{{sdkAssetUrl global_config.sdkVersion 'en' 'answers.css'}}">
+    {{#if (isRTL global_config.locale) }}
```
yen-tt, line 146: nit: The link tag below this only change what is needed in if/else isRTL check instead of the whole tag like here. Maybe we should update either one to follow the other so it's more consistent. I'm fine either way!

static/js/overlay/answers-frame/dom/promptinjector.js
```diff
@@ -117,14 +117,17 @@ export default class PromptInjector {
...
     iconEl.classList.add('OverlayPrompt-buttonIcon');
-    iconEl.innerHTML = `<?xml version="1.0" encoding="UTF-8"?>
-      <svg viewBox="0.5 0 6 9"
-        version="1.1" xmlns="http://www.w3.org/2000/svg"
-        xmlns:xlink="http://www.w3.org/1999/xlink">
-        <g fill-rule="evenodd" transform="translate(-1 -8)">
-          <path d="m2.6417004 8-1.1417004 1.0575 3.70850202 3.4425-3.70850202 3.4425 1.1417004 1.0575 4.8582996-4.5z"/>
-        </g>
-      </svg>`;
+    iconEl.innerHTML = `
```
yen-tt, line 120: Do we need another div inside the iconEl div or can we add the classes `Icon Icon--chevron` into iconEl too?

layouts/html.hbs
cea2aj: oh, your right. I'll update to match the convention you made below

static/js/overlay/answers-frame/dom/promptinjector.js
cea2aj: good catch, i'll try that!

yen-tt (approved):

PR #903, opened 2021-08-02
add the custom-init-config partial

oshi97:
I considered adding this in HitchhikerJS, which
would give us better error messages if the a Hitchhiker messes
up JS syntax. On the other hand Rose asked for a partial,
and I'm a little worried about confusing Hitchhikers.

In the future I was thinking jambo could pass a getter for
templates and partials into templatedatavalidator, or even
the whole HBS instance, so we can easily validate template/partials.

J=SLAP-1447
TEST=manual

add an additional-answers-config to the test site, with functions for onVerticalSearch
and onUniversalSearch


package.json
```diff
@@ -39,6 +39,7 @@
     "postcss": "^8.2.1",
     "serve": "^11.3.2",
     "simple-git": "^2.24.0",
+    "strip-comments": "^2.0.1",
```
tmeyer2115, line 42: Small nit: Shouldn't this be a normal dependency? It's used to power a HBS helper.

script/core.hbs
```diff
@@ -103,7 +103,8 @@
         window.iframeLoaded.then(() => {
           {{> script/iframe-messaging}}
         });
-      }
+      },
+      ...{{#stripJsComments}}{{> script/custom-init-config }}{{/stripJsComments}}
```
tmeyer2115, line 107: Should we wrap all of `onReady` in this helper?

script/custom-init-config.js
```diff
@@ -0,0 +1,5 @@
+// This file expects a JS object literal.
```
tmeyer2115, line 1: What about naming this `additional-answers-config` or something like that? It's a small nit, but there is already custom init config in some sense in the form of the `userConfig` that's passed in.

package.json
oshi97: this is only used for tests, the real dependency is in static/package.json

script/core.hbs
oshi97: I was thinking it wouldn't hurt to let them override onReady if they wanted, but at that point you should probably override core.hbs

package.json
oshi97: actually, we don't need this dep at all. javascript is pretty smart about how it parses out comments

tmeyer2115 (approved):

PR #906, opened 2021-08-04
Fix Grunt memory leak

cea2aj:
Fix the grunt memory leak by isolating the jambo and webpack build into their own process

J=SLAP-1484
TEST=manual

Run "grunt watch" and confirm the build runs when files change

### DIFF
--- a/static/Gruntfile.js
+++ b/static/Gruntfile.js
@@ -12,20 +12,20 @@ module.exports = function (grunt) {
     watch: {
       all: {
         files: ['**', '!**/node_modules/**', `!${outputDir}/**`],
-        tasks: ['jambobuild']
+        tasks: ['build-site']
       },
     },
   });
 
   grunt.loadNpmTasks('grunt-webpack');
   grunt.loadNpmTasks('grunt-contrib-watch');
-  grunt.registerTask('jambobuild', 'Jambo build.', jambobuild);
+  grunt.registerTask('build-site', 'Builds the site.', buildSite);
 }
 
 /**
- * Spawns a jambo and webpack build and prints sends all output to the console
+ * Builds the site by spawning a jambo and webpack build
  */
-function jambobuild () {
+function buildSite () {
   const spawnedProcess = spawnSync('npx jambo build && npx webpack --config webpack-config.js', {
     shell: true,
     stdio: 'inherit'
--- a/static/Gruntfile.js
+++ b/static/Gruntfile.js
@@ -1,5 +1,5 @@
 const webpackConfig = require('./webpack-config');
-const { spawnSync } = require("child_process");
+const { spawnSync } = require('child_process');
 const jamboConfig = require('./jambo.json');
 
 const outputDir = jamboConfig.dirs.output;
@@ -16,26 +16,12 @@ module.exports = function (grunt) {
       },
     },
   });
-
   grunt.loadNpmTasks('grunt-webpack');
   grunt.loadNpmTasks('grunt-contrib-watch');
-  grunt.registerTask('build-site', 'Builds the site.', buildSite);
-}
-
-/**
- * Builds the site by spawning a jambo and webpack build
- */
-function buildSite () {
-  const spawnedProcess = spawnSync('npx jambo build && npx webpack --config webpack-config.js', {
-    shell: true,
-    stdio: 'inherit'
+  grunt.registerTask('build-site', 'Builds the site.', () => {
+    spawnSync('npx jambo build && npx webpack --config webpack-config.js', {
+      shell: true,
+      stdio: 'inherit'
+    });
   });
-
-  if (spawnedProcess.error) {
-    console.error(spawnedProcess.error.message);
-  }
-
-  const { stderr, stdout } = spawnedProcess;
-  stderr && console.error(stderr);
-  stdout && console.log(stdout);
 }
--- a/static/Gruntfile.js
+++ b/static/Gruntfile.js
@@ -1,45 +1,36 @@
-const webpackConfig = require('./webpack-config');
-const { exec } = require("child_process");
+const { spawnSync } = require("child_process");
 const jamboConfig = require('./jambo.json');
 
 const outputDir = jamboConfig.dirs.output;
 
 module.exports = function (grunt) {
   grunt.initConfig({
-    webpack: {
-      myConfig: webpackConfig
-    },
     watch: {
       all: {
         files: ['**', '!**/node_modules/**', `!${outputDir}/**`],
-        tasks: ['jambobuild', 'webpack',],
-        options: {
-          spawn: false,
-        },
+        tasks: ['jambobuild']
       },
     },
   });
 
-  grunt.loadNpmTasks('grunt-webpack');
   grunt.loadNpmTasks('grunt-contrib-watch');
+  grunt.registerTask('jambobuild', 'Jambo build.', jambobuild);
+}
 
-  grunt.registerTask('jambobuild', 'Jambo build.',
-  function() {
-    // Force task into async mode and grab a handle to the "done" function.
-    var done = this.async();
-    // Run some sync stuff.
-    grunt.log.writeln('Processing task...');
-    // And some async stuff.
-    exec('npx jambo build', (error, stdout, stderr) => {
-      if (error) {
-        console.log(error.message);
-        done(false);
-        return;
-      }
-
-      stderr && console.error(stderr);
-      stdout && console.log(stdout);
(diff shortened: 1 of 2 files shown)


static/Gruntfile.js
```diff
@@ -12,34 +12,30 @@ module.exports = function (grunt) {
...
-      }
-
-      stderr && console.error(stderr);
-      stdout && console.log(stdout);
-      done();
-    });
+/**
+ * Spawns a jambo and webpack build and prints sends all output to the console
+ */
+function jambobuild () {
```
tmeyer2115, line 28: Since this does more than just the Jambo build, could we change the name? Maybe something like `buildSite` or something like that?

tmeyer2115 (commented):
Were you able to do a couple quick timings to ensure this isn't significantly slower than how we did things previously?

cea2aj:
> Were you able to do a couple quick timings to ensure this isn't significantly slower than how we did things previously?

Good call. With this change I'm seeing build times consistently near 7.5 seconds. With the previous change, the first build would be near that time, however on subsequent runs it would speed up to about 5 or 4 seconds

static/Gruntfile.js
cea2aj: sure

static/Gruntfile.js
```diff
@@ -1,5 +1,5 @@
 const webpackConfig = require('./webpack-config');
-const { exec } = require("child_process");
+const { spawnSync } = require("child_process");
```
oshi97, line 2: single quotes here! (I think there's some way to make the vscode autoimport use single quotes)

static/Gruntfile.js
```diff
@@ -12,34 +12,30 @@ module.exports = function (grunt) {
...
-        tasks: ['jambobuild', 'webpack',],
-        options: {
-          spawn: false,
-        },
+        tasks: ['build-site']
       },
     },
   });
 
   grunt.loadNpmTasks('grunt-webpack');
```
oshi97, line 20: we can remove grunt-webpack

static/Gruntfile.js
```diff
@@ -12,34 +12,30 @@ module.exports = function (grunt) {
...
-      stderr && console.error(stderr);
-      stdout && console.log(stdout);
-      done();
-    });
+/**
+ * Builds the site by spawning a jambo and webpack build
+ */
+function buildSite () {
+  const spawnedProcess = spawnSync('npx jambo build && npx webpack --config webpack-config.js', {
+    shell: true,
```
oshi97, line 30: is there a difference between using a shell and not?

static/Gruntfile.js
```diff
@@ -12,34 +12,30 @@ module.exports = function (grunt) {
...
+    stdio: 'inherit'
   });
-}
+
+  if (spawnedProcess.error) {
```
oshi97, line 34: if we're inheriting stdio we shouldn't need these logs anymore

static/Gruntfile.js
tmeyer2115: I think it's okay to rename the task. I seriously doubt anyone is invoking it themselves and as you said, the CI scripts don't reference it.

tmeyer2115:
> > Were you able to do a couple quick timings to ensure this isn't significantly slower than how we did things previously?
> 
> Good call. With this change I'm seeing build times consistently near 7.5 seconds. With the previous change, the first build would be near that time, however on subsequent runs it would speed up to about 5 or 4 seconds

Hmm, I wonder if there was some kind of caching that was happening when the Webpack build was part of the main Node app. Which site did you test on?

cea2aj:
> > > Were you able to do a couple quick timings to ensure this isn't significantly slower than how we did things previously?
> > 
> > 
> > Good call. With this change I'm seeing build times consistently near 7.5 seconds. With the previous change, the first build would be near that time, however on subsequent runs it would speed up to about 5 or 4 seconds
> 
> Hmm, I wonder if there was some kind of caching that was happening when the Webpack build was part of the main Node app. Which site did you test on?

I tested this on the test-site and I also made a new theme site

static/Gruntfile.js
cea2aj: Tom and I figured that we should keep grunt webpack because the ci scripts call `grunt webpack`

tmeyer2115:
> > > > Were you able to do a couple quick timings to ensure this isn't significantly slower than how we did things previously?
> > > 
> > > 
> > > Good call. With this change I'm seeing build times consistently near 7.5 seconds. With the previous change, the first build would be near that time, however on subsequent runs it would speed up to about 5 or 4 seconds
> > 
> > 
> > Hmm, I wonder if there was some kind of caching that was happening when the Webpack build was part of the main Node app. Which site did you test on?
> 
> I tested this on the test-site and I also made a new theme site

Could you try this out on YAnswers and see what the time difference is like there? I'm curious to get data points on a bigger site.

static/Gruntfile.js
cea2aj: If shell is false, the first parameter is the parameter is the command and the second is an array of arguments. Also, I don't think the `&&` would work. With shell=true, we can easily run both the jambo build and the webpack build since the command will run within `/bin/sh`

cea2aj:
> > > > > Were you able to do a couple quick timings to ensure this isn't significantly slower than how we did things previously?
> > > > 
> > > > 
> > > > Good call. With this change I'm seeing build times consistently near 7.5 seconds. With the previous change, the first build would be near that time, however on subsequent runs it would speed up to about 5 or 4 seconds
> > > 
> > > 
> > > Hmm, I wonder if there was some kind of caching that was happening when the Webpack build was part of the main Node app. Which site did you test on?
> > 
> > 
> > I tested this on the test-site and I also made a new theme site
> 
> Could you try this out on YAnswers and see what the time difference is like there? I'm curious to get data points on a bigger site.

Sure

static/Gruntfile.js
oshi97: ok sgtm for the hotfix! I think for 1.23 we probably don't need to set shell to true though based on recent PR to add spawnSync to the gruntfile https://github.com/yext/answers-hitchhiker-theme/pull/872

oshi97: I imagine there's some tiny overhead for using a shell

tmeyer2115 (approved):

oshi97 (approved):

oshi97:
> > > > > > Were you able to do a couple quick timings to ensure this isn't significantly slower than how we did things previously?
> > > > > 
> > > > > 
> > > > > Good call. With this change I'm seeing build times consistently near 7.5 seconds. With the previous change, the first build would be near that time, however on subsequent runs it would speed up to about 5 or 4 seconds
> > > > 
> > > > 
> > > > Hmm, I wonder if there was some kind of caching that was happening when the Webpack build was part of the main Node app. Which site did you test on?
> > > 
> > > 
> > > I tested this on the test-site and I also made a new theme site
> > 
> > 
> > Could you try this out on YAnswers and see what the time difference is like there? I'm curious to get data points on a bigger site.
> 
> Sure. Update: It's a smilar story on YAnswers. With this change I'm getting a consistent 6.5 to 7 seconds. Before it would start around 6.2 and get down to about 4.0 on subsequent runs

It looks like grunt-webpack has a CachePluginFactory that could be related to both build speed and memory usage
https://github.com/webpack-contrib/grunt-webpack/tree/main/src/plugins

cea2aj:
> > > > > > > Were you able to do a couple quick timings to ensure this isn't significantly slower than how we did things previously?
> > > > > > 
> > > > > > 
> > > > > > Good call. With this change I'm seeing build times consistently near 7.5 seconds. With the previous change, the first build would be near that time, however on subsequent runs it would speed up to about 5 or 4 seconds
> > > > > 
> > > > > 
> > > > > Hmm, I wonder if there was some kind of caching that was happening when the Webpack build was part of the main Node app. Which site did you test on?
> > > > 
> > > > 
> > > > I tested this on the test-site and I also made a new theme site
> > > 
> > > 
> > > Could you try this out on YAnswers and see what the time difference is like there? I'm curious to get data points on a bigger site.
> > 
> > 
> > Sure. Update: It's a smilar story on YAnswers. With this change I'm getting a consistent 6.5 to 7 seconds. Before it would start around 6.2 and get down to about 4.0 on subsequent runs
> 
> It looks like grunt-webpack has a CachePluginFactory that could be related to both build speed and memory usage
> https://github.com/webpack-contrib/grunt-webpack/tree/main/src/plugins

I agree, I think the issue was related to some sort of caching on the webpack side

PR #907, opened 2021-08-05
Fix Grunt memory leak

cea2aj:
Fix the grunt memory leak by isolating the jambo and webpack build into their own process

J=SLAP-1484
TEST=manual

Run "grunt watch" and confirm the build runs when files change

### DIFF
--- a/static/Gruntfile.js
+++ b/static/Gruntfile.js
@@ -12,18 +12,17 @@ module.exports = function (grunt) {
     watch: {
       all: {
         files: ['**', '!**/node_modules/**', `!${outputDir}/**`],
-        tasks: ['build',],
+        tasks: ['build-site'],
         options: {
           spawn: false,
         },
       },
     },
   });
-
   grunt.loadNpmTasks('grunt-webpack');
   grunt.loadNpmTasks('grunt-contrib-watch');
 
-  grunt.registerTask('build', 'jambo build -> grunt webpack', () => {
+  grunt.registerTask('build-site', 'Builds the site', () => {
     spawnSync('npm', ['run',  'build', '--silent'], { stdio: 'inherit' });
   });
 }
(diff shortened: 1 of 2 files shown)


static/Gruntfile.js
```diff
@@ -12,18 +12,17 @@ module.exports = function (grunt) {
...
-        tasks: ['build',],
+        tasks: ['build-site'],
         options: {
           spawn: false,
         },
       },
     },
   });
-
   grunt.loadNpmTasks('grunt-webpack');
```
oshi97, line 22: ok THIS time we can get rid of grunt-webpack right?

static/Gruntfile.js
```diff
@@ -12,18 +12,17 @@ module.exports = function (grunt) {
     watch: {
```
oshi97, line 12: Github won't let me put a comment on the line I want, but we don't need to import webpackConfig or define a webpack task

static/Gruntfile.js
cea2aj: no, we still need to be able to call `grunt webpack` since the CI scripts call it

oshi97: can't the ci scripts call the webpack cli?

oshi97: the ci scripts should probably just do an `npm run build`

cea2aj: If we remove it then any site would break if they upgraded to this version of the theme. We would need to create some patch files for the CI scripts. The CI scripts only call `grunt webpack` in one-time scenarios such as the build step, so the memory leak isn't is problem there

oshi97: I think I would like to not keep around old dependencies we don't need if possible, but we can kick the can down the road a little!

oshi97: We can definitely ask Hitchhikers to manually update their ci scripts at some point in time when they upgrade, but you're right probably not worth it here

oshi97 (approved):

static/Gruntfile.js
cea2aj: Maybe it's worth considering having the CI files part of theme 2.0?

PR #916, opened 2021-08-13
remove duplicated global config commented out props on upgrade

oshi97:
This comes from a QA item where, after upgrading the theme, the global
config had duplicated "commented out" props. For example a commented out apiKey
like `// apiKey: <REPLACE_ME>`

The way comment-json works is that each comment is associated with a
specific "prop" of the json (unless the json is empty or the comment
is outside of the main json body).

If we're not careful with how we add new config to the global_config,
the prop associated with the giant commented out config block will change.
Because we merge the new theme's global_config with the user's current
global_config on upgrade (with the user's having higher priority), this
results in said huge block being duplicated, because comment-json sees
it as a *different* comment than the preexisting giant one, since it is
now associated with a different prop.

Update comment-json to latest for small parsing improvements.
There are no breaking changes we're worried about.

J=SLAP-1514
TEST=manual, auto

unit tests
test upgrading yanswers, also with some minor tweaks to their config
to test things

### DIFF
--- a/postupgrade/utils.js
+++ b/postupgrade/utils.js
@@ -5,21 +5,83 @@ const simpleGit = require('simple-git/promise')();
 
 /**
  * Merges two comment-json files, with the original having higher priority.
- * @param {string} updatedCommentJson 
- * @param {string} originalCommentJson 
+ * Prunes away duplicated commented out props.
+ *
+ * @param {string} newJson 
+ * @param {string} originalJson 
  * @returns {string}
  */
-exports.mergeJson = function(updatedCommentJson, originalCommentJson) {
-  const merged = assign(parse(updatedCommentJson), parse(originalCommentJson));
+exports.mergeGlobalConfigs = function (newJson, originalJson) {
+  const originalParsed = parse(originalJson);
+  const commentsFromOriginal = parseCommentedOutConfig(originalParsed);
+  const newPruned = pruneDuplicatedComments(parse(newJson), commentsFromOriginal);
+  const merged = assign(newPruned, originalParsed);
   return stringify(merged, null, 2);
 }
+
+/**
+ * Parses "commented out" config values.
+ *
+ * @param {import('comment-json').CommentJSONValue} jsonWithComments
+ * @returns {import('comment-json').CommentToken[]}
+ */
+function parseCommentedOutConfig(commentJsonValue) {
+  return getPropCommentSymbols(commentJsonValue).flatMap(symbol => {
+    const commentArr = commentJsonValue[symbol] || [];
+    // We only care about non-inline LineComments that "look like" a config option
+    return commentArr.filter(c => {
+      return !c.inline && c.type === 'LineComment' && c.value.match(/^\s?"\w+":/)
+    })
+  });
+}
+exports.parseCommentedOutConfig = parseCommentedOutConfig;
+
+/**
+ * Removes all comments in CommentJSONValue that have the same value and type as the given
+ * commentsToDedupe.
+ *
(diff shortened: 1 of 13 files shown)


tests/postupgrade/utils.js
```diff
@@ -1,17 +1,93 @@
...
+      }
+    ]);
+  });
+
+  it('ignores block comments, and comments that dont "look like" config props', () => {
+    const commentJson = parse(`{
+      // ignore me
+      "experienceVersion": "<REPLACE ME>",
+      /* ignore me too */
+      "businessId": "<REPLACE ME>",
```
oshi97, line 63: will remove the trailing comma

postupgrade/utils.js
```diff
@@ -5,15 +5,77 @@ const simpleGit = require('simple-git/promise')();
...
+exports.mergeGlobalConfigs = function (newJson, originalJson) {
+  const originalParsed = parse(originalJson);
+  const commentsFromOriginal = parseCommentedOutProps(originalParsed);
+  const newPruned = pruneDuplicatedComments(parse(newJson), commentsFromOriginal);
+  const merged = assign(newPruned, originalParsed);
   return stringify(merged, null, 2);
 }
 
+/**
+ * Parses "commented out" config values.
```
tmeyer2115, line 23: Does this method get all of the "commented out" config values? If so, I think we should make that a bit clearer in either the function name or JSDocs.

postupgrade/utils.js
```diff
@@ -5,15 +5,77 @@ const simpleGit = require('simple-git/promise')();
...
+
+/**
+ * Removes all comments in CommentJSONValue that have the same value and type as the given
+ * commentsToDedupe.
+ *
+ * @param {import('comment-json').CommentJSONValue} commentJsonValue
+ * @param {import('comment-json').CommentToken[]} commentsToDedupe
+ * @returns {import('comment-json').CommentJSONValue} the updated value
+ */
+function pruneDuplicatedComments(commentJsonValue, commentsToDedupe) {
```
tmeyer2115, line 47: Small nit: Should we just call this `pruneComments(parsedJson, comments)`? That might be a bit simpler. If I understand correctly, the purpose of this method is to ensure the provided comments are removed from the `parsedJson`. This method isn't concerned with why those comments need to be removed (e.g. they are dupes), just that they are removed.

tmeyer2115 (commented):
Is this something we should file a bug report on to `comment-json`? From my understanding, this seems like something they should handle. But, maybe there is some nuance to our use case that I'm missing.

oshi97:
> Is this something we should file a bug report on to `comment-json`? From my understanding, this seems like something they should handle. But, maybe there is some nuance to our use case that I'm missing.

I think this is more our specific business logic than what comment-json is responsible for. We could maybe ask them it as a feature, but I wouldn't consider this a bug in their code. At any rate they definitely won't be quick to add support, because it would be a breaking API change.

 In order to differentiate between
``` jsonc
{
  // Put your api key below
  "apiKey": "blah"
}
```

```jsonc
{
  // "anotherProp": "this is a commented out value, not a descriptor for apiKey",
  "apiKey": "blah"
}
```

They would need to assume that the second comment "looks like" a json value, rather than a comment describing apiKey. We can safely do this, because we know we want all comments that "look like json" to be unique. On the other hand I think comment-json was more intended for "comments describing json".

postupgrade/utils.js
oshi97: yup! I'll remove all instances of words like dedupe from this method

tmeyer2115 (approved):

PR #917, opened 2021-08-13
Fix missing map pins on rtl site

yen-tt:
- use ltr instead of rtl attribute in the map div since switching to rtl would disrupt the coordinate to css mapping calculated by the map providers
- adjust map boundaries based on location of the result content wrapper, which could be left or right of the viewpoint (ltr or rtl)

J=SLAP-1515
TEST=manual

- launched multilang site, see that pins on map on arabic is present and the pin locations matched with the ones on english page.

### DIFF
--- a/static/js/theme-map/ThemeMapConfig.js
+++ b/static/js/theme-map/ThemeMapConfig.js
@@ -3,7 +3,7 @@ import { PinImages } from './PinImages.js';
 import { ClusterPinImages } from './ClusterPinImages.js';
 import { getLanguageForProvider } from './Util/helpers.js';
 import { defaultCenterCoordinate } from './constants.js';
-
+import isRTL from '../../common/rtl';
 /**
  * The configuration for the ThemeMap component.
  */
@@ -100,8 +100,8 @@ export default class ThemeMapConfig {
     this.padding = {
       top: () => window.innerWidth <= this.mobileBreakpointMax ? 150 : 50,
       bottom: () => 50,
-      right: () => 50,
-      left: () => this.getLeftVisibleBoundary(),
+      right: () => isRTL(this.language) ? this.getVisibleBoundary() : 50,
+      left: () => !isRTL(this.language) ? this.getVisibleBoundary() : 50
     };
 
     /**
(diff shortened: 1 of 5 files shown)


static/js/theme-map/ThemeMapConfig.js
```diff
@@ -3,7 +3,7 @@ import { PinImages } from './PinImages.js';
 import { ClusterPinImages } from './ClusterPinImages.js';
 import { getLanguageForProvider } from './Util/helpers.js';
 import { defaultCenterCoordinate } from './constants.js';
-
+import isRTL from '../../common/rtl';
```
oshi97, line 6: nit: line break after

static/js/theme-map/ThemeMapConfig.js
```diff
@@ -238,11 +238,12 @@ export default class ThemeMapConfig {
   }
 
   /**
-   * Get the leftmost point on the map, such that pins will still be visible
+   * Get the leftmost or rightmost point on the map, such that pins will still be visible
```
oshi97, line 241: Maybe `(or rightmost for RTL locales)`?

templates/vertical-full-page-map/markup/map.hbs
```diff
@@ -1 +1 @@
-<div id="js-answersMap" class="Answers-map js-answersMap"></div>
+<div id="js-answersMap" class="Answers-map js-answersMap" dir="ltr"></div>
```
oshi97, line 1: we don't want to flip the map for rtl?

yen-tt: it mess with and pin placement (specifically the coordinate to css styling translation) on the map

oshi97: does this cause the map to not be oriented correctly, for either google/mapbox? Like disregarding pin stuff, does anything else change if you add dir="ltr" on a rtl page

yen-tt: the map is oriented the same way for both ltr and rtl, only the pins are affected. the div that I added `dir='ltr'` only contains map related things like the canvas for the map and the pin clusters which we don't want rtl on. There's also the map control html elements but the position for those are specify for ltr and rtl in the map provider construction code. everything else on the page would follow ltr. tldr: nothing else change when I looked at the pages

oshi97: got it awesome!!

oshi97 (approved):

PR #923, opened 2021-08-18
Add table header translations

yen-tt:
- add table header translations for new languages
- move the code to extract language from locale to utils

TEST=none


static/js/hours/table/table-strings-i18n.js
```diff
@@ -28,6 +29,56 @@ export default function provideTableHeadersTranslation(locale) {
...
+      return {
+        [TableHeaders.DAY_OF_WEEK]: 'День недели',
+        [TableHeaders.HOURS]: 'Часы'
+      }
+    case 'sv':
+      return {
+        [TableHeaders.DAY_OF_WEEK]: 'Veckodag',
+        [TableHeaders.HOURS]: 'Timmar'
+      }
+    case 'zh-CH':
```
tmeyer2115, line 72: We're actually in the process of cutting over to 'zh-Hans' and 'zh-Hant'. Turns out, that's what LiveAPI and KG use instead. We want to be consistent with the rest of Yext. You should just be able to update the key names here.

static/js/utils.js
```diff
@@ -13,4 +13,16 @@ export function canonicalizeBoolean (value) {
...
   }
+}
+
+/**
+ * returns language extracted from the provided locale
+ * @param {string} locale 
+ * @returns {string} language
+ */
+export function getLanguageFromLocale(locale) {
+  const language = (locale === 'zh-CN' || locale === 'zh-TW')
```
tmeyer2115, line 24: This method will also need to be updated to use the new language codes I mentioned above. I'd also let @cea2aj and @oshi97 know you made this method. I think one of them was working on something similar for a different PR.

yen-tt: I see that connor made a parse locale function in his SDK pr. Will look over that and probably use that in theme

static/js/open-status-i18n.js
```diff
@@ -1,14 +1,12 @@
...
+import { getLanguageFromLocale } from './utils'; 
 /**
  * Provides translation data for the openStatus formatter.
  * @param {string} locale
  */
 export default function provideOpenStatusTranslation (locale) {
-  const language = (locale === 'zh-CN' || locale === 'zh-TW') ? 
-    locale :
-    locale.substring(0,2);
+  const language = getLanguageFromLocale(locale);
```
cea2aj, line 9: I have a PR open for the open status translations (https://github.com/yext/answers-hitchhiker-theme/pull/927) so this isn't needed here

cea2aj (approved):

PR #924, opened 2021-08-18
Automation process for static translations

yen-tt:
J=SLAP-1477
TEST=compile & manual

### DIFF
--- a/static/js/generate-translation.js
+++ b/static/js/generate-translation.js
@@ -1,0 +1,123 @@
+const fs = require('fs');
+const path = require('path');
+const themePath = process.argv[2];
+const outputPath = process.argv[3];
+
+const openStatusHeader = `
+import { DayNames } from './hours/constants.js';
+import { OpenStatusStrings } from './hours/open-status/constants.js';
+import { getLanguageFromLocale } from './utils';
+
+export default function provideOpenStatusTranslation (locale) {
+  const language = getLanguageFromLocale(locale);\n`;
+
+const tableStringHeader = `
+import { TableHeaders } from '../constants';
+import { getLanguageFromLocale } from '../../utils';
+
+export default function provideTableHeadersTranslation(locale) {
+  const language = getLanguageFromLocale(locale);\n`;
+
+
+const openstatustranslation = {
+  generatedfilepath: `${outputPath}/static/js/open-status-i18n.js`,
+  header: openStatusHeader,
+  potfilepath: `${themePath}/translations/open-status.pot`
+}
+
+const tablestringtranslation = {
+  generatedfilepath: `${outputPath}/static/js/hours/table/table-strings-i18n.js`,
+  header: tableStringHeader,
+  potfilepath: `${themePath}/translations/table-strings.pot`
+}
+
+function generateStaticTranslationFile({generatedfilepath, header, potfilepath}) {
+  let data = header;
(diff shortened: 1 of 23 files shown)


cea2aj:
This solution looks solid, however I fear that it might be difficult to maintain and change. I have an idea for another solution which goes like this:

1. We store a single static.pot file for all static translations
2. We use [https://github.com/i18next/i18next-gettext-converter](https://github.com/i18next/i18next-gettext-converter) to load the po file and get the translations for each language. We can then generate a single json file (or a javascript file which contains a single object) where each key is the language/modifer code and the values are the translations for that locale. I don't think the generation of the file needs to be part of the webpack build chain. It can just be made available as a script in the package.json file
3. We create a getTranslation function which takes in a language/modifer and the translation key (e.g. "Closes at") and returns the appropriate translation from that generated file. We can ignore the translation contexts for the sake of simplicity
4. We update the stringsLocalizer.getTranslation method to use the getTranslation function mentioned above. This solution would also easily support the Yes/No translations

With that being said, my bad for not coming up with this approach sooner. I suppose that I was hoping that we would be able to leverage our existing i18n infrastructure for translating the static files.

I would like to get @oshi97  and @tmeyer2115  's thoughts on if it's worth switching to the solution described above. If not, we can move forward with this and I can take a closer look at the PR

yen-tt:
> 1. We store a single static.pot file for all static translations
> 2. We use https://github.com/i18next/i18next-gettext-converter to load the po file and get the translations for each language. We can then generate a single json file (or a javascript file which contains a single object) where each key is the language/modifer code and the values are the translations for that locale. I don't think the generation of the file needs to be part of the webpack build chain. It can just be made available as a script in the package.json file
> 3. We create a getTranslation function which takes in a language/modifer and the translation key (e.g. "Closes at") and returns the appropriate translation from that generated file. We can ignore the translation contexts for the sake of simplicity
> 4. We update the stringsLocalizer.getTranslation method to use the getTranslation function mentioned above. This solution would also easily support the Yes/No translations

This solution seems good as well! The context from .po files would need to extracted out before calling i18next-gettext-conv since it looks like it attach context with the translation string as keys in the json object. If the translations or .po files for static.pot file gets merge in with the other existing .po files then maybe we would filter out the unrelated translations from the json files (but that's not needed if we will have a separate set of .po files for the static.pot). All those json files would need to be imported in the file with getTranslation() or gets copy to site's folder to read I think.

cea2aj:
> This solution seems good as well! The context from .po files would need to extracted out before calling i18next-gettext-conv since it looks like it attach context with the translation string as keys in the json object. If the translations or .po files for static.pot file gets merge in with the other existing .po files then maybe we would filter out the unrelated translations from the json files (but that's not needed if we will have a separate set of .po files for the static.pot). All those json files would need to be imported in the file with getTranslation() or gets copy to site's folder to read I think.

You're right, i18next-gettext-conv does include the context as part of the string. If it's simpler, we could keep that behavior and supply the right context when getting the translation. For example, we could call `getTranslation({phrase: 'Open status', context: 'The hours that business is open'});` This would would work the same as our existing translation setup in jambo and the SDK

oshi97:
I feel like we should try to use jambo if we can, instead of rewriting i18n another way. That would also have the added benefit of working automatically with the jambo translation extractor.

PR #926, opened 2021-08-18
Unit additional language support

cea2aj:
Support additional languages for the distance formatters

Rather than enumerating each locale, I opted to make the unit map only contain locales with imperial units since everything else will default to metric

J=SLAP-1529
TEST=auto

Run tests

### DIFF
--- a/static/js/units-i18n.js
+++ b/static/js/units-i18n.js
@@ -26,8 +26,7 @@ export const LOCALE_UNIT_MAP = {
     default: IMPERIAL
   },
   es: {
-    US: IMPERIAL,
-    default: METRIC
+    US: IMPERIAL
   }
 };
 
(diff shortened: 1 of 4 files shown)


static/js/units-i18n.js
```diff
@@ -57,12 +49,7 @@ export function getDistanceUnit(locale){
  * @returns {Object}
  */
 function getUnitsForLocale(locale) {
-  const language = locale.substring(0,2);
-  // Note: Getting region this way will be invalid if script tags are used in the future.
-  // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl
-  // Optionally, we can use Intl.locale() but we would need to polyfill it for IE11
-  const region = locale.substring(3,5); 
-  const unitSystemFallback = METRIC;
```
oshi97, line 65: do we still need unitSystemFallback?

static/js/units-i18n.js
```diff
@@ -26,21 +28,11 @@ export const LOCALE_UNIT_MAP = {
   es: {
     US: IMPERIAL,
     default: METRIC
```
oshi97, line 30: would it be possible to get rid of this default too? maybe with like a `LOCALE_UNIT_MAP[language]['default'] || METRIC` on line 61

static/js/units-i18n.js
cea2aj: The code is still using it. I just moved it below LOCALE_UNIT_MAP

static/js/units-i18n.js
cea2aj: sure

oshi97 (approved):

PR #930, opened 2021-08-19
Process chinese language locale for theme's map

yen-tt:
- add support for chinese language locale in google map and mapbox
- update jest test

J=SLAP-1527
TEST=auto

jest test

### DIFF
--- a/static/js/theme-map/Util/helpers.js
+++ b/static/js/theme-map/Util/helpers.js
@@ -14,17 +14,17 @@ import { parseLocale } from '../../utils.js';
  */
 const getLanguageForProvider = (localeStr, mapProvider) => {
   const { language, modifier, region } = parseLocale(localeStr);
-  if(language.length != 2) {
+  if (language.length !== 2) {
     return 'en';
   }
   const provider = getMapProvider(mapProvider);
-  let formattedLocaleStr = modifier? `${language}-${modifier}` : language;
-  if(provider === GoogleMaps) {
-    if(region) {
+  let formattedLocaleStr = modifier ? `${language}-${modifier}` : language;
+  if (provider === GoogleMaps) {
+    if (region) {
       formattedLocaleStr = `${language}-${region}`;
-    } else if(modifier === 'Hant') {
+    } else if (modifier === 'Hant') {
       formattedLocaleStr = `${language}-TW`;
-    } else if(modifier === 'Hans') {
+    } else if (modifier === 'Hans') {
       formattedLocaleStr = `${language}-CN`;
     }
   }
--- a/static/js/theme-map/Util/helpers.js
+++ b/static/js/theme-map/Util/helpers.js
@@ -18,7 +18,7 @@ const getLanguageForProvider = (localeStr, mapProvider) => {
     return 'en';
   }
   const provider = getMapProvider(mapProvider);
-  let formattedLocaleStr = localeStr.replace('_', '-');
+  let formattedLocaleStr = modifier? `${language}-${modifier}` : language;
   if(provider === GoogleMaps) {
     if(region) {
       formattedLocaleStr = `${language}-${region}`;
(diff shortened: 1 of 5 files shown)


static/js/theme-map/Util/helpers.js
```diff
@@ -1,30 +1,37 @@
...
-    const formattedLocaleStr = localeStr.replace('_', '-');
-    if (provider.getSupportedLocales().includes(formattedLocaleStr)) {
-      return formattedLocaleStr;
+  const { language, modifier, region } = parseLocale(localeStr);
+  if(language.length != 2) {
+    return 'en';
+  }
+  const provider = getMapProvider(mapProvider);
+  let formattedLocaleStr = modifier? `${language}-${modifier}` : language;
+  if(provider === GoogleMaps) {
```
oshi97, line 22: spaces after if

static/js/theme-map/Util/helpers.js
```diff
@@ -1,30 +1,37 @@
...
-    return localeStr;
-  } 
-
-  if (localeStr.length > 2) {
-    const provider = getMapProvider(mapProvider);
-    const formattedLocaleStr = localeStr.replace('_', '-');
-    if (provider.getSupportedLocales().includes(formattedLocaleStr)) {
-      return formattedLocaleStr;
+  const { language, modifier, region } = parseLocale(localeStr);
+  if(language.length != 2) {
```
oshi97, line 17: !==

static/js/theme-map/Util/helpers.js
```diff
@@ -1,30 +1,37 @@
...
+    } else if (modifier === 'Hans') {
+      formattedLocaleStr = `${language}-CN`;
     }
-    return localeStr.substring(0, 2);
   }
-  return 'en';
+  if (provider.getSupportedLocales().includes(formattedLocaleStr)) {
+    return formattedLocaleStr;
+  }
+  return language;
```
oshi97, line 34: just double checking we want to return just `zh` if say `zh-hans-cn` is not supported

yen-tt: are you referring to the case where `zh-hans` is supported but not `zh-hans-cn`, we should return `zh`?

static/js/theme-map/Util/helpers.js
```diff
@@ -1,30 +1,37 @@
...
-    return localeStr;
-  } 
-
-  if (localeStr.length > 2) {
-    const provider = getMapProvider(mapProvider);
-    const formattedLocaleStr = localeStr.replace('_', '-');
-    if (provider.getSupportedLocales().includes(formattedLocaleStr)) {
-      return formattedLocaleStr;
+  const { language, modifier, region } = parseLocale(localeStr);
+  if (language.length !== 2) {
```
cea2aj, line 17: Is defaulting to 'en' necessary here? If you do remove this if statement, remember to update the function description

static/js/theme-map/Util/helpers.js
oshi97: oh I see formattedLocaleStr never is equal to something like `zh-hans-cn`, would it be possible to move the GoogleMaps specific logic here into the provider itself? Maybe a `transformLocaleForProvider` or something

cea2aj: In the SDK, `zh-hans-cn` would return `zh-CN` for google maps. I think we should have the same behavior here

oshi97: yeah I misread the code @cea2aj I assumed formattedLocale would just be the locale canonicalized jambo style

static/js/theme-map/Util/helpers.js
yen-tt: the old code default to en if the locale doesn't follow 2 characters lang code (e.g. if user pass in 'mul' or 'r')

cea2aj (approved):

PR #935, opened 2021-08-25
Automation process for static translations

yen-tt:
Automate translation process for static translations using jambo translation helper functionality

- move translations from static folder to a partial. This include open-status, headers, and prettyPrintBoolean strings.
- use the translations partial in html.hbs layout to put translations in window
- access window.translations in `stringlocalizer.js` and formatters-internal.js` to pull the appropriate translated string
- remove now unused translations js files

J=SLAP-1477
TEST=compile & manual

launched test site with en and es locales, see that the text in KM page is translated propertly.


static/js/hours/open-status/constants.js
```diff
@@ -11,6 +11,6 @@ export const OpenStatusStrings = {
   OPEN_24_HOURS: 'Open 24 Hours',
   CLOSED: 'Closed',
   OPENS_AT: 'Opens at',
-  OPEN_NOW: 'Open now',
+  OPEN_NOW: 'Open Now',
```
oshi97, line 14: Will Hitchhikers think we've added a typo if we change this?

script/translations.js
```diff
@@ -0,0 +1,18 @@
+window.translations = {
```
oshi97, line 1: Could we use a more unique sounding variable name? Just cause stuff on the window can get a little messy. Alternatively if you have any ideas for not putting this on the window! I thought briefly about either jambo building out a static js file for us that we could import directly, but that seemed like a lot of gymnastics just for this.

script/translations.js
```diff
@@ -0,0 +1,18 @@
...
+  "Closed": {{ translateJS phrase="Closed" context="Indicates that a store is closed" }},
+  "Open 24 Hours": {{ translateJS phrase="Open 24 Hours" context="The hours that a store is open" }},
+  "Opens at": {{ translateJS phrase="Opens at" context="Indicates the time a store will open" }},
+  "Open Now": {{ translateJS phrase="Open Now" context="Indicates that a store is open" }},
+  "Closes at": {{ translateJS phrase="Closes at" context="Indicates the time a store will close" }},
+  "Day of the Week": {{ translateJS phrase="Day of the Week"}},
+  "Hours": {{ translateJS phrase="Hours"}},
+  "Yes": {{ translateJS phrase="Yes" }},
+  "No": {{ translateJS phrase="No" }},
+  MONDAY: {{ translateJS phrase="Monday" }},
```
oshi97, line 11: Would it make sense to make the days of the week keys equal to the phrase too? (like same MONDAY -> Monday to match the phrase)

yen-tt: the keys used here match with the text pass to `stringlocalizer.js` getTranslation(text). In this case the week days are all passed in as all capitalized. We would have to do some string transformation inside getTranslation() if we want to change this to match the phrase

script/translations.js
yen-tt: what about `hitchhiker_theme_translations`? I'm not sure if there's another way of passing information between partial and static file at the moment. Jambo doesn't generate any static js files currently right? so that would need some changes to support that

static/js/hours/open-status/constants.js
yen-tt: There was some inconsistency in the theme. From the `open-status-i18n.js` file, the default english text to display is `Open **N**ow`. So I made the msgid as 'Open **N**ow' but the the translations for other languages seems to follow 'Open **n**ow'. I could revert this change and update the .po/.pot files to all use 'Open **n**ow'. This mean the english text would now be `'Open **n**ow'`. (percy will note a bunch of changes in snapshots since english text would be 'Open **n**ow'). What do you think?

oshi97: Ah, yeah let's just not change any behavior! I assumed the current text was "Open now"

script/translations.js
oshi97: yeah that works! Just something a little more unique for when you need to `ctrl + f` it, since you can use window variables without specifying the `window.` prefix

script/translations.js
oshi97: ok sgtm!

PR #938, opened 2021-08-31
Add feedback buttons to cards

nmanu1:
Add thumbs up/down feedback buttons to all cards without doing card-specific styling yet. Created a shared partial for the buttons, as well as a new config option in 'dataForRender', called 'feedback' that controls whether or not the buttons appear. Used the shared partial in all cards, including direct answers cards.

J=SLAP-1544
TEST=manual

Smoke testing to see if thumbs up/down buttons appear on each card


tmeyer2115:
We can ignore the failing Translation Test for now. The Theme should not be running that test on every PR. Fixing this check will be done as part of another item.

cards/common-partials/thumbsfeedback.hbs
```diff
@@ -0,0 +1,29 @@
+<form class="HitchhikerCard-thumbs js-HitchhikerCard-feedbackForm">
+  <fieldset class="HitchhikerCard-fieldset">
+    <label class="HitchhikerCard-thumb">
+      <span class="HitchhikerCard-thumbUpIcon">
+        {{> icons/builtInIcon iconName='thumb' }}
+      </span>
+      <input type="radio"
```
cea2aj, line 7: I'm a little surprised these are radio inputs rather than buttons, but since that's the current direct answers feedback works, I think we're good to leave it

cea2aj:
Based on the Percy snapshots, I think some of the feedback buttons need to be adjusted so that they are on the bottom right of the cards. We may also need to add some more padding and make the buttons a little smaller

cards/document-standard/template.hbs
```diff
@@ -90,4 +93,20 @@
   </a>
 </div>
 {{/if}}
+{{/inline}}
+
+{{#*inline 'feedback'}}
```
cea2aj, line 98: What do you think about making a feedbackWrapper partial? It looks like every cards follows this same structure, but with different class names. We may be able to make a partial which allows the card class name to be passed in as a param, and then we wouldn't have to define this inline helper in every card.

cards/document-standard/component.js
```diff
@@ -48,7 +48,11 @@ class document_standardCardComponent extends BaseCard['document-standard'] {
         eventType: 'CTA_CLICK',
         eventOptions: this.addDefaultEventOptions(),
         // ariaLabel: '',
-      }
+      },
+      feedback: false, // Shows thumbs up/down buttons to provide feedback on the result card
```
cea2aj, line 52: These are set to false on every card but the buttons are still appearing on every card on the test site

nmanu1: I'll add that

cards/common-partials/thumbsfeedback.hbs
```diff
@@ -0,0 +1,29 @@
+<form class="HitchhikerCard-thumbs js-HitchhikerCard-feedbackForm">
```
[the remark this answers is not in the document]
nmanu1: We could because it's almost exactly the same. The only difference is that direct answers card feedback has text that accompanies the buttons. If we wanted to combine them, it wouldn't be hard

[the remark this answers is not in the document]
cea2aj: If we can combine them, that would be great, but I would prioritize the positioning of the buttons first

cards/common-partials/thumbsfeedback.hbs
```diff
@@ -0,0 +1,43 @@
+{{#if card.feedback}}
```
tmeyer2115, line 1: Could this partial be used with the DA cards as well? It seems like it would be nice for both results and DA cards to use the same partial for feedback.

cards/document-standard/component.js
```diff
@@ -48,7 +48,11 @@ class document_standardCardComponent extends BaseCard['document-standard'] {
...
+      feedbackTextOnSubmission: 'Thanks for the feedback!', // Text to display after a thumbs up/down is clicked
```
tmeyer2115, line 53: I think this message should be consistent between the results cards and the DA cards. The DA cards have it as "Thank you for your feedback!". I think have consistency between these messages will also make the translation test failure go away. We already have translations for the DA cards static text.

nmanu1: Makes sense. I'll change that

cards/common-partials/thumbsfeedback.hbs
```diff
@@ -0,0 +1,46 @@
+<div class="{{cardType}}-feedbackWrapper">
+  <div class="{{cardType}}-feedback">
+    {{#if feedbackSubmitted}}
+      <div class="{{cardType}}-feedbackText">
+        {{feedbackTextOnSubmission}}
```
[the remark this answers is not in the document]
nmanu1: There is feedbackTextOnSubmission for results cards, just no feedbackText

tmeyer2115 (approved):

PR #939, opened 2021-09-02
Add feedback analytics

nmanu1:
Add analytics for the thumbs up/down feedback buttons on cards. Update the analytics on direct answer cards to include additional event attributes that are consistent with those for results cards.

J=SLAP-1545
TEST=manual

Smoke testing to see if clicking feedback buttons on each card sent the correct analytics event and displayed the feedback submission message

### DIFF
--- a/cards/card_component.js
+++ b/cards/card_component.js
@@ -9,6 +9,11 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
 
     this.verticalKey = data.verticalKey;
     this.result = data.result || {};
+
+    this.globalOptions = this.analyticsReporter && this.analyticsReporter._globalOptions;
+    this.experienceKey = this.globalOptions && this.globalOptions.experienceKey;
+    this.experienceVersion = this.globalOptions && this.globalOptions.experienceVersion;
+    this.queryId = this.globalOptions && this.globalOptions.queryId;
   }
 
   /**
@@ -25,6 +30,34 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
           excessDetailsEls.forEach(detailsEl => detailsEl.classList.toggle('js-hidden'));
         })
       );
+    }
+
+    let feedbackFormSelector = '.js-HitchhikerCard-feedbackForm';
+    let feedbackFormEl = this._container.querySelector(feedbackFormSelector);
+    if (feedbackFormEl) {
+      // For WCAG compliance, the feedback should be a submittable form
+      feedbackFormEl.addEventListener('submit', (e) => {
+        let formTargetEl = e.target;
+        let checkedValue = formTargetEl.querySelector('input:checked').value === 'true';
+
+        this.reportQuality(checkedValue);
+        this.updateState({
+          feedbackSubmitted: true
+        });
+      });
+
+      let thumbSelectorEls = this._container.querySelectorAll('.js-HitchhikerCard-thumbInput');
+      if (thumbSelectorEls) {
+        thumbSelectorEls.forEach(el => {
+          el.addEventListener('click', (e) => {
+            let input = el.querySelector('input');
+            if (input) {
+              input.checked = true;
+            }
+            this._triggerCustomEvent(feedbackFormSelector, 'submit');
+          });
+        });
+      }
     }
     
     const rtfElement = this._container.querySelector('.js-yxt-rtfValue');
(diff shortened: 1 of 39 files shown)


cards/card_component.js
```diff
@@ -26,6 +31,34 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
         })
       );
     }
+
+    let feedbackFormSelector = '.js-HitchhikerCard-feedbackForm';
```
cea2aj, line 35: We typically use `const` instead of `let` when the value doesn't change. There are a few spots in this pr where we can use `const` instead of let

cards/card_component.js
```diff
@@ -9,6 +9,11 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
 
     this.verticalKey = data.verticalKey;
     this.result = data.result || {};
+
+    this.globalOptions = this.analyticsReporter && this.analyticsReporter._globalOptions;
```
cea2aj, line 13: nit: can we use optional chaining here? That would let us shorten these lines. Here's an example: `this.globalOptions = this.analyticsReporter?._globalOptions;`

cards/card_component.js
```diff
@@ -80,6 +113,73 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
...
+      THUMBS_UP: 'THUMBS_UP',
+      THUMBS_DOWN: 'THUMBS_DOWN'
+    };
+    const eventType = isGood === true ? EventTypes.THUMBS_UP : EventTypes.THUMBS_DOWN;
+    const event = new ANSWERS.AnalyticsEvent(eventType)
+      .addOptions({
+        directAnswer: false,
+        experienceKey: this.experienceKey,
+        experienceVersion: this.experienceVersion,
+        queryId: this.queryId,
```
cea2aj, line 165: I may be mistaken, but I thought that the queryId was automatically sent when analytics are reported? Is that not the behavior you are seeing?

cards/card_component.js
```diff
@@ -26,6 +31,34 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
...
       );
     }
+
+    let feedbackFormSelector = '.js-HitchhikerCard-feedbackForm';
+    let feedbackFormEl = this._container.querySelector(feedbackFormSelector);
+    if (feedbackFormEl) {
+      // For WCAG compliance, the feedback should be a submittable form
+      feedbackFormEl.addEventListener('submit', (e) => {
+        let formTargetEl = e.target;
+        let checkedValue = formTargetEl.querySelector('input:checked').value === 'true';
```
cea2aj, line 41: nit: would `isChecked` be a little more accurate of a name?

cards/card_component.js
```diff
@@ -26,6 +31,34 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
         })
       );
     }
+
```
cea2aj, line 34: nit: can we move the registration of these listeners into a function called something like: `addFeedbackListeners()`?

cards/card_component.js
```diff
@@ -9,6 +9,11 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
 
     this.verticalKey = data.verticalKey;
     this.result = data.result || {};
+
+    this.globalOptions = this.analyticsReporter && this.analyticsReporter._globalOptions;
+    this.experienceKey = this.globalOptions && this.globalOptions.experienceKey;
+    this.experienceVersion = this.globalOptions && this.globalOptions.experienceVersion;
```
oshi97, line 15: just double checking that globalOptions aren't automatically part of the analytics event?

cards/card_component.js
```diff
@@ -80,6 +113,73 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
...
+    let e = this._customEvent(event, settings);
+    this._container.querySelector(selector).dispatchEvent(e);
+  }
+
+  /**
+   * _customEvent is an event constructor polyfill
+   * @param {string} event event to fire
+   * @param {Object} settings additional settings
+   */
+  _customEvent(event, settings) {
```
oshi97, line 132: I think this is duplicated with the base direct answer card, is there a common place we can put this?

cards/card_component.js
nmanu1: Just checked, and that is the case, so I'll remove it from here

cards/card_component.js
nmanu1: Just checked, and they are, so I'll remove these from here

cards/card_component.js
nmanu1: It seems like checkedValue indicates which button (thumbs up or down) is clicked, not whether there is a click. So checkedValue is false when a thumbs down is selected

cea2aj: ah, got it. Maybe we should rename to `isGood` to be consistent then?

nmanu1: Makes sense

directanswercards/card_component.js
```diff
@@ -40,64 +40,40 @@ BaseDirectAnswerCard["{{componentName}}"] = class extends ANSWERS.Component {
...
-        this.reportQuality(checkedValue);
+        this.reportQuality(isGood);
         this.updateState({
-          'feedbackSubmitted': true
+          feedbackSubmitted: true
         });
       });
 
-      let thumbSelectorEls = this._container.querySelectorAll('.js-HitchhikerDirectAnswerCard-thumbInput');
+      let thumbSelectorEls = this._container.querySelectorAll('.js-HitchhikerCard-thumbInput');
```
oshi97, line 64: Do we need to worry about backwards compatibility, if somebody forked one fo the direct answer cards?

oshi97: (If we already asked Rose and got her approval then we good here!)

nmanu1: This change is because the direct answer cards are now using the same shared feedback partial as results cards. So, this would be a problem for someone forking a previous version of the direct answer card

nmanu1: I think Tom had mentioned that he would let Rose know that we were switching the direct answer cards to use the shared partial

tmeyer2115: It's not an issue for there to be a shared feedback partial between DA and Results cards. But, we can't break any custom DA cards a user has created. If someone made a custom DA card before this change, that means they are using now invalid classnames. So, feedback submission wouldn't work. 

We can still have a shared partial and be backwards compatible. Instead of using the class name `js-HitchhikerCard` for both, we would make it dynamic. When a DA card invokes the partial, the class names would have a `.js-HitchhikerDirectAnswerCard` prefix.

nmanu1: Ok, I'll fix that

oshi97 (approved):
lgtm!

PR #940, opened 2021-09-07
Add card-specific feedback button styling

nmanu1:
Add card-specific styling for thumbs up/down feedback buttons on all results cards.

J=SLAP-1544
TEST=manual

Smoke testing to see if buttons match mock for each card

### DIFF
--- a/cards/common-partials/thumbsfeedback.hbs
+++ b/cards/common-partials/thumbsfeedback.hbs
@@ -1,53 +1,54 @@
 {{#if directAnswerClass}}
-  {{assign 'className' directAnswerClass}}
-  {{assign 'cardType' 'HitchhikerDirectAnswerCard'}}
+  {{> feedback className=directAnswerClass cardType='HitchhikerDirectAnswerCard'}}
 {{else}}
-  {{assign 'className' 'HitchhikerCard'}}
-  {{assign 'cardType' 'HitchhikerCard'}}
+  {{> feedback className='HitchhikerCard' cardType='HitchhikerCard'}}
 {{/if}}
-<div class="{{@root.className}}-feedbackWrapper">
-  <div class="{{@root.className}}-feedbackContent">
-    {{#if feedbackSubmitted}}
-      <div class="{{@root.className}}-feedbackText">
-        {{feedbackTextOnSubmission}}
-      </div>
-    {{else}}
-      {{#if feedbackText}}
-        <div class="{{@root.className}}-feedbackText">
-          {{feedbackText}}
+
+{{#*inline 'feedback'}}
+  <div class="{{className}}-feedbackWrapper">
+    <div class="{{className}}-feedbackContent">
+      {{#if feedbackSubmitted}}
+        <div class="{{className}}-feedbackText">
+          {{feedbackTextOnSubmission}}
+        </div>
+      {{else}}
+        {{#if feedbackText}}
+          <div class="{{className}}-feedbackText">
+            {{feedbackText}}
+          </div>
+        {{/if}}
+        <div class="{{className}}-thumbsWrapper">
+          <form class="HitchhikerCard-thumbs js-{{cardType}}-feedbackForm">
+            <fieldset class="HitchhikerCard-fieldset">
+              <label class="HitchhikerCard-thumb">
+                <span class="HitchhikerCard-thumbUpIcon">
+                  {{> icons/builtInIcon iconName='thumb' }}
+                </span>
+                <input type="radio"
+                  name="feedback"
+                  value="true"
+                  class="HitchhikerCard-feedback HitchhikerCard-thumbUpButton js-{{cardType}}-thumbInput">
+                <span class="sr-only">
+                  {{positiveFeedbackSrText}}
+                </span>
+              </label>
+              <label class="HitchhikerCard-thumb">
+                <span class="HitchhikerCard-thumbDownIcon">
+                  {{> icons/builtInIcon iconName='thumb' }}
+                </span>
+                <input type="radio"
+                  name="feedback"
+                  value="false"
+                  class="HitchhikerCard-feedback HitchhikerCard-thumbDownButton js-{{cardType}}-thumbInput">
+                <span class="sr-only">
+                  {{negativeFeedbackSrText}}
+                </span>
+              </label>
+            </fieldset>
+            <button type="submit" class="sr-only sr-only-focusable">Send feedback</button>
+          </form>
         </div>
       {{/if}}
-      <div class="{{@root.className}}-thumbsWrapper">
-        <form class="HitchhikerCard-thumbs js-{{@root.cardType}}-feedbackForm">
-          <fieldset class="HitchhikerCard-fieldset">
-            <label class="HitchhikerCard-thumb">
-              <span class="HitchhikerCard-thumbUpIcon">
-                {{> icons/builtInIcon iconName='thumb' }}
-              </span>
-              <input type="radio"
-                name="feedback"
-                value="true"
-                class="HitchhikerCard-feedback HitchhikerCard-thumbUpButton js-{{@root.cardType}}-thumbInput">
-              <span class="sr-only">
-                {{positiveFeedbackSrText}}
-              </span>
-            </label>
-            <label class="HitchhikerCard-thumb">
-              <span class="HitchhikerCard-thumbDownIcon">
-                {{> icons/builtInIcon iconName='thumb' }}
-              </span>
-              <input type="radio"
-                name="feedback"
-                value="false"
-                class="HitchhikerCard-feedback HitchhikerCard-thumbDownButton js-{{@root.cardType}}-thumbInput">
-              <span class="sr-only">
-                {{negativeFeedbackSrText}}
-              </span>
-            </label>
-          </fieldset>
-          <button type="submit" class="sr-only sr-only-focusable">Send feedback</button>
-        </form>
-      </div>
-    {{/if}}
+    </div>
   </div>
-</div>
+{{/inline}}
--- a/cards/common-partials/thumbsfeedback.hbs
+++ b/cards/common-partials/thumbsfeedback.hbs
@@ -1,23 +1,24 @@
-<div class="{{cardType}}-feedbackWrapper">
-  <div class="{{cardType}}-feedbackContent">
+{{#if directAnswerClass}}
+  {{assign 'className' directAnswerClass}}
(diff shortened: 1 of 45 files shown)


cards/multilang-faq-accordion/component.js
```diff
@@ -68,11 +69,28 @@ class multilang_faq_accordionCardComponent extends BaseCard['multilang-faq-accor
 
     const contentEl = this._container.querySelector(accordionContentSelector);
     let isExpanded = this._container.querySelector(`.${accordionExpandedClass}`);
-    contentEl.style.height = `${isExpanded ? contentEl.scrollHeight : 0}px`;
+    let stayExpanded = this._container.querySelector(`.${stayExpandedClass}`);
+
+    const cardEl = this._container.querySelector(accordionCardSelector);
     const linkEls = contentEl.querySelectorAll('a');
+
+    if (stayExpanded && this._state._state.feedbackSubmitted) {
```
oshi97, line 77: I know what this._state is, but what is this._state._state?

oshi97: is there a getter we can use like this.getState()?

cards/event-standard/template.hbs
```diff
@@ -52,7 +52,7 @@
         </div>
         {{/if}}
         {{#if card.feedback}}
-          {{> thumbsfeedback card cardType="HitchhikerEventStandard" feedbackSubmitted=feedbackSubmitted}}
+          {{> thumbsfeedback card cardType="HitchhikerCard" feedbackSubmitted=feedbackSubmitted}}
```
oshi97, line 55: maybe we shouldn't call this cardType anymore? Alternatively we could just always have the `HitchhikerCard-[blah]` css class on the feedback partial and only have special styling for the ones that currently have cardType="not HitchhikerCard"

static/scss/answers/cards/product-prominentimage.scss
```diff
@@ -112,4 +112,36 @@
...
+  &-feedbackWrapper
+  {
+    display: flex;
+    flex-grow: 1;
+    background-color: var(--yxt-color-background-highlight);
+  }
+
+  &-feedbackContent
+  {
+    display: flex;
```
oshi97, line 130: if there's repeated styling you can try using a mixin to factor them out

cards/professional-location/template.hbs
```diff
@@ -157,6 +157,7 @@
 {{/inline}}
 
 {{#*inline 'ctas'}}
+{{!-- skdf,zksjdbvv vzxjvzlxcjk vlxj cx --}}
```
oshi97, line 160: skdf,zksjdbvv vzxjvzlxcjk vlxj cx

cards/multilang-faq-accordion/component.js
nmanu1: That's how state variables seemed to be stored. I can add a getter for that

cards/event-standard/template.hbs
nmanu1: I'll change that

cards/common-partials/thumbsfeedback.hbs
```diff
@@ -1,23 +1,24 @@
-<div class="{{cardType}}-feedbackWrapper">
-  <div class="{{cardType}}-feedbackContent">
+{{#if directAnswerClass}}
+  {{assign 'className' directAnswerClass}}
```
oshi97, line 2: Instead of using the SDK's built in helper, can we make a version of it for the theme that adds the variable in the current scope, instead of the root scope? Adding variables to data.root affects the global scope which could bite us later. You can register a custom helper using `ANSWERS.registerHelper` (we already have a few of these). The last param of every handlebars helper is a special "HelperOptions" param, which is what the SDK's assign helper is mutating
typescript interface is here but they have an "any" type for HelperOptions.data which is not that helpful, but you can do console.logs to inspect the shape
https://github.com/handlebars-lang/handlebars.js/blob/6790c080c641ef2b44e663800e1794fae180977a/types/index.d.ts#L45

oshi97: alternatively, if you don't want to mess with that, you could make the feedback HTML an inline partial, and that partial with the logic for choosing params. We use inline partials a good bit, they look something like this

```hbs
{{#if some condition}}
  {{> inline-partial-name someParam=1 }}
{{else if something else}}
  {{> inline-partial-name someParam=2 }}
{{/if}}
{{#*inline 'inline-partial-name'}}
// blah
{{/inline}}
```

nmanu1: That makes sense. I'll change it to use an inline partial

cards/faq-accordion/component.js
```diff
@@ -68,11 +69,28 @@ class faq_accordionCardComponent extends BaseCard['faq-accordion'] {
 
     const contentEl = this._container.querySelector(accordionContentSelector);
     let isExpanded = this._container.querySelector(`.${accordionExpandedClass}`);
-    contentEl.style.height = `${isExpanded ? contentEl.scrollHeight : 0}px`;
+    let stayExpanded = this._container.querySelector(`.${stayExpandedClass}`);
```
oshi97, line 72: can we use const instead of let here and below

cards/multilang-faq-accordion/component.js
```diff
@@ -68,11 +69,28 @@ class multilang_faq_accordionCardComponent extends BaseCard['multilang-faq-accor
...
+      isExpanded = true;
+      cardEl.classList.toggle(accordionExpandedClass, isExpanded);
```
oshi97, line 79: does this work in ie11? you can test ie11 using browserstack (you can find a link to browserstack in your okta extension)

nmanu1: I assume it should, since it is the same as in line 97 below. But I can test if there might still be issues

yen-tt: it does not suppose second parameter for the toggle method for ie11 (https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle)

nmanu1: Fixed

cards/faq-accordion/component.js
```diff
@@ -76,18 +76,24 @@ class faq_accordionCardComponent extends BaseCard['faq-accordion'] {
...
     if (stayExpanded && this.getState('feedbackSubmitted')) {
       isExpanded = true;
-      cardEl.classList.toggle(accordionExpandedClass, isExpanded);
+      if (!cardEl.classList.contains(accordionExpandedClass)) {
+        cardEl.classList.toggle(accordionExpandedClass, isExpanded);
+      }
       accordionToggleEl.setAttribute('aria-expanded', isExpanded ? 'true' : 'false');
     }
-    cardEl.classList.toggle(stayExpandedClass, false);
+    if (cardEl.classList.contains(stayExpandedClass)) {
```
yen-tt, line 84: this always get remove for the next re-render right? I think you can just do `cardEl.classList.remove(stayExpandedClass); ` without if statement

cards/faq-accordion/component.js
```diff
@@ -76,18 +76,24 @@ class faq_accordionCardComponent extends BaseCard['faq-accordion'] {
 
     if (stayExpanded && this.getState('feedbackSubmitted')) {
       isExpanded = true;
-      cardEl.classList.toggle(accordionExpandedClass, isExpanded);
+      if (!cardEl.classList.contains(accordionExpandedClass)) {
+        cardEl.classList.toggle(accordionExpandedClass, isExpanded);
```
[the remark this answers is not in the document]
nmanu1: I'll change it to use that

cards/faq-accordion/component.js
```diff
@@ -76,18 +76,24 @@ class faq_accordionCardComponent extends BaseCard['faq-accordion'] {
...
+    }
     contentEl.style.height = `${isExpanded ? contentEl.scrollHeight : 0}px`;
     this._setLinksInteractivity(linkEls, isExpanded);
 
     const thumbSelectorEls = this._container.querySelectorAll('.js-HitchhikerCard-thumbInput');
     if (thumbSelectorEls) {
       thumbSelectorEls.forEach(el => {
         el.addEventListener('click', (e) => {
-          cardEl.classList.toggle(stayExpandedClass, true);
+          if (!cardEl.classList.contains(stayExpandedClass)) {
```
yen-tt, line 94: similarly, this always get added when clicking the feedback button, so I think just having `cardEl.classList.add(stayExpandedClass);` without if statement should be good!

cards/multilang-faq-accordion/template.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="HitchhikerFaqAccordion yxt-{{cardName}} js-HitchhikerFaqAccordion{{#if card.isExpanded}} HitchhikerFaqAccordion--expanded{{/if}}"
+<div class="HitchhikerFaqAccordion yxt-{{cardName}} js-HitchhikerFaqAccordion{{#if card.isExpanded}} HitchhikerFaqAccordion--expanded{{/if}}  js-HitchhikerFaqAccordion{{#if card.isExpanded}} HitchhikerFaqAccordion--expanded{{/if}}  HitchhikerFaqAccordion--stayExpanded"
```
yen-tt, line 1: do we need `HitchhikerFaqAccordion--stayExpanded` to be in the template? in the case that card.isExpanded is false, we don't want that class right? this seems like something that should be added in JS code when you hit the feedback button

static/scss/answers/module.scss
```diff
@@ -31,6 +31,53 @@
...
+  {
+    display: flex;
+    align-items: flex-end;
+
+    @include bpgte(sm)
+    {
+      flex-direction: column;
+    }
+  }
+
```
yen-tt, line 44: can we double check if the styling is good on mobile and small/wide screen browser page?

cards/multilang-faq-accordion/template.hbs
nmanu1: Yes, it can be removed

nmanu1: Actually, it can't be removed because the component is mounted again when feedback is submitted. Instead, I can remove the code adding the class on a feedback button click

static/scss/answers/module.scss
nmanu1: yes, it looks good in all 3 cases

yen-tt (approved):
lgtm!

PR #941, opened 2021-09-08
uniform product image size

yen-tt:
- update css for product-prominentimage-clickable and product-prominentimage cards to have fixed height of 300px that fill the whole width (maintain its aspect ratio, clipped to fit)
- the height is a customizable css variable
- since object-fit doesn't work in ie11, a polyfill library (object-fit-images) is used to adjust the image.

J=SLAP-1553
TEST=manual

- add product entities with different image sizes, launched test site and see that all images are properly sized. tested in IE11, chome, and safari

### DIFF
--- a/cards/card_component.js
+++ b/cards/card_component.js
@@ -17,10 +17,8 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
    */
   onMount() {
     //polyfill for image styling (object-fit) in ie11
-    if (['product-prominentimage', 'product-prominentimage-clickable'].includes('{{componentName}}')) {
-      var images = document.querySelectorAll('.HitchhikerProductProminentImage-img');
-      HitchhikerJS.objectFitImages(images);
-    }
+    var images = document.querySelectorAll('.HitchhikerProductProminentImage-img');
+    HitchhikerJS.objectFitImages(images);
 
     const showExcessDetailsToggleEls = this._container.querySelectorAll('.js-HitchhikerCard-detailsToggle');
     const excessDetailsEls = this._container.querySelectorAll('.js-HitchhikerCard-detailsText');
--- a/cards/card_component.js
+++ b/cards/card_component.js
@@ -17,7 +17,7 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
    */
   onMount() {
     //polyfill for image styling (object-fit) in ie11
-    var images = document.querySelectorAll('.HitchhikerProductProminentImage-img');
+    const images = document.querySelectorAll('.HitchhikerProductProminentImage-img');
     HitchhikerJS.objectFitImages(images);
 
     const showExcessDetailsToggleEls = this._container.querySelectorAll('.js-HitchhikerCard-detailsToggle');
--- a/cards/card_component.js
+++ b/cards/card_component.js
@@ -16,6 +16,12 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
    * only if such a toggle is present in the DOM.
    */
   onMount() {
+    //polyfill for image styling (object-fit) in ie11
+    if (['product-prominentimage', 'product-prominentimage-clickable'].includes('{{componentName}}')) {
+      var images = document.querySelectorAll('.HitchhikerProductProminentImage-img');
+      HitchhikerJS.objectFitImages(images);
+    }
+
     const showExcessDetailsToggleEls = this._container.querySelectorAll('.js-HitchhikerCard-detailsToggle');
     const excessDetailsEls = this._container.querySelectorAll('.js-HitchhikerCard-detailsText');
     if (showExcessDetailsToggleEls && excessDetailsEls) {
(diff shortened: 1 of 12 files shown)


static/scss/answers/cards/product-prominentimage.scss
```diff
@@ -55,14 +55,16 @@
...
     width: 100%;
     flex-shrink: 0;
   }
 
   &-img
   {
     width: 100%;
-    height: auto;
+    height: 300px;
+    object-fit: cover;
```
oshi97, line 67: does this work on ie11? I had issues getting object-fit to work on it before

yen-tt: you are right, it didn't work. I updated to use a div with background image instead of img tag, so we can use background-position and background-size similiar to object-position and object-fit. this should work in all browser

static/scss/answers/cards/product-prominentimage-clickable.scss
```diff
@@ -77,15 +77,20 @@
...
+    justify-content: center;
     width: 100%;
     flex-shrink: 0;
   }
 
   &-img
   {
     width: 100%;
-    height: auto;
+    height: 300px;
```
oshi97, line 90: just checking that there's no drawbacks to having a fixed height that Rose might care about?

yen-tt: For uniform size, I think a fixed height have to be implemented. This was used in the HH link from the jira item as well. But it seems like Rose is reconsidering this change and leaving the images as what we have before.

oshi97 (approved):
if Rose doesn't double back this lgtm!

cards/product-prominentimage/component.js
```diff
@@ -20,6 +20,10 @@ class product_prominentimageCardComponent extends BaseCard['product-prominentima
     }
     const linkTarget = AnswersExperience.runtimeConfig.get('linkTarget') || '_top';
 
+    //polyfill styling for ie11
+    var someImages = document.querySelectorAll('.HitchhikerProductProminentImage-img');
+    HitchhikerJS.objectFitImages(someImages);
```
oshi97, line 25: would it be possible to do the global polyfill for this, and take this out of the specific card components? We probably don't want to run this every render, and this isn't really the responsibility of the card anyway. Maybe we could do it only the legacy bundle

yen-tt: I will try, does ie11 only use the legacy bundles?

oshi97: yup!

yen-tt: So this can't be in entry-legacy.js because this script is invoked before the html pages are loaded. The polyfill function would requires the src link populated in the img tag then convert to some inline css styling in the tag.

oshi97: Ah ok got it, maybe we can shuffle this off to the base card component to hide some implementation? like maybe we can add a flag to the child card's constructor (or even have a list of cards in the base card component?)
taking it out of the child cards, in whatever way we decide, will also make it easier to tweak forked versions of these cards in the future.

oshi97: Does this polyfill work in a way where it doesn't do anything if you run it on a browser that *does* support object-fit? If so I don't mind running this a few extra times on ie11, in order to improve our non ie-11 cases and keep the cards a little cleaner

yen-tt: the polyfill function would check if the element.style contains object-fit. If so, it wouldn't do anything

yen-tt: I moved the implementations into the base card component, which check the list of cards that would require this polyfill. Let me know if the approach is good!

cards/card_component.js
```diff
@@ -16,6 +16,12 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
    * only if such a toggle is present in the DOM.
    */
   onMount() {
+    //polyfill for image styling (object-fit) in ie11
+    if (['product-prominentimage', 'product-prominentimage-clickable'].includes('{{componentName}}')) {
```
oshi97, line 20: is this going to be an issue if somebody forks the card with a custom name? could we just run the polyfill on the entire card all the time?

yen-tt: you are right. updated!

cards/card_component.js
```diff
@@ -16,6 +16,10 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
    * only if such a toggle is present in the DOM.
    */
   onMount() {
+    //polyfill for image styling (object-fit) in ie11
+    var images = document.querySelectorAll('.HitchhikerProductProminentImage-img');
```
oshi97, line 20: I hate to be that guy can we use const

cards/card_component.js
```diff
@@ -16,6 +16,10 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
    * only if such a toggle is present in the DOM.
    */
   onMount() {
+    //polyfill for image styling (object-fit) in ie11
```
oshi97, line 19: I was thinking it might be worth it to run it on the entire card instead of just this css class, on the off chance somebody renames their card css classes, but this is probably fine for now!

oshi97 (approved):
lgtm after changing to const!

cards/card_component.js
yen-tt: ah nah my bad, missed this one

PR #943, opened 2021-09-08
currency pricerange formatter

yen-tt:
- Added Formatter.priceRange() function that utilize `locale-currency` and `currency-symbol-map` npm library to retrieve currency symbol based on location's country code (ISO format), and construct the price range accordingly.
- exclude `currency-symbol-map` from the node_modules exclusion in webpack babel for production build so it can be transpile to es5

J=SLAP-1552
TEST=manual&auto

- added jest test for Formatter.priceRange
- add priceRange for a UK location entity. Launched test site and see that it's displayed in £ instead of default $


static/js/formatters-internal.js
```diff
@@ -522,6 +523,28 @@ export function price(fieldValue = {}, locale) {
   return price.toLocaleString(localeForFormatting, { style: 'currency', currency: currencyCode });
 }
 
+/**
+ * Returns a localized price range string for the given price range ($-$$$$) and country code (ISO format)
+ * @param {Object} pricerange The price range from LiveAPI entity
+ * @param {string} countrycode The country code from LiveAPI entity (e.g. profile.address.countryCode)
+ * @return {string} The price range with correct currency symbol formatting according to country code
+ */
+export function priceRange(priceRange, countryCode) {
```
tmeyer2115, line 532: Minor nit, but I find it confusing to have a function called `priceRange` that takes in a `priceRange`. How about naming this something like `localizePriceRange`?

static/js/formatters-internal.js
```diff
@@ -522,6 +523,28 @@ export function price(fieldValue = {}, locale) {
...
+  }
+  const currencyCode = LocaleCurrency.getCurrency(countryCode);
+  if (currencyCode) {
+    const currencySymbol = getSymbolFromCurrency(currencyCode);
+    if (currencySymbol) {
+      return Array(priceRange.length+1).join(currencySymbol); 
+    }
+  }
+  console.warn(`Unable to determine currency symbol from ISO country code ${countryCode}.`);
+  return '';
```
tmeyer2115, line 545: In this case, would it be better to simply return the original price range, non-localized?

static/js/formatters-internal.js
```diff
@@ -522,6 +523,28 @@ export function price(fieldValue = {}, locale) {
...
+export function priceRange(priceRange, countryCode) {
+  if (!priceRange || !countryCode) {
+    console.warn(`No price range or country code given.`);
+    return '';
+  }
+  const currencyCode = LocaleCurrency.getCurrency(countryCode);
+  if (currencyCode) {
+    const currencySymbol = getSymbolFromCurrency(currencyCode);
+    if (currencySymbol) {
+      return Array(priceRange.length+1).join(currencySymbol);
```
tmeyer2115, line 541: For my own edification, what does the `priceRange` object look like?

static/js/formatters-internal.js
yen-tt: I think keeping the function name as priceRange would be more consistent with the other formatter functions like Formatter.price(). I can rename priceRange param to `defaultPriceRange` or `entityPriceRange`?

static/js/formatters-internal.js
yen-tt: the given priceRange from entity, if it's defined, is like this: '$', '$$', '$$$', or '$$$$'.

static/js/formatters-internal.js
yen-tt: I thought it might be a bit misleading to have default dollar signs on international locations if we can't localized it so I left it blank. But I can return the original price range if you think that's fine!

tmeyer2115: My first inclination is that we return the original price range. That way, there's at least something displayed. The formatter is then strictly better than just plopping the original price range on the page. Let's go with that approach for now and see if we get QA feedback.

static/js/formatters-internal.js
tmeyer2115: So it's a string? The param type listed in the method signature is object.

static/js/formatters-internal.js
tmeyer2115: Sure, `defaultPriceRange` or `priceRangeFieldValue` sgtm!

static/js/formatters-internal.js
yen-tt: got it. Updated!

static/js/formatters-internal.js
yen-tt: oops, I missed that. Updated!

static/js/formatters-internal.js
```diff
@@ -525,24 +525,24 @@ export function price(fieldValue = {}, locale) {
...
+  if (!defaultPriceRange || !countryCode) {
     console.warn(`No price range or country code given.`);
     return '';
   }
   const currencyCode = LocaleCurrency.getCurrency(countryCode);
   if (currencyCode) {
     const currencySymbol = getSymbolFromCurrency(currencyCode);
     if (currencySymbol) {
-      return Array(priceRange.length+1).join(currencySymbol); 
+      return Array(defaultPriceRange.length+1).join(currencySymbol);
```
tmeyer2115, line 541: Could we just do defaultPriceRange.replaceAll('$", currencySymbol) instead? Seems a bit simpler.

yen-tt: hm you are right. I guess the backend will only have $ for now. I will update this!

tmeyer2115 (approved):

PR #946, opened 2021-09-10
Formatter for printing categories

yen-tt:
- add a formatter getCategoryNames that will return a list of category names based on the list of category ids and HH supplied map of id to display name

J=SLAP-823
TEST=auto

add jest test and see that it passed sucessfully

### DIFF
--- a/static/js/formatters-internal.js
+++ b/static/js/formatters-internal.js
@@ -568,23 +568,22 @@ export function getYoutubeUrl(videos = []) {
 }
 
 /**
- * construct a string of displayable category names based on given category ids from liveAPI
+ * construct a list of displayable category names based on given category ids from liveAPI
  * and a mapping of category ids to names.
  * 
  * @param {string[]} categoryIds category ids from liveAPI
  * @param {Object[]} categoryMap mapping of category ids to names
  * @param {string} categoryMap[].id id of a category entry
  * @param {string} categoryMap[].category name of a category entry
- * @returns {string} category names separated by comma
+ * @returns {string[]} a list of category names
  */
  export function getCategoryNames(categoryIds, categoryMap) {
   if (!categoryIds || !categoryMap) {
-    return '';
+    return [];
   }
-  const categoryNames = categoryIds.reduce((list, id) => {
+  return categoryIds.reduce((list, id) => {
     const categoryEntry = categoryMap.find(category => category.id === id);
     categoryEntry ? list.push(categoryEntry.category) : console.error(`Unable to find category name for id ${id}.`);
     return list;
   }, []);
-  return categoryNames.join(', ');
 }
--- a/static/js/formatters-internal.js
+++ b/static/js/formatters-internal.js
@@ -566,3 +566,24 @@ export function getYoutubeUrl(videos = []) {
     : null;
   return youtubeVideoUrl;
 }
+
+/**
+ * construct an array of displayable category names based on given category ids from liveAPI
+ * and a mapping of category ids to names.
+ * 
+ * @param {string[]} categoryIds category ids from liveAPI
+ * @param {Object[]} categoryMap mapping of category ids to names
+ * @param {string} categoryMap[].id id of a category entry
+ * @param {string} categoryMap[].category name of a category entry
+ * @returns {string[]} an array of category names based on given category ids
+ */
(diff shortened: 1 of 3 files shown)


static/js/formatters-internal.js
```diff
@@ -566,3 +566,24 @@ export function getYoutubeUrl(videos = []) {
     : null;
   return youtubeVideoUrl;
 }
+
+/**
+ * construct an array of displayable category names based on given category ids from liveAPI
+ * and a mapping of category ids to names.
+ * 
+ * @param {string[]} categoryIds category ids from liveAPI
+ * @param {Object[]} categoryMap mapping of category ids to names
```
tmeyer2115, line 575: I think we want to remove the square brackets here, it gives the appearance of an Array of Objects.

tmeyer2115 (commented):
I think it could be worth asking Rose if this formatter should output a list of strings, as they outline in the DocMD, or if it should output a single string. In the card, they composed this formatter with the `joinList` one. I'm wondering if they will want that to happen automatically inside this `Formatter` or if there's another use-case for a list of category names, unconcatted.

tests/static/js/formatters.js
```diff
@@ -153,4 +153,43 @@ describe('Formatters', () => {
...
+      const categoryIds = ['5', '0'];
+      const categoryNames = Formatters.getCategoryNames(categoryIds, categoryMap);
+      expect(categoryNames).toEqual([]);
+    });
+    it('return a list of matching category names', () => {
+      const categoryIds = ['1', '3'];
+      const categoryNames = Formatters.getCategoryNames(categoryIds, categoryMap);
+      expect(categoryNames).toEqual(['Neurology', 'Psychiatry']);
+    });
+  });
```
nmanu1, line 194: do we want to add a test for a mix of matching and not matching category names?

yen-tt: sure, updated!

yen-tt:
> I think it could be worth asking Rose if this formatter should output a list of strings, as they outline in the DocMD, or if it should output a single string. In the card, they composed this formatter with the `joinList` one. I'm wondering if they will want that to happen automatically inside this `Formatter`. If there's another use-case for a list of category names, unconcatted, then we would not want to add that additional step in this new formatter.

Checked in with Rose, updated to return a single string!

static/js/formatters-internal.js
tmeyer2115: Got it!

static/js/formatters-internal.js
```diff
@@ -568,22 +568,23 @@ export function getYoutubeUrl(videos = []) {
...
  */
  export function getCategoryNames(categoryIds, categoryMap) {
   if (!categoryIds || !categoryMap) {
-    return [];
+    return '';
   }
-  return categoryIds.reduce((list, id) => {
+  const categoryNames = categoryIds.reduce((list, id) => {
     const categoryEntry = categoryMap.find(category => category.id === id);
     categoryEntry && list.push(categoryEntry.category);
```
tmeyer2115, line 586: Maybe if there's not a corresponding category entry, we display a `console.error`?

yen-tt: Updated!

static/js/formatters-internal.js
```diff
@@ -566,3 +566,25 @@ export function getYoutubeUrl(videos = []) {
...
+ export function getCategoryNames(categoryIds, categoryMap) {
+  if (!categoryIds || !categoryMap) {
+    return '';
+  }
+  const categoryNames = categoryIds.reduce((list, id) => {
+    const categoryEntry = categoryMap.find(category => category.id === id);
+    categoryEntry ? list.push(categoryEntry.category) : console.error(`Unable to find category name for id ${id}.`);
+    return list;
+  }, []);
+  return categoryNames.join(', ');
```
oshi97, line 589: are they going to ask for the last element to be separated with an "and" instead of a comma?
like would they ever want
`"category 1, category2, and category3`
(I hope not because i18n would be annoying with this)

yen-tt: good point. Talked to Rose, I will revert this to return an array of category instead, and it will be up to the user to customize it with join list formatter or manipulate the list however they want

tmeyer2115 (approved):

PR #947, opened 2021-09-10
fix partial extra parenthesis

yen-tt:
check common-partial scripts, remove extra paranthesis from filterbox


templates/common-partials/script/filterbox.hbs
```diff
@@ -10,4 +10,4 @@ ANSWERS.addComponent("FilterBox",
   {{{json componentSettings.FilterBox}}}
 ]
 {{/shallowMergeConfig}}
-));
\ No newline at end of file
+);
```
oshi97, line 13: could we add a snapshot or something for this? maybe just add static filters to one of the vertical pages

oshi97 (approved):

PR #948, opened 2021-09-10
error with getUrlWithTextHighlight

yen-tt:
direct answer may return a snippet with no matchSubstring, add a check for that before constructing url with text highlighting

TEST=manual&auto
- test a query that return a Direct answer without any matched substrings, see that no error occur and url is still working
- add jest test

### DIFF
--- a/static/js/formatters-internal.js
+++ b/static/js/formatters-internal.js
@@ -599,7 +599,7 @@ export function getYoutubeUrl(videos = []) {
  * @returns a URL with text fragment URI component attached
  */
 export function getUrlWithTextHighlight(snippet, baseUrl) {
-  if (!isChrome() || snippet.matchedSubstrings.length === 0) {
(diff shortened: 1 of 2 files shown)


static/js/formatters-internal.js
```diff
@@ -599,7 +599,7 @@ export function getYoutubeUrl(videos = []) {
  * @returns a URL with text fragment URI component attached
  */
 export function getUrlWithTextHighlight(snippet, baseUrl) {
-  if (!isChrome()) {
+  if (!isChrome() || snippet.matchedSubstrings.length === 0) {
```
oshi97, line 602: do we also need to check whether snippet isn't undefined?

yen-tt: as of now with documentsearch-directanswers card, I assumed snippet is always defined if the card is used. But yeah that should be check in case this could be use else where. will update!

oshi97 (approved):

cea2aj (approved):

PR #949, opened 2021-09-13
QA item: use site locale for price range as default

yen-tt:
- update priceRange formatter function to determined the price range string in following order:
	- use provided country code to determine the currency symbol
	- if country code is invalid or undefined, use site locale to determine the currency symbol
	- if all else fails, return the default price range, which is in $

TEST=auto

add more cases in jest test, all passed

### DIFF
--- a/static/js/formatters-internal.js
+++ b/static/js/formatters-internal.js
@@ -537,15 +537,15 @@ export function priceRange(defaultPriceRange, countryCode) {
     console.warn(`No price range or country code given.`);
     return '';
   }
-  if(countryCode) {
+  if (countryCode) {
     const currencySymbol = getSymbolFromCurrency(LocaleCurrency.getCurrency(countryCode));
-    if(currencySymbol) {
+    if (currencySymbol) {
       return defaultPriceRange.replace(/\$/g, currencySymbol); 
     }
   }
   const locale = _getDocumentLocale();
   const currencySymbol = getSymbolFromCurrency(LocaleCurrency.getCurrency(locale));
-  if(currencySymbol) {
+  if (currencySymbol) {
     return defaultPriceRange.replace(/\$/g, currencySymbol); 
   }
   console.warn(`Unable to determine currency symbol from ISO country code "${countryCode}" or locale "${locale}".`);
--- a/static/js/formatters-internal.js
+++ b/static/js/formatters-internal.js
@@ -534,7 +534,7 @@ export function price(fieldValue = {}, locale) {
  */
 export function priceRange(defaultPriceRange, countryCode) {
   if (!defaultPriceRange) {
-    console.warn(`No price range or country code given.`);
+    console.warn(`Price range is not provided.`);
     return '';
   }
   if (countryCode) {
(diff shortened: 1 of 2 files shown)


static/js/formatters-internal.js
```diff
@@ -525,24 +525,30 @@ export function price(fieldValue = {}, locale) {
...
-  if (currencyCode) {
-    const currencySymbol = getSymbolFromCurrency(currencyCode);
+  if (countryCode) {
+    const currencySymbol = getSymbolFromCurrency(LocaleCurrency.getCurrency(countryCode));
     if (currencySymbol) {
       return defaultPriceRange.replace(/\$/g, currencySymbol); 
     }
   }
-  console.warn(`Unable to determine currency symbol from ISO country code ${countryCode}.`);
+  const locale = _getDocumentLocale();
```
oshi97, line 546: does _getDocumentLocale work properly for chinese? I was curious into what it was and it looks like it does some normalization that might not work with chinese?

static/js/formatters-internal.js
```diff
@@ -525,24 +525,30 @@ export function price(fieldValue = {}, locale) {
...
+ * If country code is invalid or undefined, use locale of the site to determine the currency symbol.
+ * If all else fails, use the default priceRange with dollar sign.
  * @param {string} defaultPriceRange The price range from LiveAPI entity
  * @param {string} countrycode The country code from LiveAPI entity (e.g. profile.address.countryCode)
  * @return {string} The price range with correct currency symbol formatting according to country code
  */
 export function priceRange(defaultPriceRange, countryCode) {
-  if (!defaultPriceRange || !countryCode) {
+  if (!defaultPriceRange) {
     console.warn(`No price range or country code given.`);
```
nmanu1, line 537: This no longer checks for country code, right? So, it should only warn that no price range is given

static/js/formatters-internal.js
yen-tt: ah good point! in this formatter, chinese locale would require some parsing to get the region for the library to work. Updated!

static/js/formatters-internal.js
```diff
@@ -525,24 +526,31 @@ export function price(fieldValue = {}, locale) {
...
-  if (currencyCode) {
-    const currencySymbol = getSymbolFromCurrency(currencyCode);
+  if (countryCode) {
+    const currencySymbol = getSymbolFromCurrency(LocaleCurrency.getCurrency(countryCode));
     if (currencySymbol) {
       return defaultPriceRange.replace(/\$/g, currencySymbol); 
     }
   }
-  console.warn(`Unable to determine currency symbol from ISO country code ${countryCode}.`);
+  const { region, language } = parseLocale(_getDocumentLocale());
```
oshi97, line 547: does getDocumentLocale need to he updated to use parseLocale?

yen-tt: I think it's fine to keep it as is since there are multiple locations that invoke this method to get the locale as a single string as is. If there's a place, like in priceRange, that requires some parsing then they can use parseLocale to get each sections.

PR #950, opened 2021-09-13
Move video to top of video cards

cea2aj:
Move the videos in the video cards to the top so that the videos are aligned

By moving the videos to the top of the card, we match the layout of the product image cards. The original video card spec has the video below the title, however this causes alignment issues when the card title overflows onto the next line since the videos are now out of alignment. CSS Subgrid is the proper way to align content across cards, (https://www.smashingmagazine.com/2018/07/css-grid-2/) however currently it's only supported on firefox so we can't go that route yet. For now, we simply move the video to the top of the cards.

J=none
TEST=manual

Test the card locally and confirm the layout looks the same as the product image cards. Add CTAs and subtitles locally and confirm they look good as well

### DIFF
--- a/static/scss/answers/cards/product-prominentvideo.scss
+++ b/static/scss/answers/cards/product-prominentvideo.scss
@@ -8,11 +8,15 @@
   width: 100%;
   height: 100%;
 
+  &-body
+  {
+    padding: calc(var(--yxt-base-spacing) * 3/4) var(--yxt-base-spacing);
+  }
+
   &-header
   {
     display: flex;
     flex-direction: column;
-    padding: calc(var(--yxt-base-spacing) * 3/4) var(--yxt-base-spacing);
   }
 
   &-contentWrapper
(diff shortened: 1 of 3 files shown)


static/scss/answers/cards/product-prominentvideo.scss
```diff
@@ -8,19 +8,22 @@
   width: 100%;
   height: 100%;
 
+  &-body
+  {
+    padding: calc(var(--yxt-base-spacing) * 3/4) var(--yxt-base-spacing);
+  }
+
   &-header
```
yen-tt, line 16: we may not need this anymore if that div with classname `HitchhikerProductProminentVideo-header` is replaced to `-body` instead

cea2aj: you're right! i'll remove

cea2aj: updated!

yen-tt (approved):

PR #954, opened 2021-09-15
avoid results render on locator map view

yen-tt:
Improve performance for map pages on mobile by only render result cards on list view

- VerticalFullPageMapOrchestrator use a field in core storage `DISABLE_RENDER_RESULTS` to communicate with VerticalResults component whether it should render the results on page.
- VerticalResults component's config pass in a beforeMountOverride function to avoid mount() being call in the component lifecycle and render if `DISABLE_RENDER_RESULTS` is true.
- VerticalResults component's config also pass in a onCreate function to add a listener on any changes `DISABLE_RENDER_RESULTS`, to potentially trigger render update for the rsults if user switch from map view to list view, or screen resize between mobile and desktop view
- Since VerticalResults now will not render any location cards on map view, VerticalFullPageOrchestrator can no longer copy the card and display that when a pin is clicked. So,cardType is passed into the config for VerticalFullPageOrchestrator. When a pin is click, the cardId is used to pull the corresponding entity data from results in storage, and construct a card component based on cardType to append to the top level map container.

Note: require update to SDK's component.js file for beforeMountOverride to work

J=SLAP-1302
TEST=manual & auto

smoke test from theme's locations_full_page_map vertical page:
- resize screen from mobile to desktop and vice versa, see that results and focused pins are still loaded properly.
- toggle between map and list view, see that the results are only render on list view in elemnt inspection (except for the first time when search is trigger from searchbar)
- in map view, drag across the map a few times and see that pins are loaded correctly but results are not render on the page.
- see that detail card is displayed correctly when click on pin from map view, before and after dragging acrross the map

recorded performance metric on browser inspector, see that the tasks to re-render the components shrink to around ~≤27ms compare to previous time depending on result count (~≥100ms for 2000+ results) (got rid of most of the 'long tasks' - where the main UI thread is busy for 50 ms or longer)

no changes in map related snapshots from percy

### DIFF
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -189,14 +189,14 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
       this.searchThisArea();
     });
 
-    this.setupCssForBreakpoints();
+    this.setupMobileBreakpointListener();
     this.addMapComponent();
   }
 
   /**
    * Properly set CSS classes for mobile and desktop
    */
-  setupCssForBreakpoints () {
+   setupMobileBreakpointListener () {
     if (!this.isMobile()) {
       this.updateCssForDesktop();
     }
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -20,6 +20,13 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
     super(config, systemConfig);
 
     /**
+     * Name of a location card type
+     * 
+     * @type {string}
+     */
+    this.cardType = config.cardType;
+
+    /**
      * The container in the DOM for the interactive map
      * @type {HTMLElement}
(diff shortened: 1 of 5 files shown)


static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -198,6 +205,7 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
       if (this.isMobile()) {
         this.updateCssForMobile();
       } else {
+        this.core.storage.set('DISABLE_RENDER_RESULTS', false);
```
cea2aj, line 208: since the function `setupCssForBreakpoints` no longer just updates the CSS, what are your thoughts on renaming it to `setupMobileBreakpointListener`?

cea2aj:
Great job getting rid of those long tasks! That's awesome

static/js/theme-map/VerticalFullPageMapOrchestrator.js
yen-tt: sounds good!

cea2aj (approved):

PR #957, opened 2021-09-21
Distance formatter to include locale with region

yen-tt:
Int.FormatNumber and toLocaleString would fail on locale with underscore, such as chinese locale like 'zh-Hant_CN'. Update function to replace underscore with dash for locale

TEST=auto

pass updated jest test

### DIFF
--- a/static/js/formatters-internal.js
+++ b/static/js/formatters-internal.js
@@ -82,15 +82,21 @@ export function toLocalizedDistance(profile, key = 'd_distance', displayUnits) {
   return this.toMiles(profile, undefined, undefined, locale);
 }
 
+
+
+export function _getJSFormattedLocale(locale) {
+  return locale && locale.replace(/_/g, '-');
+}
+
 export function _getDocumentLocale() {
-  return document.documentElement.lang.replace(/_/g, '-');
+  return _getJSFormattedLocale(document.documentElement.lang);
 }
 
 export function toKilometers(profile, key = 'd_distance', displayUnits = 'km', locale) {
   if (!profile[key]) {
     return '';
   }
-  locale = (locale && locale.replace(/_/g, '-')) || _getDocumentLocale()
+  locale = _getJSFormattedLocale(locale) || _getDocumentLocale();
   const distanceInKilometers = profile[key] / 1000; // Convert meters to kilometers
   return new Intl.NumberFormat(locale,
     { style: 'decimal', maximumFractionDigits: 1, minimumFractionDigits: 1})
@@ -101,7 +107,7 @@ export function toMiles(profile, key = 'd_distance', displayUnits = 'mi', locale
   if (!profile[key]) {
     return '';
   }
-  locale = (locale && locale.replace(/_/g, '-')) || _getDocumentLocale()
+  locale = _getJSFormattedLocale(locale) || _getDocumentLocale();
   const distanceInMiles = profile[key] / 1609.344; // Convert meters to miles
   return new Intl.NumberFormat(locale,
     { style: 'decimal', maximumFractionDigits: 1, minimumFractionDigits: 1 })
@@ -234,7 +240,7 @@ export function snakeToTitle(snake) {
  * @returns {string} The pretty printed value.
  */
 export function prettyPrintObject(obj, locale) {
-  locale = locale || _getDocumentLocale();
+  locale = _getJSFormattedLocale(locale) || _getDocumentLocale();
 
   switch (typeof obj) {
     case 'string':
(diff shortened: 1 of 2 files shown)


static/js/formatters-internal.js
```diff
@@ -101,7 +101,7 @@ export function toMiles(profile, key = 'd_distance', displayUnits = 'mi', locale
   if (!profile[key]) {
     return '';
   }
-  locale = locale || _getDocumentLocale()
+  locale = (locale && locale.replace(/_/g, '-')) || _getDocumentLocale()
```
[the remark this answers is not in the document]
yen-tt: updated to use `_getJSFormattedLocale` function

static/js/formatters-internal.js
```diff
@@ -82,15 +82,23 @@ export function toLocalizedDistance(profile, key = 'd_distance', displayUnits) {
...
+  return _getJSFormattedLocale(locale) || _getDocumentLocale() || 'en';
+}
+
+export function _getJSFormattedLocale(locale) {
+  return locale && locale.replace(/_/g, '-');
+}
+
 export function _getDocumentLocale() {
-  return document.documentElement.lang.replace('_', '-');
+  return _getJSFormattedLocale(document.documentElement.lang);
```
oshi97, line 94: can we just update this to use parseLocale?

yen-tt: parseLocale provide an object of different section of the locale, and we would have to piece them together with dash afterward, which would give us the same result. Is there another reason why you think we should use parseLocale here?

oshi97: For some reason I thought parseLocale would canonicalize stuff for us. I'm still a little confused on the difference between the 3 different methods we have here though

tests/static/js/formatters.js
```diff
@@ -62,10 +71,13 @@ describe('Formatters', () => {
       const price = Formatters.price(priceField, 'en');
       expect(price).toEqual('$100.99');
     });
+
     it('Formats a price in USD with no provided locale', () => {
+      document.documentElement.lang = 'en';
```
oshi97, line 76: does the documentElement.lang state need to be cleared after each test?

yen-tt: document.documentElement.lang takes after the value set to it in the previous 'it' function. It doesn't affect our tests with the way we set them up now, but I will clear it out to avoid confusion if we do update these tests later

static/js/formatters-internal.js
yen-tt: `_getJSFormattedLocale` is just to replace underscore with dashes in a **given locale string** as of now because JS functions like toLocaleString and Int.NumberFormat would throw an error otherwise. `_getDocumentLocale()` is to provide a JS formatted locale from the **document**. In a lot of places we would go through the logic of using the given locale if that exist or use the document locale or fallback to en as last resort, so I added `_getDefinedLocale` to avoid duplicating that code, which return a defined locale in JS format.

oshi97: could we just use the same locale everywhere? it feels like a potential source of confusion where some formatters use different flavors of the same locale. Is it necessary that some formatters directly use the documents locale and others do some preprocessing first?

oshi97: like it feels like this introduces a lot of entropy that we might not need

cea2aj: getDefinedLocale makes it more difficult for me to reason about where the locale came from and what format it is. Another approach would be to create a helper such as `formatLocaleForIntl` which replaces the underscores with dashes so that it doesn't break when passed to Intl. The line to get the locale would then be `locale = formatLocaleForIntl(locale || _getDocumentLocale())`

cea2aj: Another change we could make which isn't strictly necessary is to also return a full canonicalized locale from parseLocale

tests/static/js/formatters.js
```diff
@@ -5,19 +5,25 @@ describe('Formatters', () => {
   describe('toLocalizedDistance', () => {
     const profile = { d_distance: 10000000 }; // Distance in meters
     const distanceKey = 'd_distance';
-    document.documentElement.lang = 'en';
+    
+    beforeEach(function() {
```
cea2aj, line 9: nit: I'm used to seeing arrow functions here, but this works

static/js/formatters-internal.js
cea2aj: Maybe renaming '_getJSFormattedLocale' to '_getLocaleWithDashes' will be clearer? Otherwise, PR looks good

oshi97 (approved):

cea2aj (approved):

PR #962, opened 2021-09-30
add apiKey validation when using injected data and useJWT=false

oshi97:
J=TECHOPS-2183
TEST=manual

tested using my local JAMBO_INJECTED DATA
tested that errors are emitted when using the deprecated injected api key, but the build continues
tested that the build throws an error when no production or deprecated api key exist


hooks/templatedatavalidator.js
```diff
@@ -25,13 +26,50 @@ module.exports = function (pageData, partials) {
...
+  const { experienceKey } = globalConfig;
+  if (!experienceKey) {
     error('Missing Info: no experienceKey found.');
     return false;
   }
+  if (globalConfig.useJWT || globalConfig.apiKey) {
+    return true;
+  }
+
+  const injectedDataForExperience = JAMBO_INJECTED_DATA.answers.experiences[experienceKey];
```
tmeyer2115, line 42: Could we use the `?` syntax here? It's possible something on Spruce's end could cause there to be no `experiences` under `answers`. If that were the case, the validator would throw an NPE.

hooks/templatedatavalidator.js
```diff
@@ -25,13 +26,50 @@ module.exports = function (pageData, partials) {
...
+  if (!productionApiKey) {
+    if (!deprecatedApiKey) {
+      error(`No injected production api key found for experience key: "${experienceKey}"`);
+    } else {
+      warn('No injected production api key found, using the default apiKey instead.');
+    }
+  }
+
+  const stagingApiKey = injectedDataForExperience.configByLabel.STAGING.apiKey;
+  if (!stagingApiKey) {
```
tmeyer2115, line 60: It'd be nice if we had access to `isStaging` here, so we knew which of these blocks to care about. I'm wondering if we need something like that here. What about the case where isStaging is true and we can't find an `apiKey` for the `PRODUCTION` env. Does that really matter? We can still build the `STAGING` pages.

oshi97: isStaging is determined by the browser's URL so unfortunately we can't use it at build time

hooks/templatedatavalidator.js
tmeyer2115: I think blowing up is probably fine for now. The NPE will be easy to diagnose I expect.

hooks/templatedatavalidator.js
tmeyer2115: I'm asking Max if it's possible to ever have one version without the other. I wouldn't want to have an overly harsh validator.

tmeyer2115: Okay, sounds like you can never have one without the other! So we're good here.

tmeyer2115 (approved):

PR #966, opened 2021-10-07
Use color-text-primary for card body text

cea2aj:
Use the color-text-primary variable for the color of text on the body of cards

This change makes most text slightly lighter because the value of --yxt-color-text-primary is `#212121;`. Previously the cards were falling back to the default of the browser which is black.

J=SLAP-1617
TEST=manual

For testing purposes, set the text to red and view all of the cards of the test site. See that the color of the body of text is red as expected.

### DIFF
--- a/static/scss/answers/cards/document-standard.scss
+++ b/static/scss/answers/cards/document-standard.scss
@@ -64,6 +64,7 @@
   &-detailsText
   {
     @include rich_text_formatting;
+    color: var(--yxt-color-text-primary);
   }
 
   &-ctasWrapper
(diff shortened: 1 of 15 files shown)


static/scss/answers/cards/document-standard.scss
```diff
@@ -64,6 +64,7 @@
   &-detailsText
   {
     @include rich_text_formatting;
+    color: var(--yxt-color-text-primary);
```
yen-tt, line 67: is it possible to just define text color at top level for as `--yxt-color-text-primary` and let other element by overriding it with their own color like --yxt-color-text-primary? So this doesn't have to be included in every classname related to text

cea2aj: We could, however I discussed that with Tom and we decided to be more specific when placing the color because there is less risk of side effects

yen-tt: i see. sounds good!

yen-tt (approved):

PR #968, opened 2021-10-07
add sorting facets by optionsOrder

oshi97:
J=SLAP-1631
TEST=manual,auto

saw facets were sorted on the people page

### DIFF
--- a/static/js/transform-facets.js
+++ b/static/js/transform-facets.js
@@ -17,7 +17,7 @@ export default function transformFacets (facets, config) {
       return facet;
     }
 
-    if (typeof(config.fields[facet.fieldId]) !== 'object') {
+    if (typeof config.fields[facet.fieldId] !== 'object') {
       console.error(
         `The "fields" config for ${facet.fieldId} should be an object. ` +
         `Received ${config.fields[facet.fieldId]} instead.`);


static/js/transform-facets.js
```diff
@@ -12,33 +14,45 @@ export default function transformFacets (facets, config) {
...
+        options = options.sort((a, b) =>  b.displayName.toString().localeCompare(a.displayName.toString()));
+      } else {
+        console.error(`Unknown facet optionsOrder "${optionsOrder}" for the "${facet.fieldId}" facet.`);
       }
-      return filterOptions;
-    }, {});
+    }
+
+    // Don't expose Hitchhiker theme specific details to the SDK
+    const filterOptionsConfig = clonedeep(facetConfig);
```
yen-tt, line 48: Do we need to do a deep copy of it? I think we can use spread operator here like `const { fieldLabels, optionsOrder, ...filterOptionsConfig } = facetConfig;` and just pass in filterOptionsConfig that way

oshi97: ah, yeah I was just following the original code's logic, but that's better!

static/js/transform-facets.js
```diff
@@ -18,13 +16,15 @@ export default function transformFacets (facets, config) {
     if (!hasConfigurationForFacet) {
       return facet;
     }
-    const facetConfig = config.fields[facet.fieldId];
+    const {
+      fieldLabels,
+      optionsOrder,
+      ...filterOptionsConfig
+    } = config.fields[facet.fieldId];
```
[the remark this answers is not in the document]
oshi97: I can add an additional check to make sure it's not a falsy value

PR #969, opened 2021-10-08
RTL snapshots + parallel tests

yen-tt:
- update test site to generate pages with 'ar' as locale: Add index.ar.json, update locale config and create-vertical script
- update lockfile to v2
- add percy snapshots using pages with locale ar
- split snapshots in percy index.js script into smaller chunks so they can be run in parallel
- modify percy github action to conduct jobs that run percy snapshots in parallel to reduce wait time. all snapshot jobs' results will be condense to one percy build afterward. 

J=1483
TEST=auto

see that all percy snapshots completed succesfully

### DIFF
--- a/tests/percy/index.js
+++ b/tests/percy/index.js
@@ -21,7 +21,7 @@ async function iframeSnapshots(page) {
   await (new Photographer(iframePageNavigator, iframeCamera).captureSnapshots());
 }
 
-async function nonEnglishSnapshots(page) {
+async function spanishSnapshots(page) {
   const standardPageNavigator = new StandardPageNavigator(page, `http://localhost:${PORT}`);
   const standardCamera = new Camera(percySnapshot, page);
   standardPageNavigator.setCurrentLocale('es');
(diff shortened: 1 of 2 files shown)


tests/percy/index.js
```diff
@@ -9,30 +9,56 @@ const puppeteer = require('puppeteer');
...
+  });
+  server.start();
+  const browser = await puppeteer.launch();
+  const page = await browser.newPage();
+
+  const snapshotType = process.argv[2];
+  if (snapshotType === 'rtl') {
+    await RtlSnapshots(page);
+  } else if (snapshotType === 'non-english') {
+    await nonEnglishSnapshots(page);
```
oshi97, line 53: was a little confused for second since our RTL snapshots are arabic which is non english

tests/percy/index.js
```diff
@@ -9,30 +9,56 @@ const puppeteer = require('puppeteer');
...
+  }else if (snapshotType === 'iframe') {
```
oshi97, line 54: space

tests/percy/index.js
yen-tt: yea, updated

PR #971, opened 2021-10-11
handle different optionsFieldType

oshi97:
Originally, I thought running a toString() on the displayName would handle number display names.
But, string.prototype.localeCompare doesn't work with number strings the way I hoped it would.
If you try to sort an array of numbers like [100, 20, 120], the result will be [100, 120, 20] rather than
the expected [20, 100, 120].
My unit tests only had single digit numbers, so they didn't run test that case.

TEST=auto
J=SLAP-1631

### DIFF
--- a/static/js/transform-facets.js
+++ b/static/js/transform-facets.js
@@ -57,10 +57,10 @@ export default function transformFacets(facets, config) {
  * Sorts the facet options in place.
  * 
  * @param {{ displayName: string }[]} options The facet options to sort.
- * @param {'STRING' | 'INT'} optionsOrder 
- * @param {'ASC' | 'DESC'} optionsFieldType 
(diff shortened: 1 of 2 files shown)


static/js/transform-facets.js
```diff
@@ -41,23 +42,51 @@ export default function transformFacets (facets, config) {
...
 }
+
+/**
+ * Sorts the facet options in place.
+ * 
+ * @param {{ displayName: string }[]} options The facet options to sort.
+ * @param {'STRING' | 'INT'} optionsOrder 
+ * @param {'ASC' | 'DESC'} optionsFieldType 
+ * @param {string} fieldId 
+ * @returns
```
cea2aj, line 63: missing return type

cea2aj (approved):

PR #972, opened 2021-10-11
support sorting facets using optionsOrderList

oshi97:
I changed the sorting to not mutate the original array.
Originally I was worried about performance, but it should be very minor.

J=SLAP-1632
TEST=manual,auto

added optionsOrderList to the people page's config override for percy snapshots
manually built the page and saw the facets being sorted

### DIFF
--- a/static/js/transform-facets.js
+++ b/static/js/transform-facets.js
@@ -26,6 +26,7 @@ export default function transformFacets(facets, config) {
     const {
       fieldLabels,
       optionsOrder,
+      optionsOrderList,
       optionsFieldType = 'STRING',
       ...filterOptionsConfig
     } = config.fields[facet.fieldId];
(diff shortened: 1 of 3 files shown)


static/js/transform-facets.js
```diff
@@ -87,6 +90,26 @@ function sortFacetOptions(options, optionsOrder, optionsFieldType, fieldId) {
...
+ * Sorts the facet options using the priority specified in
+ * the optionsOrderList and returns a new array.
+ * 
+ * @param {{ displayName: string }[]} options The facet options to sort.
+ * @param {string[] | number[]} optionsOrderList
+ * @returns {{ displayName: string }[]}
+ */
+function sortFacetOptionsCustom(options, optionsOrderList) {
+  const getPriority = displayName => {
+    const index = optionsOrderList.indexOf(displayName);
```
tmeyer2115, line 108: If the optionsOrderList contains numbers, do we need to convert the `displayName` first?

oshi97: I talked with Rose and she discovered that numerical fields do not support facets. Currently, if you want "numerical facets" you would use a regular string field. So, long story short we're not going to support number facets because they don't exist in the backend.

PR #979, opened 2021-10-13
Version 1.25.0

cea2aj:
### Features
- Facets can be sorted by optionsOrder (#968) (#972)

### Enhancements
- Improved performance on the locator map view on mobile (#954)
- Improved build validation (#962)
- Vertical icons are commented-out by default (#967)

### Bug Fixes
- Fixed a bug with Direct Answer analytics (#978)
- Improve locale support with the distance formatter (#795)
- The primary text color variable now applies to the body of all cards (#966)

### DIFF
--- a/global_config.json
+++ b/global_config.json
@@ -1,5 +1,5 @@
 {
-  "sdkVersion": "release/v1.11", // The version of the Answers SDK to use
+  "sdkVersion": "1.11", // The version of the Answers SDK to use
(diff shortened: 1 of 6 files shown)


cea2aj:
> The google maps bug fix should be removed from the description notes, right?

good catch! updated

yen-tt:
should visitor support be in this release?

PR #983, opened 2021-10-14
Merge master (v1.25.0) into develop

cea2aj:
Merge and resolve conflicts.

Several reverts were added to the release branch but not the develop branch, and we need to revert those reverts. I opted to follow git's suggestion of not removing existing commits so that commits don't disappear from master in the next release.

### DIFF
--- a/directanswercards/card_component.js
+++ b/directanswercards/card_component.js
@@ -26,6 +26,7 @@ BaseDirectAnswerCard["{{componentName}}"] = class extends ANSWERS.Component {
 
     return super.setState({
       ...cardData,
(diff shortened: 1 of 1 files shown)


directanswercards/card_component.js
```diff
@@ -26,7 +26,6 @@ BaseDirectAnswerCard["{{componentName}}"] = class extends ANSWERS.Component {
 
     return super.setState({
       ...cardData,
-      feedbackEnabled: ANSWERS.getAnalyticsOptIn(),
```
nmanu1, line 29: This line should be added back in. I think it got removed because the more recent fix was reverted after the original fix was re-reverted.

oshi97 (approved):

PR #985, opened 2021-10-14
Use window.iframeLoaded promise

nmanu1:
Update iframe checks to use window.iframeLoaded promise. This is to avoid possible race conditions with the iframe (e.g. if the SDK adds the pagination component before iframe-resizer is spun up).

J=SLAP-1102
TEST=manual

Check default and iframe pages in test-site to make sure pagination scrolling works as expected for vertical map templates.

### DIFF
--- a/script/core.hbs
+++ b/script/core.hbs
@@ -38,12 +38,12 @@
       ...token && { apiKey: token },
       querySource: window.isOverlay ? 'OVERLAY' : 'STANDARD',
       onStateChange: (objParams, stringParams, replaceHistory) => {
-        if ('parentIFrame' in window) {
+        window.iframeLoaded.then(() => {
           parentIFrame.sendMessage(JSON.stringify({
             params: iframeGetSearchParams(),
             replaceHistory: replaceHistory
           }));
-        }
+        });
       },
       ...{{> script/additional-answers-config }},
       onReady: () => {
--- a/static/js/overlay/answers-frame/controllers/messagedirector.js
+++ b/static/js/overlay/answers-frame/controllers/messagedirector.js
@@ -73,6 +73,8 @@ export default class MessageDirector {
    * @param {IFrameMessage} message
    */
   _notifyParentFrame(message) {
-    window.parentIFrame && window.parentIFrame.sendMessage(message.toObject());
+    window.iframeLoaded.then(() => {
+      window.parentIFrame.sendMessage(message.toObject());
+    });
   }
 }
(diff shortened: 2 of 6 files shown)


templates/vertical-map/script/pagination.hbs
```diff
@@ -1,13 +1,14 @@
 ANSWERS.addComponent("Pagination", Object.assign({}, {
   container: "#js-answersPagination",
   onPaginate: (newPageNumber, oldPageNumber, totalPages) => {
-    if (window.parentIFrame) {
+    document.documentElement.scrollTop = 0;
+    document.body.scrollTop = 0; // Safari
```
yen-tt, line 5: to clarify, this should be applied to both default and iframe page?

templates/common-partials/script/pagination.hbs
```diff
@@ -1,13 +1,13 @@
 ANSWERS.addComponent("Pagination", Object.assign({}, {
   container: "#js-answersPagination",
   onPaginate: (newPageNumber, oldPageNumber, totalPages) => {
-    if (window.parentIFrame) {
+    document.documentElement.scrollTop = 0;
+    document.body.scrollTop = 0; // Safari
```
tmeyer2115, line 5: Does this scroll to the top of the iFrame or the top of the parent page?

templates/vertical-map/script/pagination.hbs
nmanu1: yes, this scrolls the page in the default case and scrolls the iFrame in an iframe page

script/core.hbs
```diff
@@ -38,12 +38,12 @@
       ...token && { apiKey: token },
       querySource: window.isOverlay ? 'OVERLAY' : 'STANDARD',
       onStateChange: (objParams, stringParams, replaceHistory) => {
-        if ('parentIFrame' in window) {
+        window.iframeLoaded.then(() => {
```
oshi97, line 41: is there some chance of this being a memory leak, if window.iframeLoaded is never fulfilled? (i.e. we're on a non-iframe integration)

oshi97: despite onStateChange running for literally every SDK storage update I would think the memory still wouldn't matter, but hopefully there would be a relatively easy fix?

oshi97: note: Connor thought of this first I just parroted it

static/js/overlay/answers-frame/controllers/messagedirector.js
```diff
@@ -73,6 +73,8 @@ export default class MessageDirector {
    * @param {IFrameMessage} message
    */
   _notifyParentFrame(message) {
-    window.parentIFrame && window.parentIFrame.sendMessage(message.toObject());
+    window.iframeLoaded.then(() => {
```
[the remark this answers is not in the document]
nmanu1: You're right. Looks like this is a different iframe, so I'll switch back to the previous check

PR #987, opened 2021-10-20
Automate WCAG testing

yen-tt:
- add github action to automate wcag test on pull requests to release, hotfix, support, and develop branches
- add axe-core/puppeteer to repo, and run analysis on all pages used in percy snapshot tests

J=1644
TEST=manual/auto

- ran `npm run wcag` with puppeteer's browser headless mode disabled, see that all pages are loaded and analyzed by axe-core, see that the violation results are console log out.
- see that github action `wcag test` ran in this pr

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
       "/test-utils/",
       "/test-site/",
       "/percy/",
+      "/wcag/",
       "/acceptance/"
     ]
   }


.github/workflows/wcag_test.yml
```diff
@@ -0,0 +1,24 @@
+name: WCAG tests
```
tmeyer2115, line 1: I think we should make an item to add the WCAG testing infra to the common repo.

tests/wcag/wcagreporter.js
```diff
@@ -0,0 +1,147 @@
...
+    await this._analyzeVerticalSearch();
+    await this._analyzeVerticalGridSearch();
+    await this._analyzeVerticalMapSearch();
+    await this._analyzeVerticalFullPageMapSearch();
+    await this._analyzeDirectAnswers();
+    return this.results;
+  }
+
+
+  async _analyzeUniversalSearch() {
```
tmeyer2115, line 27: This seems to mimic the exact flow of the `Photographer` class. I think we should do some re-factoring here. For both classes, we go to the same pages, run the same queries, etc. If we add a new page or want to run a new query, we will have to update both files.

tests/wcag/index.js
```diff
@@ -0,0 +1,66 @@
...
+async function wcagTester() {
+  const server = new HttpServer({
+    dir: 'test-site/public',
+    port: PORT
+  });
+  server.start();
+  const browser = await puppeteer.launch();
+  const page = await browser.newPage();
+  
+  const standardPageNavigator = new StandardPageNavigator(page, `http://localhost:${PORT}`);
```
tmeyer2115, line 35: There are WCAG issues specific to the iFrame integration itself. For instance, with WCAG it's important that the iFrame have a title attribute, etc. For the purposes of this item (SLAP-1644) I think using the `StandardPageNavigator` is fine. But, I will make a followup to ensure we're testing WCAG compliance of the iFrame integration as well.

yen-tt: I see, sounds good

tests/wcag/wcagreporter.js
cea2aj: Maybe we move towards a CAC approach? Something like:

```
interface TestCommand {
  type: 'click',
  params?: any[]
}

interface TestLocation {
  pageType: 'vertical' | 'universal',
  page?: string
  name: string
  viewPort?: {
    height?: number,
    width?: number
  }
  commands?: TestCommand[]
}

const testingLocations : TestLocation[] = [
  {
    name: 'universal-search',
    pageType: 'universal'
  },
  {
    name: 'vertical-grid-search',
    pageType: 'vertical',
    page: 'people'
  },
  {
    name: 'universal-search-faq-accordion',
    pageType: 'universal',
    commands: [
      { type: 'click', params: ['.HitchhikerFaqAccordion-toggle']}
    ]
  }
]
```

yen-tt: I see, that makes sense. Although we wouldn't have nice interface/typing since theme is only in js right now. Yeah I think  having a PageOperator that runs through these test configs would be good.

tmeyer2115: How would this `TestLocation[]` fit into what we have currently? Would the `Photographer` and `WCAGReporter` classes take them in? Then, for each entry in the array, the respective class would know what action to take? Another approach is to simply have a class that does all the navigating, clicking, etc. and takes the snapshot and does the WCAG analysis. For example: 

```
    await this._pageNavigator.gotoUniversalPage({ query: 'what if i forget my password?'});
    await this._pageNavigator.click('.HitchhikerFaqAccordion-toggle')
    await this._camera.snapshot('universal-search--faq-accordion');
    this.results.push(await this._analyzer.analyze());
```
This approach is fairly straightforward and we don't need to create a CaC framework.

yen-tt: I think the CaC approach is pretty neat, and it might also help us out with the multilangphotographer's tests as well. 
I imagine it's something like this:
```
while(pageOperator.hasNextTestLocation()) {
   const testName = pageOperator.nextTestLocation();
   this._camera.snapshot(testName); OR this.results.push(await this._analyzer.analyze());
}
```
I already have the pageOperator code that would parse the file and run the commands, but yes there would need to be more changes in other files and writing up the testLocations itself.
Perhaps, we can keep this PR as is for the format, and make a new item to address the CaC or page navigation refactoring in general?

PR #989, opened 2021-11-01
Update styling for items in More tab

nmanu1:
Update padding, line-height, and word break of navigation items in the 'More' tab. The padding should not apply to the first item in the list.

J=SLAP-1662
TEST=manual

Check that styling changes are present and the break occurs correctly with multiple word items.

### DIFF
--- a/static/scss/answers/theme.scss
+++ b/static/scss/answers/theme.scss
@@ -67,14 +67,17 @@
         padding-left: var(--yxt-base-spacing);
         padding-right: var(--yxt-base-spacing);
         padding-bottom: calc(var(--yxt-base-spacing) * 0.2);
-        padding-top: calc(var(--yxt-base-spacing) * 0.5);
-        line-height: var(--yxt-base-spacing);
-        word-break: break-word;
         letter-spacing: 1.1px;
       }
 
       .yxt-Nav-item.is-active {
         border-bottom-width: .185rem;
+      }
+
+      .yxt-Nav-modal .yxt-Nav-item {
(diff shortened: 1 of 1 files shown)


static/scss/answers/theme.scss
```diff
@@ -67,6 +67,9 @@
         padding-left: var(--yxt-base-spacing);
         padding-right: var(--yxt-base-spacing);
         padding-bottom: calc(var(--yxt-base-spacing) * 0.2);
+        padding-top: calc(var(--yxt-base-spacing) * 0.5);
```
tmeyer2115, line 70: I would double check with Rose that they want the `padding-top` even on the first item in the "More" drop-down.

static/scss/answers/theme.scss
```diff
@@ -67,6 +67,9 @@
         padding-left: var(--yxt-base-spacing);
```
tmeyer2115, line 67: How is the DOM in the SDK component structured? Will this style impact all Nav tabs (not just those in the drop-down)?

nmanu1: Yes, this styling would affect all Nav tabs. Updated to only change the drop-down items!

tmeyer2115 (approved):

static/scss/answers/theme.scss
nmanu1: I checked with Rose and she said that the `padding-top` should not be applied to the first item, so I will change that

oshi97 (approved):

PR #993, opened 2021-11-02
add --pageNames arg to setup-test-site npm script

oshi97:
It can take a long time to build up the full test-site, so I added
"pageNames" arg that let you specify a subset of pages to build.
The pages and config dirs are also now cleared when you rerun the command.

J=SLAP-1677
TEST=manual

tested building a subset of pages with `npm run setup-test-site -- --pageNames people faqs`
tested an arg-less `npm run setup-test-site`


yen-tt (approved):
nice!

PR #995, opened 2021-11-04
move on-document-load after ManualInitializer is instantiated

oshi97:
This allows users to call window.AnswersExperience.init() from within the on-document-load partial.
I added an error message if you try to call window.AnswersExperience.init() before ManualInitializer.setup()
has been called.

While testing, I noticed a bug with our script/* partials that get inserted into inline js.
If the last line of one of these partials is a single line js comment, then the line immediately following
the partial in the template would also be commented out

i.e., the below template
```hbs
{{#*inline 'script-partial'}}// I will comment out the line after me too{{/inline}}
function () {
  {{> script-partial }}
}
```

gets the output
```js
function () {
  // I will comment out the line after me too}
```
You can easily test this out in the handlebars playground or in a local node script.
To fix this I added new lines to the end of script/* partials that would be affected by this.

J=SLAP-1673
TEST=manual

tested calling window.AnswersExperience.init inside on-document-load with initializeManually set to true/false, and doing an end-to-end with consumer auth


layouts/html.hbs
```diff
@@ -7,6 +7,7 @@
         window.isOverlay = true;
       }
       {{> script/translations}}
+
```
oshi97, line 10: At first, I wanted to have it look like 
```{{ jsPartial 'script/translations' }}``` and have the jsPartial helper render the partial.
This isn't possible without reaching into the global handlebars instance, which I didn't want to do since the fact that we use the global hbs instance is kind of a jambo implementation detail, and we actually talked about having jambo not use the global instances and instead use Handlebars.create() and pass around an isolated instance.

Then, I tried out `{{#addNewLineToEnd}}{{> script/translations }}{{/addNewLineToEnd}}`, which I thought was a little ugly, but also don't mind this kind of approach if people prefer it. It's a lot more explicit. I intentionally what it would look like as the first commit to this pr.

cea2aj:
Looks like the 'es--field-direct-answer' percy snapshot isn't returning the right data anymore, however that's likely out of scope

cea2aj (approved):

PR #997, opened 2021-11-04
set default cta linkType to URL

oshi97:
The API is allowed to send an undefined cta linkType. Setting
the default to URL effectively grabs the link with no extra
formatting.

J=SLAP-1655
TEST=auto

### DIFF
--- a/static/js/formatters/generate-cta-field-type-link.js
+++ b/static/js/formatters/generate-cta-field-type-link.js
@@ -10,7 +10,7 @@ export function generateCTAFieldTypeLink(cta) {
   }
   const normalizedCTA = {
     ...cta,
-    linkType: normalizeCtaLinkType(cta.linkType)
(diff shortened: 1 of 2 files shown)


static/js/formatters/generate-cta-field-type-link.js
```diff
@@ -10,7 +10,7 @@ export function generateCTAFieldTypeLink(cta) {
   }
   const normalizedCTA = {
     ...cta,
-    linkType: normalizeCtaLinkType(cta.linkType)
+    linkType: normalizeCtaLinkType(cta.linkType || 'URL')
```
cea2aj, line 13: can we add a comment indicating that 'URL' will use the link with no formatting applied?

oshi97: updated! I put it in the function's main jsdoc instead of the function body

cea2aj (approved):

PR #998, opened 2021-11-08
Update image formatter to use dynamic thumbnailer

nmanu1:
Change the image formatter to use dynamic thumbnailer instead of expecting an array of thumbnails. A thumbnail array will no longer be provided after the streams migration. Dynamic thumbnailer uses the host `dynl` for images at least as large as the specified size (the smallest image with at least one dimension greater than or equal to the specified values). The host `dynm` gives the largest image that is smaller than the specified values in both dimensions.

Note: for `dynl`, desired size values of 1 or smaller are a special case which indicates that dimension should be ignored and the returned thumbnail should be larger than the other dimension.

J=SLAP-1670
TEST=auto, manual

Check that the current profile images still appear as before and those in the streams test account now appear as well.

### DIFF
--- a/static/js/formatters-internal.js
+++ b/static/js/formatters-internal.js
@@ -281,77 +281,71 @@ export function joinList(list, separator) {
   return list.join(separator);
 }
 
-/*
- * Given object with url and alternateText, changes url to use https
+/**
+ * Given an image object with a url, changes the url to use dynamic thumbnailer and https.
+ * 
+ * @param {Object} simpleOrComplexImage An image object with a url
+ * @param {string} desiredSize The desired size of the image ('<width>x<height>')
+ * @param {boolean} atLeastAsLarge Whether the image should be larger than the desired size or smaller
+ * @returns {Object} An object with a url for dynamic thumbnailer
  */
-export function image(simpleOrComplexImage = {}, size = '200x', atLeastAsLarge = true) {
-  let img = simpleOrComplexImage.image || simpleOrComplexImage;
-  if (!img) {
+export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAsLarge = true) {
+  let image = simpleOrComplexImage.image || simpleOrComplexImage;
+  if (!image) {
     return {};
   }
-
-  if (!img.url) {
-    return img;
+  if (!image.url) {
+    return image;
+  }
+  if (!(Object.prototype.toString.call(image).indexOf('Object') > 0)) {
+    throw new Error("Expected parameter of type Map");
+  }
+  if ((typeof desiredSize !== 'string') || (desiredSize == null)) {
+    throw new Error(`Object of type string expected. Got ${typeof desiredSize}.`);
+  }
+  if (desiredSize.indexOf('x') === -1) {
+    throw new Error("Invalid desired size");
+  }
+  if ((typeof atLeastAsLarge !== 'boolean') || (atLeastAsLarge == null)) {
+    throw new Error(`Object of type boolean expected. Got ${typeof atLeastAsLarge}.`);
   }
 
-  function createDynamicUrl(image, desiredSize, atLeastAsLarge = true) {
-    if ((image == null) || !(Object.prototype.toString.call(image).indexOf('Object') > 0)) {
-      throw new Error("Expected parameter of type Map");
-    }
-    if ((typeof desiredSize !== 'string') || (desiredSize == null)) {
-      throw new Error(`Object of type string expected. Got ${typeof desiredSize}.`);
-    }
-    if (desiredSize.indexOf('x') === -1) {
-      throw new Error("Invalid desired size");
-    }
-    if ((typeof atLeastAsLarge !== 'boolean') || (atLeastAsLarge == null)) {
-      throw new Error(`Object of type boolean expected. Got ${typeof atLeastAsLarge}.`);
-    }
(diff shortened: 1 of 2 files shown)


static/js/formatters-internal.js
```diff
@@ -340,15 +324,28 @@ export function image(simpleOrComplexImage = {}, size = '200x', atLeastAsLarge =
...
-      .filter(thumb => thumb.width && thumb.height)
-      .sort((a, b) => b.width - a.width);
-    return atLeastAsLarge
-      ? _getSmallestThumbnailOverThreshold(thumbnails, desiredWidth, desiredHeight)
-      : _getLargestThumbnailUnderThreshold(thumbnails, desiredWidth, desiredHeight);
+
+    const [urlWithoutExtension, extension] = _splitUrlOnIndex(image.url, image.url.lastIndexOf('.'));
+    const [urlBeforeDimensions, dimensions] = _splitUrlOnIndex(urlWithoutExtension, urlWithoutExtension.lastIndexOf('/') + 1);
+
+    if (desiredDims[0] === '' || desiredDims[1] === '') {
```
yen-tt, line 331: I think you can combine/refactor this chunk of logic with the above lines 314-234 where it checks for `desiredDims[0|1] !== ''` so it's more concise

static/js/formatters-internal.js
```diff
@@ -294,7 +294,7 @@ export function image(simpleOrComplexImage = {}, size = '200x', atLeastAsLarge =
     return img;
   }
 
-  function imageBySizeEntity(image, desiredSize, atLeastAsLarge = true) {
+  function createDynamicUrl(image, desiredSize, atLeastAsLarge = true) {
```
yen-tt, line 297: I don't know if it needs to its own separate function if it's call once here only and it contains almost all of function `image`'s logic. Maybe we could just use the logic as is without the function or move this out to its own function instead of nested inside image?

static/js/formatters-internal.js
```diff
@@ -294,7 +294,7 @@ export function image(simpleOrComplexImage = {}, size = '200x', atLeastAsLarge =
     return img;
```
yen-tt, line 294: can we add jsdoc for the params of this image function?

static/js/formatters-internal.js
```diff
@@ -340,15 +324,28 @@ export function image(simpleOrComplexImage = {}, size = '200x', atLeastAsLarge =
...
+      } else {
+        const fullSizeDims = dimensions.split('x');
+        desiredWidth = desiredWidth ?? Number.parseInt(fullSizeDims[0]);
+        desiredHeight = desiredHeight ?? Number.parseInt(fullSizeDims[1]);
+      }
+    }
+
+    const urlWithDesiredDims = urlBeforeDimensions + desiredWidth + 'x' + desiredHeight + extension;
+
+    return atLeastAsLarge ? _replaceUrlHost(urlWithDesiredDims, 'dynl.mktgcdn.com')
```
yen-tt, line 344: nit: newline on `?`

static/js/formatters-internal.js
nmanu1: sure!

static/js/formatters-internal.js
nmanu1: updated!

static/js/formatters-internal.js
nmanu1: updated!

static/js/formatters-internal.js
```diff
@@ -281,134 +281,97 @@ export function joinList(list, separator) {
...
-    const thumb = thumbnails[index];
-    const widthOverThreshold = minWidth ? thumb.width >= minWidth : true;
-    const heightOverThreshold = minHeight ? thumb.height >= minHeight : true;
-    if (widthOverThreshold && heightOverThreshold) {
-      return thumb.url
-    }
-    index--;
-  }
-  return thumbnails[0].url;
+function _splitUrlOnIndex(url, index) {
```
yen-tt, line 360: probably just splitStringOnIndex? since it's not specific to url

static/js/formatters-internal.js
```diff
@@ -281,134 +281,97 @@ export function joinList(list, separator) {
...
-    if (widthOverThreshold && heightOverThreshold) {
-      return thumb.url
-    }
-    index++;
-  }
-  return thumbnails[thumbnails.length - 1].url;
+function _replaceUrlHost(url, host) {
+  const splitUrl = url.split('://');
+  const urlAfterHost = splitUrl[1].slice(splitUrl[1].indexOf('/'));
+  return splitUrl[0] + '://'  + host + urlAfterHost;
```
yen-tt, line 374: I wonder if there's a simple Regex.replace(...) that we can do here? If not that's fine. This is good too

nmanu1: I think the regex would be something like `url.replace(/(\w+:\/\/)([^\/]+)(\/.*)/, '$1${host}$3')`. It seems harder to understand to me, but I can use that if it would be better

static/js/formatters-internal.js
nmanu1: changed!

static/js/formatters-internal.js
yen-tt: I see, yeah that's alright then. Let's keep it as is

yen-tt (approved):

PR #999, opened 2021-11-09
CaC Percy and WCaG testing infrastructure

yen-tt:
Update testing infrastructure to CaC approach for percy and wcag tests.

- added `PageOperator` which is responsible fo automating browser navigation using the given test locations
- test locations provide information on what page to navigate to and what action/query to perform on that page
- update wcag and percy index.js file to use PageOperator
- added some snapshots for es/ar pages that wasn't there before
- remove photographer, multilangphotographer, and wcagreporter class as they are no longer needed with pageOperator in use

J=1657
TEST=auto

see that wcag tests run successfully on local machine
see that percy snapshots are taken properly in this pr


tests/browser-automation/pageoperator.js
```diff
@@ -0,0 +1,59 @@
+const { BrowserPageWidths } = require("./constants");
+
+/**
+ * Responsible for browser navigation based on given test locations.
+ * Each test location is a JSON object which may contains:
```
cea2aj, line 5: nit: It looks like they are regular javascript objects and not JSON, so I would say: "... is an object which may contain:"

cea2aj:
Can you take a look at the es--vertical-full-page-map__mobile-detail-view snapshot? It looks a little odd in percy. Most of the snapshots look good, but I would just double check they look as expected

cea2aj:
Looks great, Yen! Plus the PR even removes over a 100 lines of code

yen-tt:
> Can you take a look at the es--vertical-full-page-map__mobile-detail-view snapshot? It looks a little odd in percy. Most of the snapshots look good, but I would just double check they look as expected

accidentally used `viewPort` instead of `viewport`. The snapshots for mobile view should be fixed now. The rest look good when I checked!

cea2aj (approved):

PR #1005, opened 2021-11-15
Card layout updates

cea2aj:
Update the layout of the card feedback buttons so that they are located in their own row underneath the CTAs and the card content

This change was in response to feedback from Product/UX. This PR moves the feedback buttons out from the `HitchhikerCard-actions` div into the card body. We also remove the `HichhikerCard-actions` div since that was only there so that the buttons could be next to the CTAs. This allows the card content to be full width when CTAs are not present. It adds some extra height to most cards since there's an entire row dedicated to the feedback buttons, however UX was happy with this decision. As part of this change, I removed some styling which is no longer relevant.

This PR also ensures that the feedback buttons are in the bottom right on grid layout. (Previously the buttons could appear just below the card content in grid layout just below the CTAs)

Also set a fixed feedbackContent height so that the cards don't shift when the buttons are pressed.

J=SLAP-1702
TEST=manual

Visual inspect all cards with feedback buttons enabled and also check mobile layouts. Test the product cards on the grid layout and confirm the feedback buttons are in the bottom right. Test clicking the feedback buttons and observing that the cards do not shift

### DIFF
--- a/cards/document-standard/template.hbs
+++ b/cards/document-standard/template.hbs
@@ -9,9 +9,9 @@
       </div>
       <div class="HitchhikerCard-actions">
         {{> ctas }}
-        {{> thumbsfeedback card feedbackSubmitted=feedbackSubmitted}}
       </div>
     </div>
+    {{> thumbsfeedback card feedbackSubmitted=feedbackSubmitted}}
   </div>
 </div>
 
--- a/cards/document-standard/template.hbs
+++ b/cards/document-standard/template.hbs
@@ -7,9 +7,7 @@
       <div class="HitchhikerDocumentStandard-info">
         {{> details }}
       </div>
-      <div class="HitchhikerCard-actions">
-        {{> ctas }}
-      </div>
+      {{> ctas }}
     </div>
     {{> thumbsfeedback card feedbackSubmitted=feedbackSubmitted}}
   </div>
--- a/cards/product-prominentvideo/template.hbs
+++ b/cards/product-prominentvideo/template.hbs
@@ -50,7 +50,7 @@
 {{/inline}}
 
 {{#*inline 'content'}}
-{{#if (any card.details (all card.CTA1 card.CTA1.url card.CTA1.label) (all card.CTA2 card.CTA2.url card.CTA2.label) card.feedbackEnabled)}}
+{{#if (any card.details (all card.CTA1 card.CTA1.url card.CTA1.label) (all card.CTA2 card.CTA2.url card.CTA2.label))}}
   <div class="HitchhikerProductProminentVideo-contentWrapper">
     {{> details }}
     {{> ctas }}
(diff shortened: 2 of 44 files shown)


static/scss/answers/cards/product-prominentimage-clickable.scss
```diff
@@ -40,6 +40,9 @@
   {
     width: 100%;
     padding: calc(var(--yxt-base-spacing) * 3/4) var(--yxt-base-spacing);
+    display: flex;
+    flex-direction: column;
+    height: 100%;
```
oshi97, line 45: why do we need height: 100%? can we use flex-grow?

cards/document-standard/template.hbs
```diff
@@ -9,9 +9,9 @@
       </div>
       <div class="HitchhikerCard-actions">
```
nmanu1, line 10: I had added the `HitchhikerCard-actions` div to have the feedback buttons follow the placement of the CTAs. Since this is no longer wanted, the div and it's styling can probably be removed from all templates

static/scss/answers/cards/product-prominentimage-clickable.scss
cea2aj: ah, you're right, I'll try that instead

cards/document-standard/template.hbs
cea2aj: got it, will do!

cea2aj: I removed all `HichhikerCard-actions`

static/scss/answers/module.scss
```diff
@@ -59,10 +44,8 @@
     justify-content: flex-end;
     align-items: flex-end;
     flex-grow: 1;
-    @include bpgte(sm)
-    {
-      margin-top: calc(var(--yxt-base-spacing) / 2);
-    }
+    margin-top: calc(var(--yxt-base-spacing) / 2);
+    height: auto;
```
oshi97, line 48: just checking that we need to specify height: auto?

cards/product-prominentvideo/template.hbs
```diff
@@ -52,10 +53,7 @@
 {{#if (any card.details (all card.CTA1 card.CTA1.url card.CTA1.label) (all card.CTA2 card.CTA2.url card.CTA2.label) card.feedbackEnabled)}}
```
nmanu1, line 53: `card.feedbackEnabled` can be removed now that the feedback buttons aren't included in the `contentWrapper` div

static/scss/answers/module.scss
cea2aj: you're right, I needed it in a previous version of this PR, but it's not needed anymore

cards/product-prominentvideo/template.hbs
cea2aj: good catch! updated

yen-tt (approved):

PR #1006, opened 2021-11-15
Fix distance alignment

nmanu1:
Fix distance alignment for financial-professional-location and professional-location cards so that distance is on the far right of the top row of the card instead of right next to the title. Also, fix distance alignment with the close card button on vertical full page map for both cards and location-standard cards so that distance is in-line to the left of the button.

J=SLAP-1718
TEST=manual

Check that both professional location cards (and multilang) have correct distance alignment, matching the location-standard card. Check that mobile view of vertical full page map has correct alignment of distance with close card button for location-standard card


yen-tt (commented):
instead of changing the templates and adjusting the ordinal and title for all cards, I believe you can just add `margin-left: auto` to `HitchhikerLocationCard-closeCardButtonWrapper` for these cards and that should push the distance to far right. This would work for the last element, and that the topRow div's display mode is flex

nmanu1:
I was mimicking how this is done for the location-standard card for consistency. But I can change it to this if it's preferred

cards/financial-professional-location/template.hbs
```diff
@@ -1,10 +1,13 @@
...
   {{> image }}
   <div class="HitchhikerFinancialProfessionalLocation-body">
     {{#if (any card.title card.distance)}}
     <div class="HitchhikerFinancialProfessionalLocation-topRow">
-      {{> ordinal }}
-      {{> title }}
+      {{> ordinalAndTitle displayOrdinal=(all card.showOrdinal result.ordinal) }}
       <div class="HitchhikerLocationCard-closeCardButtonWrapper">
         {{> closeCardButton }}
         {{> distance }}
```
yen-tt, line 13: Should distance be part of `closeCardButtonWrapper` div? wouldn't this look a bit weird if both distance and closeCardButton is display on the card?

nmanu1: yes, I think it would look weird. but I think we would want both of these to be right-aligned. so maybe distance should go before the button and the wrapper should have flex styling?

yen-tt: looks like the button is not configurable? seems to have css display: none by default. Perhaps you could checkin with rose if this needs to be fix. And regard to the distance alignment, maybe she would have an opinion between the template changes to be same as location-standard card, or css styling changes only, which is what I'm kind of leaning toward instead of merging ordinal and title

cea2aj: Since `distance` is already part of `closeCardButtonWrapper`, I think that's out of scope of this QA item. As for the DOM changes, no CSS classes are being removed in this PR which means that most styling that is currently targeting them wouldn't break during an upgrade. So I'd say handling this with CSS only changes isn't critical

cea2aj (approved):

nmanu1:
Note: translation test will fail until merged into the release branch because of how the merge conflict was resolved

yen-tt (approved):

PR #1010, opened 2021-11-17
Update default universal limit

nmanu1:
Change the default universalLimit for all vertical page-configs to 4.

J=SLAP-1681
TEST=manual

Smoke test in test-site and see that faq universal results are limited to 4 when uncommented.

### DIFF
--- a/templates/vertical-full-page-map/page-config.json
+++ b/templates/vertical-full-page-map/page-config.json
@@ -64,7 +64,7 @@
     "<REPLACE ME>": { // The vertical key from your search configuration
       // "label": "", // The name of the vertical in the section header and the navigation bar
       // "verticalLimit": 15, // The result count limit for vertical search
-      // "universalLimit": 5, // The result count limit for universal search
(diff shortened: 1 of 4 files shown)


templates/vertical-full-page-map/page-config.json
```diff
@@ -64,7 +64,7 @@
     "<REPLACE ME>": { // The vertical key from your search configuration
       // "label": "", // The name of the vertical in the section header and the navigation bar
       // "verticalLimit": 15, // The result count limit for vertical search
-      // "universalLimit": 5, // The result count limit for universal search
+      // "universalLimit": 4, // The result count limit for universal search
```
cea2aj, line 67: Just confirming, Rose doesn't want us to remove the comment here so that this takes effect by default, correct?

nmanu1: Checked with Rose and it actually should be uncommented now, so I'll update that

yen-tt (approved):

oshi97 (approved):

PR #1012, opened 2021-11-24
Escape angle brackets in highlightField

yen-tt:
- perform full HTML escape on the constructed highlighted text (will transform ", ', &, <, and >)
- `substr` is deprecated, replaced with `substring`

J=SLAP-1703
TEST=manual & auto

update bryan's entity description to include script tags, search for 'where is joe exotic' and see that the script tags appear. (without the replacement, they do not appear)
see added jest tests passed

### DIFF
--- a/static/js/formatters-internal.js
+++ b/static/js/formatters-internal.js
@@ -543,7 +543,7 @@ export function priceRange(defaultPriceRange, countryCode) {
 export function highlightField(fieldValue, matchedSubstrings = []) {
   let highlightedString = '';
 
-  // We must first sort the matchedSubstrings by decreasing offset. 
+  // We must first sort the matchedSubstrings by ascending offset. 
   const sortedMatches = matchedSubstrings.slice()
     .sort((match1, match2) => match1.offset - match2.offset);
   
--- a/static/js/formatters-internal.js
+++ b/static/js/formatters-internal.js
@@ -540,7 +540,7 @@ export function priceRange(defaultPriceRange, countryCode) {
  *                                          highlight.
  */
 export function highlightField(fieldValue, matchedSubstrings = []) {
-  let highlightedString = fieldValue.replace(/</g,'&lt;').replace(/>/g,'&gt;');
+  let highlightedString = fieldValue;
 
   // We must first sort the matchedSubstrings by decreasing offset. 
   const sortedMatches = matchedSubstrings.slice()
@@ -554,6 +554,13 @@ export function highlightField(fieldValue, matchedSubstrings = []) {
       highlightedString.substr(offset + length);
   });
 
+  /**
+   * Use regex to match all tags that's not <mark> tag, and replaced angle brackets
+   * with '&lt;' and '&gt;' to properly display them. The regex pattern uses
+   * lookahead syntax to check if the angle bracket '<' is follow by 'mark' or '/mark'
+   * before closing it with '>'. If the tag is not a mark tag, replace the angle brackets.
+   */
+  highlightedString = highlightedString.replace(/(<)((?!\/?mark)[^>]*)(>)/g,'&lt;$2&gt;');
   return highlightedString;
 }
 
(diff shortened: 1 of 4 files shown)


static/js/formatters-internal.js
```diff
@@ -554,6 +554,12 @@ export function highlightField(fieldValue, matchedSubstrings = []) {
       highlightedString.substr(offset + length);
   });
 
+  /**
+   * Use regex with uses lookahead syntax to match all tags that's not <mark> or </mark> tag,
```
tmeyer2115: which uses lookahead syntax

static/js/formatters-internal.js
```diff
@@ -554,6 +554,12 @@ export function highlightField(fieldValue, matchedSubstrings = []) {
       highlightedString.substr(offset + length);
   });
 
+  /**
+   * Use regex with lookahead syntax to match all tags that's not <mark> or </mark> tag,
+   * and replace angle brackets with '&lt;' and '&gt;' to properly display them as non-html code.
+   * If matched, preserve the content ($2) inside but replace the surrounding angle brackets.
+   */
+  highlightedString = highlightedString.replace(/(<)((?!\/?mark)[^>]*)(>)/g,'&lt;$2&gt;');
```
cea2aj, line 562: If there are multiple angle brackets, then they will still be left in the string. (e.g. "<<<script>>>" becomes `&lt;<<\script&gt;>>`). I'm wondering if it would be simpler to do a full escape of `<`, `>`, and `&`, and then we unescape just the mark tags

cea2aj: Alternatively we could indicate the marks with some other syntax, do a full HTML escape, and then transform the mark tags into HTML

oshi97: I was thinking the replacement should happen before the forEach loop. That way we don't have to rely on any kind of regex, or any issues with somebody including `<mark>` tags in the original string

tmeyer2115: I'm a little wary of trying to accomplish this via a complex regex. We've been bitten by that before. As Connor noted, it's hard to get it right. Instead of trying to roll our own here, maybe we can use an existing library for escaping the HTML (such as https://www.npmjs.com/package/escape-html). I like Connor's idea of marking the highlights with some other (non-HTML) syntax, doing the escaping, then adding in the mark tags.

oshi97: Specifically I was thinking
1. slice up the string into substrings
2. replace all angle brackets in the substrings
3. wrap all substrings (except for the first and last) with mark tags

yen-tt: I had the replacement before the for loop initially but that messes up the indexes. Connor's suggestion makes sense, I will try out the library and add in the mark tag afterward

static/js/formatters-internal.js
```diff
@@ -550,10 +551,11 @@ export function highlightField(fieldValue, matchedSubstrings = []) {
     const { offset, length } = match;
     highlightedString = 
       highlightedString.substr(0, offset) +
-      `<mark>${fieldValue.substr(offset, length)}</mark>`+
+      `|mark|${fieldValue.substr(offset, length)}|/mark|`+
```
oshi97, line 554: doesn't this run into issues if somebody has `|mark|` in their actual text?

static/js/formatters-internal.js
```diff
@@ -540,20 +541,20 @@ export function priceRange(defaultPriceRange, countryCode) {
  *                                          highlight.
  */
 export function highlightField(fieldValue, matchedSubstrings = []) {
-  let highlightedString = fieldValue;
+  let highlightedString = '';
 
   // We must first sort the matchedSubstrings by decreasing offset.
```
oshi97, line 546: This comment is out of date now that we are sorting by ascending offset

oshi97: could we get some unit tests as well? specifically for cases with multiple highlighted substrings, since those are a little tricky

yen-tt: yup, just added!

static/js/formatters-internal.js
yen-tt: yea decided to just escape in the substrings

cea2aj (approved):

PR #1014, opened 2021-11-29
Update escape packages

yen-tt:
- update escape-html to be devDependency in top level package
- add escape-html as depeendency in static package


static/package-lock.json
```diff
@@ -1,8 +1,9479 @@
 {
   "name": "answers-hitchhiker-theme",
   "version": "1.26.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
oshi97, line 4: I think we don't want to change the lockfile ver until we're sure the CBD containers can handle it

oshi97:
think the original lockfile was deleted by accident

yen-tt:
> think the original lockfile was deleted by accident

oh it's just a lot of changes since I was changed it to lockfile version 1 for consistency. the file is still there

oshi97 (approved):

PR #1015, opened 2021-12-02
Remove scrollbars from filters

nmanu1:
Change styling of `yxt-FilterOptions-options` to remove scrollbars that were appearing on filters even when all options were displayed for some cases with custom font sizes. Add a `--yxt-filters-and-sorts-line-height` variable so that customers can change the line-height of the `optionLabel` easily if the scrollbar appears with even larger custom font sizes.

J=SLAP-1727
TEST=visual

See that filter scrollbar was removed on a site where it was previously appearing.

### DIFF
--- a/static/scss/answers/collapsible-filters/sdk-filter-overrides.scss
+++ b/static/scss/answers/collapsible-filters/sdk-filter-overrides.scss
@@ -23,8 +23,9 @@
     }
   
     &-options {
-      margin-top: 4px;
-      margin-bottom: 4px;
+      margin: 0;
+      padding-top: 4px;
+      padding-bottom: 4px;
       padding-left: 4px;
     }
   
--- a/static/scss/answers-variables.scss
+++ b/static/scss/answers-variables.scss
@@ -51,7 +51,7 @@ $container-width: 1000px;
   --yxt-searchbar-button-text-color-hover: var(--yxt-color-brand-primary);
   --yxt-filter-options-options-max-height: none;
   --yxt-filters-and-sorts-font-size: var(--yxt-font-size-md);
-  --yxt-filters-and-sorts-line-height: 20px;
+  --yxt-filters-option-label-line-height: 20px;
   --yxt-alternative-verticals-emphasized-font-weight: var(--yxt-font-weight-semibold);
   --yxt-text-snippet-highlight-color: #eef3fb;
   --yxt-text-snippet-font-color: var(--yxt-color-text-primary);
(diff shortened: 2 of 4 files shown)


static/scss/answers/collapsible-filters/sdk-filter-overrides.scss
```diff
@@ -23,8 +23,9 @@
     }
   
     &-options {
-      margin-top: 4px;
-      margin-bottom: 4px;
+      margin: 0;
```
tmeyer2115, line 26: Is the `margin: 0` necessary here? Do the options inherit margin from somewhere up the CSS chain?

tmeyer2115 (commented):
Can you also verify that a site without this issue still looks/works as expected? Rose said not all sites seemed to suffer from this problem. I think it would also be good to verify Rose's suspicion that font size and margin are the culprit here. Were you able to verify that?

nmanu1:
The underlying problem is that if a custom font-size is specified for `yxt-FilterOptions-optionLabel`, but the line-height is left as is, then a large custom font-size can make the text height larger than the line height, causing the scrollbar to appear.

static/scss/answers/collapsible-filters/sdk-filter-overrides.scss
nmanu1: Yes, it's inheriting margin from the SDK styling for `yxt-FilterOptions-options`

tmeyer2115:
> The underlying problem is that if a custom font-size is specified for `yxt-FilterOptions-optionLabel`, but the line-height is left as is, then a large custom font-size can make the text height larger than the line height, causing the scrollbar to appear.
> 
> Edit: After talking with Rose, I added a variable so that the line height can be changed easily from the same place the font size is customized.

Given this, do we still need to remove the margin and replace it with padding in `sdk-filter-overrides.scss`? If the problem is really the relationship between font size and line height, I think customers experiencing this issue should use the new variable to update their line height. Us making the CSS changes in the overrides file almost masks the real issue from them.

nmanu1:
> Given this, do we still need to remove the margin and replace it with padding in `sdk-filter-overrides.scss`? If the problem is really the relationship between font size and line height, I think customers experiencing this issue should use the new variable to update their line height. Us making the CSS changes in the overrides file almost masks the real issue from them.

I asked Rose if we should just add the variable, but she was in favor of also modifying the padding/margin to give customers a little more lee-way before the scrollbar appears.

tmeyer2115 (approved):

PR #1021, opened 2021-12-13
Fix handling of the `universalLimit` in `verticalsToConfig`.

tmeyer2115:
Previously, if a `universalLimit` were specified for a vertical, we would perform client-side filtering. Instead, we should be creating a `search.universalLimit` object and passing that to `ANSWERS.init`. Not only does this remove responsibility from the front-end, but it allows the HH to display more than 10 results per vertical on the Universal page.

This PR adds a `getDefaultUniversalLimit` helper that iterates through the vertical page configs and uses the `verticalsToConfig.universalLimits` to build a `search.universalLimit`. 

J=SLAP-1749
TEST=manual

Tested the following scenarios and verified correct behavior in each:

- No `universalLimit` values are set and there is no limit specified in the `global_config` or `pageSettings`.
- Some `universalLimit` values are set, but not others. There is no universal limit in the `global_config` or `pageSettings`.
- Some `universalLimit` values are set, but a universal limit is also present in the `global_config`.

### DIFF
--- a/hbshelpers/getDefaultUniversalLimit.js
+++ b/hbshelpers/getDefaultUniversalLimit.js
@@ -9,15 +9,15 @@
  */
  module.exports = function getDefaultUniversalLimit(pageConfigs) {
   const universalLimit = Object.entries(pageConfigs)
-    .filter(([key, _]) => key != 'index')
-    .reduce((limit, [_, config]) => {
-      const verticalKey = config.verticalKey;
+    .filter(([pageName, _]) => pageName != 'index')
+    .reduce((limit, [_, pageConfig]) => {
+      const verticalKey = pageConfig.verticalKey;
       const hasUniversalLimit = 
-        config.verticalsToConfig &&
-        config.verticalsToConfig[verticalKey] &&
-        config.verticalsToConfig[verticalKey].universalLimit;
+        pageConfig.verticalsToConfig &&
+        pageConfig.verticalsToConfig[verticalKey] &&
+        pageConfig.verticalsToConfig[verticalKey].universalLimit;
       if (hasUniversalLimit) {
-        limit[verticalKey] = config.verticalsToConfig[verticalKey].universalLimit;
+        limit[verticalKey] = pageConfig.verticalsToConfig[verticalKey].universalLimit;
       }
 
       return limit;
--- a/hbshelpers/getDefaultUniversalLimit.js
+++ b/hbshelpers/getDefaultUniversalLimit.js
@@ -5,7 +5,7 @@
  * be used.
  * 
  * @param {Object} pageConfigs - The configurations for each page.
- * @returns {Object} The default search.universalLimit for the page.
+ * @returns {Object} The partial of the search configuration related to universal limits.
  */
  module.exports = function getDefaultUniversalLimit(pageConfigs) {
   const universalLimit = Object.entries(pageConfigs)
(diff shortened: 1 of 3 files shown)


hbshelpers/getDefaultUniversalLimit.js
```diff
@@ -0,0 +1,27 @@
...
+ * default is taken from the vertical's `universalLimit`, specified in the related
+ * page's VTC. If there is no such `universalLimit`, the back-end default of 10 will
+ * be used.
+ * 
+ * @param {Object} pageConfigs - The configurations for each page.
+ * @returns {Object} The default search.universalLimit for the page.
+ */
+ module.exports = function getDefaultUniversalLimit(pageConfigs) {
+  const universalLimit = Object.entries(pageConfigs)
+    .filter(([key, _]) => key != 'index')
```
oshi97, line 12: I think you can just omit the second variable in the tuple instead of using an underscore

hbshelpers/getDefaultUniversalLimit.js
```diff
@@ -0,0 +1,27 @@
...
+    .reduce((limit, [_, config]) => {
+      const verticalKey = config.verticalKey;
+      const hasUniversalLimit =
```
oshi97, line 15: the following code might be slightly simpler if this is just `const universalLimit = ...` but also feels like a personal preference thing

hbshelpers/getDefaultUniversalLimit.js
```diff
@@ -0,0 +1,27 @@
...
+        config.verticalsToConfig[verticalKey] &&
+        config.verticalsToConfig[verticalKey].universalLimit;
+      if (hasUniversalLimit) {
+        limit[verticalKey] = config.verticalsToConfig[verticalKey].universalLimit;
+      }
+
+      return limit;
+    }, {});
+
+    return { universalLimit };
```
oshi97, line 26: should we wrap the universalLimit object or just return it directly?

tmeyer2115: I return it wrapped as the SDK expects something of the form: 
```
search: {
   universalLimit: {
      someVertical: someNumber
   }
}
```
Wrapping it here means less clutter in the HBS template.

hbshelpers/getDefaultUniversalLimit.js
tmeyer2115: This is just a personal preference thing, but I actually prefer the underscore. I think it quickly communicates to the reader that I don't care about the other value. Seeing `.filter([key])` would give me pause for a second or too since I know `Object.entries` returns a tuple.

oshi97: unrelated to the first comment but github won't let me start a new thread. Could we rename `key` to maybe `pageName`?

hbshelpers/getDefaultUniversalLimit.js
tmeyer2115: So I had something like that initially.

```
const universalLimit = blah?.blah?.universalLimit
```

But somehow that syntax routinely caused our Percy snapshot generator to crash.

hbshelpers/getDefaultUniversalLimit.js
oshi97: should we update the jsdoc a little? it says `The default search.universalLimit for the page.` but it's really a partial of the `search` object

tmeyer2115: Updated!

PR #1017, opened 2021-12-09
move spellcheck above results count in vertical-map template

oshi97:
J=SLAP-1678
TEST=manual

rebuilt test site, looks as expected

PR #996, opened 2021-11-04
Sandbox/Production environment config

yen-tt:
- add commented out `environment` in global_config.json

J=SLAP-1663
TEST=manual

see the environment is change based on value in global_config.json

PR #1022, opened 2021-12-14
Version 1.27

tmeyer2115:
### Features
- Consumer Authentication support was added for the Sandbox environment. (#996)
- The existing `image` formatter was updated to support photos sent from the Streams API. (#998)
- Support for Direct Answers on Vertical was added to the `vertical-standard` template. It is commented out by default. (#994)

### Changes
- To better support Consumer Authentication, the `AnswersExperience.init()` method can be called on `Document` load. (#995)
- The default `universalLimit` for all Vertical page configs was updated to 4. (#1010, #1021)

### Bugfixes
- Ensured that in all page templates, the `SpellCheck` appears above the `ResultsCount`. (#1011, #1017)
- In the `highlightedField` formatter, any HTML tag that appears in the text, that is not `<mark>` or `</mark>`, is now escaped. (#1012)
- Font pre-loads on Multi-lang sites now work correctly. (#1018)
- A new CSS variable was added: `--yxt-filter-options-option-label-line-height`. This variable, when kept in proper proportion to `--yxt-filters-and-sorts-font-size`, will ensure the scroll bar does not erroneously appear for filter options. (#1015, #1019)


nmanu1:
Should the referenced PR number for updating the default universalLimit be 1010 (and maybe also 1021), instead of 1681?

nmanu1 (commented):
Should the referenced PR number for updating the default universalLimit be 1010 (and maybe also 1021), instead of 1681?

tmeyer2115:
> Should the referenced PR number for updating the default universalLimit be 1010 (and maybe also 1021), instead of 1681?

Fixed! I think I put the item number in by mistake.

nmanu1 (commented):
Do we need the PR number for the CSS variable for filter scroll bars?

yen-tt (approved):

PR #1024, opened 2021-12-14
Testcafe Fixes

cea2aj:
Fix Testcafe errors in safari by bumping the Testcafe version number

This PR fixes the "undefined is not an object (evaluating 'qa.objectHasOwnProperty.call(s,"writable")') hasOwnProperty@[native code]" error which was occurring in Testcafe on Safari.

Also fix the path to the testcafe config file.

J=none
TEST=none

Run the acceptance tests and see that they now pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "puppeteer": "^10.2.0",
     "serve": "^11.3.2",
     "simple-git": "^2.24.0",
-    "testcafe": "^1.15.0",
+    "testcafe": "^1.17.0",
     "testcafe-browser-provider-browserstack": "^1.13.1",
     "underscore.string": "^3.3.5",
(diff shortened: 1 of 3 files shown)


tmeyer2115 (approved):

package.json
```diff
@@ -47,7 +47,7 @@
     "puppeteer": "^10.2.0",
     "serve": "^11.3.2",
     "simple-git": "^2.24.0",
-    "testcafe": "^1.15.0",
+    "testcafe": "^1.17.0",
```
oshi97, line 50: should this be 1.17.1?

cea2aj: Based on the package-lock, it looks like 1.17.1 is the version which is actually being used. I can bump it here as well just to be safe

PR #1028, opened 2021-12-20
Acceptance Test Speedup

cea2aj:
Improve Acceptance Test Speed

The primary reason for the slowdown is that Browserstack will run IE11 tests across many browserstack sessions. The firefox and safari tests wait for each test to pass in IE11, which is why there are long periodes of time where nothing is happening in the safari and firefox test videos.  I am submitting a support ticket to browserstack to see if they can explain why this happens for IE11. Separating the tests into their own jobs prevents the safari and firefox tests from waiting for the IE11 video, which makes the browserstack videos much more useful.

For now, increasing the concurrency of IE11 runs helps alleviate this issue. I don't recommend we run IE11 with a concurrency less than 3 because it improves the test runtime to about 7 from 20-25. This puts us at 5 concurrent tests for non-dev branches. If Browserstack gets back with a fix, then we can decrease or remove the IE11 concurrency. So far we haven't ran into issues with too many running tests, but if it's a problem then we can consider decreasing the concurrency or not testing IE11 on so many branches.

Decreasing the zoom stabilization interval helped with performance by a few seconds overall.

Finally, I moved all the locator acceptance tests into one page because the request logger wasn't being setup properly with the two separate files. The request logger not working was causing the tests to take an additional 30 seconds because the `waitOnSearchComplete` would wait the full 10 seconds before proceeding. The issue was that `SearchRequestLogger.createVerticalSearchLogger()` was being called twice which created two separate `RequestLoggers` rather than one. When I reuse the same request logger across both test fixtures, the issue is resolved.

The browser in safari would sometimes not be wide enough for the full page map tests, so I set a fixed value which keeps the map in desktop mode.

J=1756
TEST=auto

Confirm tests all still pass

See that browserstack test speed improves from 1min 15seconds to about 45 seconds locally. See that browserstack speed improves from over 20 minutes to about 7.5 minutes for IE11.


.github/run_browserstack_acceptance.sh
```diff
@@ -5,13 +5,15 @@ GITHUB_BRANCH=${GITHUB_REF#refs/heads/}
...
+export BROWSERSTACK_DEBUG="true"
+export BROWSERSTACK_CONSOLE="verbose"
+export BROWSERSTACK_NETWORK_LOGS="true"
+export BROWSERSTACK_DISPLAY_RESOLUTION="1920x1080"
 
-if [[ $GITHUB_BRANCH == release/*
-  || $GITHUB_BRANCH == hotfix/*
-  || $GITHUB_BRANCH == master
-  || $GITHUB_BRANCH == support/* ]]
+if [[ $BROWSER == 'browserstack:ie@11.0' ]]
```
yen-tt, line 14: maybe we can just have this script take in a concurrency arg, and have the different jobs in the acceptance test github action pass in a number. What do you think?

.github/workflows/acceptance_headless.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - run: npm ci
+      - run: npm run setup-test-site
+      - run: npm run build-test-site
+      - name: Run Acceptance Tests
+        run: npm run acceptance -- --browsers chrome:headless
```
yen-tt, line 18: should this execute the script `run_browserstack_acceptance.sh` as well?

tests/acceptance/suites/vertical-full-page-map.js
```diff
@@ -6,12 +6,14 @@ import Pagination from '../blocks/pagination';
 import SearchRequestLogger from '../searchrequestlogger';
 import { VERTICAL_SEARCH_URL_REGEX } from '../constants';
 import { registerIE11NoCacheHook } from '../../test-utils/testcafe';
+import CollapsibleFilters from '../blocks/collapsiblefilters';
 
 fixture`Vertical Full Page Map`
-  .page(`http://localhost:${PORT}/locations_full_page_map`)
+  .page(`http://localhost:${PORT}/locations_full_page_map_with_filters`)
```
yen-tt, line 12: is there not a lot of differences between `locations_full_page_map` and `locations_full_page_map_with_filters` page? I'm not sure if we should just stop running tests on `locations_full_page_map`.  Does it fail if you have two separate fixtures, uses different url, in the same file?

.github/run_browserstack_acceptance.sh
cea2aj: sure, updated

.github/workflows/acceptance_headless.yml
cea2aj: The run_browserstack_acceptance script is used to setup the browserstack environment variables which isn't needed for chrome:headless

tests/acceptance/suites/vertical-full-page-map.js
cea2aj: The pages are exactly the same except the `_with_filters` page has filters enabled. All of the map functionality tested on the `locations_full_page_map` page should pass on the `locations_full_page_map_with_filters` page.  I can check if two separate fixtures work in the same file, but I don't see a need for it

cea2aj: Multiple fixtures worked fine, but I learned that I needed to re-use the same verticalSearchLogger or else the request logger wouldn't work

yen-tt: does the fixtures run concurrently? I'm wondering if using the same verticalSearchLogger could be misleading if one fixture receive response triggered from another fixture and just resolve the waitOnSearchComplete call.

cea2aj: The fixtures run sequentially so I don't think that's an issue

yen-tt (approved):

PR #1029, opened 2021-12-21
Add Locator support for the Job Entity Type

yen-tt:
Update ThemeMap to support coordinate and pin constructions based on data from job entity type. 
- update transformDataToUniversalData and transformDataToVerticalData to guard access to countryCode field since job entity will not have the address field like location entity.
- update ThemeMap to NOT construct pin if entity doesn't have `entity.profile.yextDisplayCoordinate`. 
   - Note: job entity may include only mapMarker field without a location field. This would not give us `entity.profile.yextDisplayCoordinate` but instead in `entity.profile.displayCoordinate`. Product only wants display from `entity.profile.yextDisplayCoordinate` for now.
- move cardFocusUpdateListener up to make sure selecting cards that doesn't have pin would still deselect previously selected pin

J=SLAP-1744
TEST=manual

Added some more job entities into knowledge graph: some with location field, some with map marker, some without any location or map marker:
- test with `locations_full_page_map` page with verticalKey: 'jobs' and cardName: 'location-standard'
  - See no errors in console.
  - Click on cards and pins. See that cards with corresponding pins work as normal. See that cards without pins would still highlight the card but unhighlight the previous selected pin.
  - Drag map arounds and see that only cards with pins shows up when rerender
  - modify pin to display index number, see that the index displayed is correct corresponding to the ordinal number of the cards
- test with `locations_full_page_map` page with verticalKey: 'jobs' and cardName: 'job-standard'
  - see no errors in console. 
  - pins still work as expected
- test with `locations_full_page_map` page with verticalKey: 'KM' and cardName: 'location-standard'
  - smoke test the page with location entities to confirm that it still work as expected with clusters and non clusters pins 
 - test with  job entities on other pages: 
   - `locations_full_page_map_with_filters`: no errors, pins and cards show up and can be selected as expected
   -  `locations_google` and `locations` (uses map from SDK): no errors, pins and cards show up
       - NOTE: SDK map component currently will NOT have the right the pin ordering with the card ordinal numbers if there are cards created from job entity without location field. This will be investigated in another item

### DIFF
--- a/cards/location-standard/component.js
+++ b/cards/location-standard/component.js
@@ -28,7 +28,7 @@ class location_standardCardComponent extends BaseCard['location-standard'] {
       // subtitle: '', // The sub-header text of the card
       hours: Formatter.openStatus(profile),
       // services: [], // Used for a comma delimited list of services for the location
-      address: Formatter.address(profile), // The address for the card
+      address: Formatter.address(profile) || profile.locationString || '', // The address for the card
       phone: Formatter.nationalizedPhoneDisplay(profile), // The phone number for the card
       phoneEventOptions: this.addDefaultEventOptions(), // The analytics event options for phone clicks
       distance: Formatter.toLocalizedDistance(profile), // Distance from the user’s or inputted location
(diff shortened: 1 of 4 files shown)


cea2aj (approved):
lgtm!

PR #1031, opened 2022-01-05
refactor global config merging strategy on upgrade

oshi97:
This commit refactors how global configs are merged on theme upgrades.
Now, global configs will always have the same "structure" as the incoming
config, instead of sometimes placing properties in seemingly random spots,
if the two configs don't match up exactly in terms of structure.

Per Rose's request, the sdkVersion is now always updated to the incoming
config's value. Besides that, preexisting properties will favor the
original config, while commented out properties will favor the incoming config.

`Property`s and `CommentedOutProperty`s in the original that do not exist in the incoming config
will be moved to the bottom of the config.
Comments that are neither inline nor `CommentedOutProperty`s will be ignored gracefully, and an error message displayed.

J=SLAP-1717
TEST=manual,auto

ran an upgrade from 1.26 to this branch
ran an upgrade from 1.26 to this branch, but with the global config changed to have all values replaced with new ones, and with an extra prop at the bottom


postupgrade/mergeGlobalConfigs.js
```diff
@@ -0,0 +1,174 @@
...
+
+  /**
+   * @param {CommentToken} comment 
+   * @returns {CommentedOutPropertyToken}
+   */
+  const parseCommentedOutProperty = (comment) => {
+    const propCommentRegex = /^\s?"([^"]+)":.*$/g
+    const val = comment.value
+    const match = [...val.matchAll(propCommentRegex)][0]
+    if (!match || match.length !== 2) {
```
tmeyer2115, line 68: In what situations could we hit this conditional? Does being a `CommentToken` not automatically imply that the regex will be matched?

postupgrade/mergeGlobalConfigs.js
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  const handleComment = c => {
+    if (c.inline) return;
+    tokens.push(parseCommentedOutProperty(c));
+  }
+
+  Object.keys(commentJsonObject).forEach(key => {
+    const beforeSymbol = Symbol.for(`before:${key}`)
```
tmeyer2115, line 88: Missing semi-colons at EOL here and below.

postupgrade/mergeGlobalConfigs.js
```diff
@@ -0,0 +1,174 @@
...
+ * @param {GlobalConfigToken[]} original 
+ * @param {GlobalConfigToken[]} incoming 
+ * @returns 
+ */
+ function mergeTokens(original, incoming) {
+  const mergedTokens = []
+  const originalCopy = [...original]
+
+  incoming.forEach(incomingToken => {
+    const i = originalCopy.findIndex(t => t.key === incomingToken.key)
```
tmeyer2115, line 154: Would it make sense to store the tokens in a Map instead of an array? That could save us this `findIndex` call here. Both `incoming` and `original` would have tokens by key.

postupgrade/mergeGlobalConfigs.js
```diff
@@ -0,0 +1,174 @@
...
+    switch (token.type) {
+      case 'Property':
+        commentJsonObject[token.key] = token.value
+        previousPropertyKey = token.key
+        if (token.inlineComment) {
+          const symbol = Symbol.for(`after:${token.key}`);
+          commentJsonObject[symbol] = (commentJsonObject[symbol] || []).concat([token.inlineComment])
+        }
+        break;
+      case 'CommentedOutProperty':
```
tmeyer2115, line 130: Do we handle the case of a line that's a `CommentedOutProperty` that also has an `inline-comment`?

postupgrade/mergeGlobalConfigs.js
oshi97: the regex matches on strings that looks like a json property, aka a comment that looks like 
```
// "token": "<REPLACE ME>", // The auth token to access Answers experience.
```
Being a comment token just means that it is some kind of comment, it doesn't say anything about what the actual string is like. For instance if we forgot the double quotes around "token". A user could also add their own random comments to the global config, which we've decided not to handle on upgrades.

postupgrade/mergeGlobalConfigs.js
oshi97: The order of the tokens is important, so I opted for an array. ES6 Maps preserve order by insertion order but I was worried about that being a point of failure in the future, say if we need to refactor this code and somebody isn't careful about insertion order.

postupgrade/mergeGlobalConfigs.js
oshi97: `CommentedOutProperty`s  don't know about whether they contain inline comments or not. All they know is that they have some string value. That string value might have what we recognize as an inline comment inside of it, but, currently, `CommentedOutProperty`s don't need to know about that. I don't think we would need to make large changes to add that information later on.

postupgrade/PostUpgradeHandler.js
```diff
@@ -51,7 +52,7 @@ class PostUpgradeHandler {
   async mergeThemeGlobalConfig(userGlobalConfigPath, themeGlobalConfigPath) {
     const updatedCommentJson = fs.readFileSync(themeGlobalConfigPath, 'utf-8');
     const originalCommentJson = fs.readFileSync(userGlobalConfigPath, 'utf-8');
-    return mergeGlobalConfigs(updatedCommentJson, originalCommentJson);
+    return mergeGlobalConfigs(parse(updatedCommentJson), parse(originalCommentJson));
```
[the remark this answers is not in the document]
oshi97: I missed this, I was looking at the global config in the theme during manual testing

postupgrade/mergeGlobalConfigs.js
```diff
@@ -0,0 +1,174 @@
...
+    tokens.push(parseCommentedOutProperty(c));
+  }
+
+  Object.keys(commentJsonObject).forEach(key => {
+    const beforeSymbol = Symbol.for(`before:${key}`)
+    const afterSymbol = Symbol.for(`after:${key}`)
+    const beforeComments = commentJsonObject[beforeSymbol] || []
+    const afterComments = commentJsonObject[afterSymbol] || []
+    beforeComments.forEach(handleComment)
+    const inlineComment = afterComments.find(c => c.inline);
```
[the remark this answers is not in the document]
oshi97: should be a bit clearer now though it's inherently a little low level

postupgrade/mergeGlobalConfigs.js
oshi97: tldr is yes it's handled correctly, since it's not a case we need to do anything about

postupgrade/mergeGlobalConfigs.js
oshi97: yeah we don't handle arbitrary comments

oshi97: We can also just ignore arbitrary comments and let the merge continue, which I didn't like yesterday but am starting to warm up to

tmeyer2115 (approved):

cea2aj (approved):
I'm a fan of the use of tokens here!

tmeyer2115 (approved):

PR #1034, opened 2022-01-14
close autocomplete on click in parent window

yen-tt:
close autocomplete on click in parent window by sending message from parent window to iframe using iFrameResizer library

in iframe-common.js, run by parent page, add an event listener to the document for click event. When trigger, check if the event is trusted (invoke by user, not external script) and send a message to iframe. In html.hbs, run by iframe, modify onMessage function in iframeResizer to check for message that contains event type 'click'. If so, get all active autocomplete components and invoke their close function.

J=SLAP-1182
TEST=manual

host a site in SGS with a page that uses iframe.js from localhost port (with the pr changes) for cross-origin test. Also modify page to have multiple search bar. See that autocomplete closes when click outside of iframe.

### DIFF
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -145,6 +145,10 @@
             HitchhikerJS.RuntimeConfigReceiver.handle(message.runtimeConfig);
           } else if (window.isOverlay) {
             window.Overlay.onMessage(message);
+          } else if (message.eventType && message.eventType === 'click') {
+            ANSWERS.components._activeComponents
+              .filter(component => component.name === 'SearchBar.autocomplete')
+              .forEach(autocompleteComponent => autocompleteComponent.close());
           }
         }
       };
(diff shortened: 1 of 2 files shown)


layouts/html.hbs
```diff
@@ -145,6 +145,10 @@
             HitchhikerJS.RuntimeConfigReceiver.handle(message.runtimeConfig);
           } else if (window.isOverlay) {
             window.Overlay.onMessage(message);
+          } else if (message.eventType && message.eventType === 'click') {
+            ANSWERS.components._activeComponents
```
tmeyer2115, line 149: We should probably make a method under `ANSWERS.components` called `getByName`. That way, we don't leak private, implementation details here. But, that's beyond this scope of this PR. I'll make a separate item to address.

cea2aj, line 149: @oshi97 or @tmeyer2115 , I'm wondering how you all feel about using `ANSWERS.components._activeComponents` since it's a private property. Alternatively, we could add a public function to the SDK which would allow us to get all of autocomplete components

tmeyer2115 (approved):

PR #1035, opened 2022-01-14
Integrate with Yext Sites

cea2aj:
Integrate with Yext Sites
 
Add config to support yext sites. The main "gotcha" is that Sites renames the answers-hitchhiker-theme to 'repo', so I added an extra jambo config which we can use with Yext Sites which refers the the theme as 'repo'.

Once we merge this into develop and master, we should be able to see those sites hosted at:
- theme.slapshot.pagescdn.com
- https://develop-theme-slapshot-pagescdn-com.preview.pagescdn.com/

The live preview integration is a little out of scope, but I think it would be useful if we add it in the future. I ran into an issue with it which may also have to do with repo renaming. It would allow us to open up the test site for any PR that we push to Github which would be useful for testing

J=SLAP-1679
TEST=manual

Connect this branch to Yext Sites and see the site hosted at https://devserveonyextsites-theme-slapshot-pagescdn-com.preview.pagescdn.com/

### DIFF
--- a/sites-config/ci_config.json
+++ b/sites-config/ci_config.json
@@ -1,0 +1,24 @@
+{
+  "artifactStructure": {
+      "assets": [
+          {
+              "root": "test-site/public"
+          }
+      ],
+      "hbsTemplates": []
+  },
+  "base_image_tag": "node-12.18",
+  "dependencies": {
+      "installDepsCmd": "npm install",
+      "requiredFiles": []
+  },
+  "buildArtifacts": {
+      "build_setup_cmd": "npm install && ./test-site/scripts/setup_for_yext_sites.sh",
+      "build_cmd": "npm run setup-test-site && npm run build-test-site"
+  },
+  "livePreview": {
+      "serve_setup_cmd": "npm run setup-test-site && npm run build-test-site",
+      "serve_cmd": "npx serve -l 8080 test-site/public"
+  },
+  "watch_cmd": "npm run watch"
+}
--- a/sites-config/ci_config.json
+++ b/sites-config/ci_config.json
@@ -17,7 +17,7 @@
       "build_cmd": "./test-site/scripts/setup_for_yext_sites.sh && npm run setup-test-site && npm run build-test-site"
   },
   "livePreview": {
-      "serve_setup_cmd": "npm run build-test-site",
+      "serve_setup_cmd": "mkdir -p /temp/theme && mv * /temp/theme && mkdir answers-hitchhiker-theme && mv /temp/theme/sites-config . && mv /temp/theme/* answers-hitchhiker-theme && npm run build-test-site",
       "serve_cmd": "npx serve -l 8080 test-site/public/ &"
   },
   "watch_cmd": "npm run watch"
--- a/sites-config/ci_config.json
+++ b/sites-config/ci_config.json
@@ -13,11 +13,11 @@
       "requiredFiles": []
   },
   "buildArtifacts": {
-      "build_setup_cmd": "npm install && ./test-site/scripts/setup_for_yext_sites.sh",
-      "build_cmd": "npm run setup-test-site && npm run build-test-site"
+      "build_setup_cmd": "npm install",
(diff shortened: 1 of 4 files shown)


sites-config/ci_config.json
```diff
@@ -0,0 +1,24 @@
+{
```
yen-tt, line 1: for my own understanding, how does SGS know to look into sites-config/ folder? I see other SGS sites have a top level `config.json`, and the fields in there would point to different scripts in the ci/ folder, to configure how the build, install, serve steps should look like

sites-config/ci_config.json
```diff
@@ -0,0 +1,24 @@
...
+      "installDepsCmd": "npm install",
+      "requiredFiles": []
+  },
+  "buildArtifacts": {
+      "build_setup_cmd": "npm install",
+      "build_cmd": "./test-site/scripts/setup_for_yext_sites.sh && npm run setup-test-site && npm run build-test-site"
+  },
+  "livePreview": {
+      "serve_setup_cmd": "echo serve setup cmd",
+      "serve_cmd": "echo serve cmd"
```
yen-tt, line 21: does live preview not work if this is something like this?
```
"serve_setup_cmd": "cd test-site/ && npx jambo build",
"serve_cmd": npx serve -l 8080 test-site/public/ &
```

sites-global/global.json
```diff
@@ -0,0 +1,3 @@
+{
+	"name": "Answers Hitchhiker Theme"
```
yen-tt, line 2: how does this file get use? does SGS merge this to another configuration?

test-site/jambo-yext-sites.json
```diff
@@ -0,0 +1,18 @@
...
+      "script/on-ready.js",
+      "cards",
+      "directanswercards"
+    ],
+    "preservedFiles": [
+      "public/iframe_test.html",
+      "public/overlay.html"
+    ]
+  },
+  "defaultTheme": "repo"
```
yen-tt, line 17: huh it doesn't just use the repo's name `answers-hitchhiker-theme`?

yen-tt, line 17: I remember running pwd before in SGS and I think it rename the top level folder to `/repo-vol`?

sites-config/ci_config.json
cea2aj: I'm not really sure about the details, other than the fact that the new system (Yext Sites) checks for the config file in the `sites-config` folder. The older system checks for the top level `config.json` (it can have other names as well). If we wanted to, we could create a ci folder just like the old setup

sites-global/global.json
cea2aj: I'm not exactly sure how it's used. I saw it in other repos on Yext Sites so I figured it wouldn't hurt to add, but I can look more into it

sites-config/ci_config.json
cea2aj: I can give it a shot

test-site/jambo-yext-sites.json
cea2aj: For full builds, it renames the top level folder to `repo`. For live preview, it looks like it uses 'repo-vol'. I could try making another file which refers to `repo-vol` for live preview

sites-config/ci_config.json
yen-tt: Oh I see. I recently created a yext site, but following the old system approach. Was wondering if the system just check for files both location somehow, or prioritize the files from new approach. But either setup is fine with me!

cea2aj: In Storm, when creating a site, it asks if you want an old version or a new one. I believe that's how it determines which config files to use

sites-config/ci_config.json
cea2aj: This doesn't work because of the directory rename and some other issues. I'll defer to another item per Tom's suggestion

yen-tt (approved):
nice! 👌

PR #1037, opened 2022-01-18
Add missing rtf styling in components

yen-tt:
- Add rtf styling to event-standard component
- move rtf styling from `&-details` to `&-detailsText` for faq-accordion and location-standard component.
  - update location-standard template to use non-html escape syntax (triple curly braces) when displaying details
 
J=SLAP-1804
TEST=manual

test event page, location_google page, and faq page with rich text description. See that the text is styled properly with rtf styling.

### DIFF
--- a/cards/location-standard/template.hbs
+++ b/cards/location-standard/template.hbs
@@ -178,11 +178,11 @@
 <div class="HitchhikerLocationStandard-details">
   {{#if showExcessDetailsToggle}}
     <div class="HitchhikerLocationStandard-detailsText js-HitchhikerCard-detailsText">
-      {{truncatedDetails}}
+      {{{truncatedDetails}}}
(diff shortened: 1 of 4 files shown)


cards/location-standard/template.hbs
```diff
@@ -182,7 +182,7 @@
     </div>
   {{/if}}
   <div class="HitchhikerLocationStandard-detailsText js-HitchhikerCard-detailsText{{#if showExcessDetailsToggle}} js-hidden{{/if}}">
-    {{card.details}}
+    {{{card.details}}}
```
tmeyer2115, line 185: I think we should leave this as is with two curly braces. Not all cards were meant to support RTF by default.

tmeyer2115 (approved):

PR #1040, opened 2022-01-26
add support for custom prompts based on parent window's page path

yen-tt:
this pr add support for custom prompts in autocomplete based on parent window's page path

Add new runtime config listener for key 'customPrompts' that will call autocomplete `setCustomPrompts` function when trigger. In parent's page, `window.AnswersExperienceFrame.runtimeConfig.set(..)` should be use to configure the custom prompts for specific page path.

J=SLAP-1803
TEST=manual

set customPrompts in searchbar configuration in index config, see that it appears in the dropdown.

setup two pages (/promo-search and /random-page) that loads an iframe for universal search page. in /promo-search page, set runtime config for custom prompts. See that search bar in /promo-search have custom prompts. See that /random-page have no custom prompts and work as usual.

confirmed customPrompts appear when using sdk asset from develop branch after the sdk pr is merged in.

NOTE: this pr requires [SDK changes](https://github.com/yext/answers-search-ui/pull/1655) to work. Confirmed with product that this feature doesn't have to be in theme 1.28, will merge when the next SDK version is release.


static/js/runtime-config-listeners/custom-autocomplete-prompts.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @type {import('../runtime-config.js').RuntimeConfigListener}
+ */
+ export default {
+  key: 'customPrompts',
+  callback: value => {
+    ANSWERS.components._activeComponents
```
oshi97, line 7: I think we'd rather not reach into ANSWERS.components if possible

yen-tt: I'm not sure of another way of getting the autocomplete instances without going through ANSWERS.components at the moment. The exposed function `getActiveComponent` from ANSWERS.components only provide one instance so I had to use the internal field `_activeComponents` to work. I could add another method `getActiveComponents` to ANSWERS.component or to top level ANSWERS to avoid this, thoughts on that?

oshi97: instead of directly calling methods on the SearchBar's autocomplete, could we add config to the SearchBar instead?

oshi97: after talking with yen, trying to have some config callback would probably be more trouble than its worth

tmeyer2115 (dismissed):

layouts/html.hbs
```diff
@@ -146,8 +146,8 @@
           } else if (window.isOverlay) {
             window.Overlay.onMessage(message);
           } else if (message.eventType && message.eventType === 'click') {
-            ANSWERS.components._activeComponents
-              .filter(component => component.name === 'SearchBar.autocomplete')
+             ANSWERS.components
```
oshi97, line 149: boop extra space before ANSWERS.components

layouts/html.hbs
```diff
@@ -146,8 +146,8 @@
...
+              .getActiveComponents('AutoComplete')
```
oshi97, line 150: In the name of writing observations that don't matter in the grand scheme of things, I think closing other AutoComplete usages besides SearchBar ones is the behavior we want! It will probably never come into play since we don't have FilterSearch or the GeolocationFilter (which is different than LocationBias) but it's nice nonetheless!

oshi97 (approved):

oshi97 (approved):

PR #1039, opened 2022-01-24
Add multilang doc standard card

nmanu1:
Add a multi-lang document standard card. Update help_articles.es.json to use this card for the test-site.

J=SLAP-1839
TEST=manual

Check that help articles on the es test-site use the new card and show translated text for feedback submission.

PR #1038, opened 2022-01-24
Update vertical-grid universalSectionTemplate

nmanu1:
Change the default `universalSectionTemplate` for vertical-grid to grid-three-columns.

J=SLAP-1831
TEST=manual

Check that the test-site shows the correct layout on the universal page for the people vertical.

PR #1043, opened 2022-02-08
v1.28

tmeyer2115:
### Features
- Multi-lang versions of the `document-standard` result card and the `documentsearch-standard` direct answer card have been added. (#1041, #1039)
- We automatically bump the `sdkVersion` in the `global_config` on upgrade. (#1031)

### Changes
- RTF styling was added to the `event-standard`, `faq-accordion`, and `location-standard` cards. (#1037)
- The default `universalSectionTemplate` for a `vertical-grid` page was updated to `grid-three-columns`. (#1038)

### Bugfixes
- When clicking outside the iFrame'd experience, the query suggestions dropdown now closes. (#1034)
- There was a bug with the custom `card` command. If the new card's name contained the name of the forked card, the former would not be added properly. This is now fixed. (#1036)
- Previously, if there was a Map on the Universal page, a user could get stuck on the Map while scrolling down the page. This has been fixed. (#1030, #1042)


oshi97:
For "Previously, if there was a Map on the Universal page, a user could get stuck on the Map while scrolling down the page. This has been fixed" I think we want to add #1030 to the PR list as the original PR

I would add that we now automatically bump a uglobal_config's `sdkVersion` on upgrade.
We could also mention that on upgrade the global_config's structure will better mirror the structure of the base global_config in the theme, though I'm also fine if we want to leave this bit out.

oshi97 (approved):

PR #1049, opened 2022-03-08
Change ID used for map pins

nmanu1:
Use a result's internal `uid` instead of the external entity `id` for identifying map pins. The `id` can be any string specified by the user. This causes an error when clicking on a map pin on mobile if the `id` has spaces, apostrophes, etc. The `uid` is a unique, automatically-generated string that only consists of numbers, so it would not have this problem.

J=SLAP-1962
TEST=manual

Check that map pins on mobile can now be clicked regardless of external entity ID because the `uid` is used instead.

### DIFF
--- a/static/js/theme-map/VerticalFullPageMapOrchestrator.js
+++ b/static/js/theme-map/VerticalFullPageMapOrchestrator.js
@@ -498,7 +498,7 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
 
       const entityId = cardId.replace('js-yl-', '');
       const verticalResults = this.core.storage.get(StorageKeys.VERTICAL_RESULTS).results;
-      const entityData = verticalResults.find(entity => entity.id.toString() === entityId);
+      const entityData = verticalResults.find(entity => entity._raw.uid.toString() === entityId);
       const opts = {
         parentContainer: this._container, 
         container: `.yxt-Card-${entityId}`,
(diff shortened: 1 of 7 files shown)


static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -498,7 +498,7 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
 
       const entityId = cardId.replace('js-yl-', '');
```
oshi97, line 499: I think we'll want to rename this from entityId, entityId has a specific meaning inside the knowledge graph. For instance if you go to a specific entity the url will look like `https://www.yext.com/s/<business id>/entity/edit3?entityIds=<entityId>`

nmanu1: that URL `<entityId>` is actually the same as this `uid`. there are 2 values that are referred to as "entityId". the internal one is the one included in the entity url and is returned in the response data as `uid`. the external one is the one the user can change when editing an entity and is returned as the `id`

oshi97: ah ok so it was actually "kind of" misnamed before?

oshi97 (approved):

static/js/theme-map/VerticalFullPageMapOrchestrator.js
nmanu1: yes, it was a little misleading. in alpha, the `uid` is the `entityId` and the `id` is `externalId`

PR #1055, opened 2022-03-16
fix Map errors when the map container does not have offsetWidth/Height

oshi97:
If the full-page-map's container has an offsetWidth or offsetHeight of 0,
which essentially means that the map is not visible, the page will throw an error
and not render the map. This PR updates the map to gracefully handle these cases.

T=4215
TEST=manual

saw that the user's experience on the aforementioned techops ticket
no longer logs `'NaN' must be convertible to 'Number'`, and that
changing the page's styling so that the map becomes visible renders the map

### DIFF
--- a/static/js/theme-map/Maps/Map.js
+++ b/static/js/theme-map/Maps/Map.js
@@ -514,22 +514,34 @@ class Map {
   }
 
   getVisibleCenter() {
-    const visibleBounds = this.getVisibleBounds();
+    const visibleBounds = this._getVisibleBounds();
+    if (!visibleBounds) {
+      return null;
+    }
+
     const center = visibleBounds.getCenter();
     return center;
   }
 
   getVisibleRadius() {
-    const visibleBounds = this.getVisibleBounds();
+    const visibleBounds = this._getVisibleBounds();
+    if (!visibleBounds) {
+      return 0;
+    }
+
     return 1000 * visibleBounds.ne.distanceTo(this.getVisibleCenter(), Unit.KILOMETER);
   }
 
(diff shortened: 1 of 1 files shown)


static/js/theme-map/Maps/Map.js
```diff
@@ -552,7 +564,8 @@ class Map {
   }
 
   coordinateIsInVisibleBounds(coordinate) {
-    return this.getVisibleBounds().contains(coordinate);
+    const bounds = this._getVisibleBounds();
+    return bounds ? bounds.contains(coordinate) : false;
   }
 
   setCenterWithPadding(coordinate, animated = false) {
```
yen-tt, line 571: do we need to guard pixelHight and pixelWidth in this function as well?

oshi97: Isn't it being guarded by the ternary op?

yen-tt: which ternary op? I didn't look too deep into the rest of the code, but I see that you guarded pixelHeight/width in setBounds(), I assume setCenterWithPadding() would do something with pixelHeight/width that may cause similar error so the same guard may be needed here

oshi97:
Closing this since the customer's issue wasn't that the map wasn't working when it wasn't visible on the page, but the fact that the map wasn't visible at all. I don't think this is actually an issue we need to solve, or that we can really solve satisfactorily without spending a time cleaning out the full page map's interiors. Version 1.0 of the theme is planned for deprecation, so my gut feeling is that's not a good usage of time.

PR #1059, opened 2022-04-08
Empty commit to regenerate percy snapshots

oshi97:
Apparently, Percy will completely reset your snapshots state when
it automatically updates its browser versions. This is a little annoying.

This commit regenerates them so that the next PR into develop can still
get a diff of what has changed, as opposed to percy saying that EVERYTHING
has changed.

J=NONE
TEST=none


oshi97:
woof 
image

nmanu1 (approved):

PR #1060, opened 2022-04-08
support rtf truncation

oshi97:
This commit adds a new truncatedDetails property to the showMoreDetails
portion of dataForRender's return value. This property is intended to be used
for truncating rich text card details, though it is not necessarily limited to that.

Updated all cards that use ANSWERS.formatRichText by default to have truncatedDetails
instead of showMoreLimit in their showMoreDetails. We are not removing showMoreLimit.

Added a try catch to our percy snapshots that will exist on error, otherwise it looks like the node process
hangs forever. The snapshots still run on node 12.18, I think upgrading them to 16 also fix this issues,
since unhandled promise rejections automatically kill the process in later node versions. However,
the promise rejections would still be unhandled without the try catch. I didn't upgrade to 16 since people may still be using on 12.18.

J=SLAP-2039
TEST=manual,auto

updated the products test-site page to have rtf truncation

test non rtf truncation using showMoreLimit

cards with no details render correctly

### DIFF
--- a/cards/card_component.js
+++ b/cards/card_component.js
@@ -99,7 +99,6 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
     if (userSpecifiedTruncatedDetails) {
       const showExcessDetailsToggle = userSpecifiedTruncatedDetails.length + 3 < details.length
       const truncatedDetails = showExcessDetailsToggle ? userSpecifiedTruncatedDetails : '';
-      console.log(truncatedDetails.length, details.length)
       return {
         showExcessDetailsToggle,
         truncatedDetails
@@ -108,12 +107,15 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
 
     // Set the value of excessDetailsToggle. Note that it is not enough to have a showMoreLimit.
     // The card's details must extend past this limit as well for the toggling to be enabled.
-    const showExcessDetailsToggle = showMoreLimit && (details.length + 3) > showMoreLimit;
+    const showExcessDetailsToggle = showMoreLimit && (details.length + 3 > showMoreLimit);
     
     const truncatedDetails = showExcessDetailsToggle
       ? `${details.substring(0, showMoreLimit)}...`
       : '';
-    return truncatedDetails;
+    return {
+      showExcessDetailsToggle,
+      truncatedDetails
+    };
   }
 
   validateDataForRender(data) {
--- a/cards/card_component.js
+++ b/cards/card_component.js
@@ -80,30 +80,40 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
     
     cardData.feedbackEnabled = ANSWERS.getAnalyticsOptIn() && cardData.feedback;
 
-    const { details, showMoreDetails } = cardData;
-
-    const cardDetails = details || '';
-    const cardShowMoreConfig = showMoreDetails || {};
-    const { showMoreLimit } = cardShowMoreConfig;
-
-    // Set the value of excessDetailsToggle. Note that it is not enough to have a showMoreLimit.
-    // The card's details must extend past this limit as well for the toggling to be enabled.
-    const showExcessDetailsToggle = showMoreLimit && cardDetails.length > showMoreLimit;
-
-    const truncatedDetails = showExcessDetailsToggle
-      ? `${cardDetails.substring(0, showMoreLimit)}...`
-      : '';
+    const { showExcessDetailsToggle, truncatedDetails } = this._getTruncatedDetails(cardData);
     
     this.validateDataForRender(cardData);
 
     return super.setState({
       ...data,
       card: cardData,
-      showExcessDetailsToggle: showExcessDetailsToggle,
-      truncatedDetails: truncatedDetails,
+      showExcessDetailsToggle,
+      truncatedDetails,
       cardName: `{{componentName}}`,
       relativePath: `{{relativePath}}`
     });
+  }
+
+  _getTruncatedDetails({ details, showMoreDetails }) {
+    const { showMoreLimit, truncatedDetails: userSpecifiedTruncatedDetails } = showMoreDetails || {};
+    if (userSpecifiedTruncatedDetails) {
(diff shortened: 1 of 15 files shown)


oshi97:
created a new PR because even after merging in https://github.com/yext/answers-hitchhiker-theme/pull/1059 percy is still saying it's updated all the browsers, so I was thinking maybe it needs to be a fresh PR after regenerating the snapshots
image

cards/card_component.js
```diff
@@ -80,32 +80,44 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
...
+      truncatedDetails,
       cardName: `{{componentName}}`,
       relativePath: `{{relativePath}}`
     });
   }
 
+  _getTruncatedDetails({ details = '', showMoreDetails }) {
+    const { showMoreLimit, truncatedDetails: userSpecifiedTruncatedDetails } = showMoreDetails || {};
+    if (userSpecifiedTruncatedDetails) {
+      const showExcessDetailsToggle = userSpecifiedTruncatedDetails.length + 3 < details.length
```
cea2aj, line 100: Is the purpose of the `+ 3` to prevent situations where the truncated details is only hiding 1 to 3 characters? That way the ellipsis are always hiding at least 4 characters?

cards/faq-accordion/component.js
```diff
@@ -21,7 +21,7 @@ class faq_accordionCardComponent extends BaseCard['faq-accordion'] {
       // If the card's details are longer than a certain character count, you can truncate the
       // text. A toggle will be supplied that can show or hide the truncated text.
       // showMoreDetails: {
-      //   showMoreLimit: null, // Character count limit
+      //   truncatedDetails: profile.answer ? ANSWERS.formatRichText(profile.answer, "answer", linkTarget, 500) : null, // The truncated rich text
       //   showMoreText: '', // Label when toggle will show truncated text
```
yen-tt, line 25: If showMoreLimit is still an option, maybe we should still include the commented out showMoreLimit option in the cards? Or is this what product wanted

cards/card_component.js
oshi97: right it's to prevent the truncated text + ellipsis from being longer than the original text

cards/faq-accordion/component.js
oshi97: Rose said to get rid of it

cards/card_component.js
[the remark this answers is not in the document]
oshi97: it's to prevent the truncated text + ellipsis from being longer than the original text!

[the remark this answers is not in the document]
oshi97: It makes the assumption that the user wants the suffix to be `...`, which I was a little hesitant about at first since the suffix can be configured in ANSWERS.formatRichText(). However we already force the user to have the suffix be `...` for regular truncated text, so I figured it's probably ok.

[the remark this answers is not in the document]
cea2aj: Would an variable such as `const ellipsesLength` help make the purpose of the number clearer?

cards/faq-accordion/component.js
oshi97: I also think that it's confusing having both

cards/card_component.js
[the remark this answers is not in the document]
yen-tt: I thought doing  ANSWERS.formatRichText(blah..) would already include the suffix ellipsis in the string so do we need to do + 3? Is it fine that we assume the suffix will always be the ellipsis (length 3), which is commonly use but we do let user provide their custom suffix?

cards/card_component.js
```diff
@@ -80,32 +80,44 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
...
-      truncatedDetails: truncatedDetails,
+      showExcessDetailsToggle,
+      truncatedDetails,
       cardName: `{{componentName}}`,
       relativePath: `{{relativePath}}`
     });
   }
 
+  _getTruncatedDetails({ details = '', showMoreDetails }) {
+    const { showMoreLimit, truncatedDetails: userSpecifiedTruncatedDetails } = showMoreDetails || {};
```
cea2aj, line 98: When I see `Details` at the end of a variable, my mind thinks it's an object, but it looks like it's actually a string or the HTML. What do you think of renaming to `truncatedText`?

oshi97: I think I agree with you, but then I would want to rename it everywhere, including in the card templates. Is it ok if I just add a jsdoc type comment?

cea2aj: ya, that works

cards/card_component.js
```diff
@@ -94,14 +95,17 @@ BaseCard["{{componentName}}"] = class extends ANSWERS.Component {
...
-  _getTruncatedDetails({ details = '', showMoreDetails }) {
-    /**
-     * @type {{
-     *   showMoreLimit?: number,
-     *   truncatedDetails?: string
-     * }}
-     */
-    const { showMoreLimit, truncatedDetails: userSpecifiedTruncatedDetails } = showMoreDetails || {};
+  /**
+   * Returns the whether to render the excess details toggle, and
```
nmanu1, line 99: nit: extra "the"

cards/multilang-product-standard/component.js
```diff
@@ -33,7 +33,7 @@ class multilang_product_standardCardComponent extends BaseCard['multilang-produc
       // text. A toggle will be supplied that can show or hide the truncated text.
       // Note: If you are using rich text for the details, you should not enable this feature.
       // showMoreDetails: {
-      //   showMoreLimit: 350, // Character count limit
+      //   truncatedDetails: profile.richTextDescription ? ANSWERS.formatRichText(profile.richTextDescription, 'richTextDescription', linkTarget, 350) : null, // The truncated rich text
```
yen-tt, line 36: just double checking, do we want truncatedDetails for all cards? I see some cards like menuitem-standard and link-standard doesn't have this update (a quick search with '// Character count limit' shows 16 card files with showMoreLimit)

cards/card_component.js
oshi97: updated!

cea2aj (approved):

PR #1065, opened 2022-04-28
Fix text truncation in accordion card

nmanu1:
This PR adds an event listener specifically for the details toggle in accordion cards to update the height of the content along with toggling the visible text. The details toggle for other cards will remain as is. This change now expands/shrinks the height of the card to match the expanding/shrinking of the details text.

Since the details text is switched without animation, it looked the most natural to have the expanding behavior also not have any animation. This is accomplished by setting the `height` to `auto` to allow the `height` to be recalculated after toggling the displayed text. But, to ensure that the animation behavior for collapsing/expanding the entire accordion card remains as is, the `height` is then set to the actual, recalculated `scrollHeight` value so it can be used for the animation calculations.

J=SLAP-2047
TEST=manual

Test `faq-accordion` and `multilang-faq-accordion` cards with details truncation enabled and see that toggling the details updates the card height instantly, but collapsing/expanding the FAQ keeps it's animation behavior.

### DIFF
--- a/cards/faq-accordion/component.js
+++ b/cards/faq-accordion/component.js
@@ -111,6 +111,19 @@ class faq_accordionCardComponent extends BaseCard['faq-accordion'] {
       }
     });
 
+    const showExcessDetailsToggleEls = this._container.querySelectorAll('.js-HitchhikerFaqAccordion-detailsToggle');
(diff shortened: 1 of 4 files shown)


oshi97:
can we update the test-site's faq card to have text truncation?

oshi97 (approved):

PR #1067, opened 2022-05-04
keep parent's domain for tab navigation on new tab in iframe experience

yen-tt:
This pr ensures tab navigation open to a new tab in an iframe experience should still be under the domain of the iFrame's parent for the new page's url. 

After ANSWERS is initialized or in onReady stage, when iframeLoaded promise is resolved - indicating this is an iframe experience - a message is sent to the parent (in iframe-messaging.js) to send its url to the iframe telling it to update its tab navigation (iframe-common.js). 

In the iframe's onMessage hook, `setParentUrl` is invoked for every navigation component on the page to properly update the tabs' link to navigate under the domain of the iframe's parent. This require changes in [SDK](https://github.com/yext/answers-search-ui/pull/1731) pr. acceptance test in this pr will fail until SDK changes is merge to develop

J=SLAP-2089
TEST=manual

Tested along with the changes in theme. 
Deployed two test sites in storm, with [one being from the theme dev branch](https://devtabnaviniframe-theme-slapshot-pagescdn-com.preview.pagescdn.com/?query=virginia&referrerPageUrl=) that points to the SDK pr dev branch, under the staging commit: dev/tab-nav-in-iframe[@65f09b10](https://github.com/yext/answers-hitchhiker-theme/commit/65f09b101f882c597389717d4a8b8d814edb1bda). That is used in the src script of iframe_test.html of the [other site](https://devyttesting-theme-slapshot-pagescdn-com.preview.pagescdn.com/?query=virginia&referrerPageUrl=) under the staging commit: dev/yt-testing[@517e852f](https://github.com/yext/answers-hitchhiker-theme/commit/517e852fe3ff2f7961704090e11118273ba162b3). Navigate to different tabs in the same tab and opening them in new tab. See that the page is still under the domain of the iFrame's parent.

### DIFF
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -149,26 +149,10 @@
             ANSWERS?.components
               .getActiveComponents('AutoComplete')
               .forEach(autocompleteComponent => autocompleteComponent.close());
-          } else if (message.action ===  'update-tab-navigation') {
-            if (!ANSWERS) {
-              return;
-            }
-            function updateTabsLink () {
-              const parentUrlWithoutParams = message.parentUrl.split('?')[0];
-              const navItems = document.getElementsByClassName('js-yxt-navItem');
-              if (navItems) {
-                Array.from(navItems).forEach(navItem => {
-                  const verticalUrl = navItem.pathname.replace(/^\//, '');
-                  const params = navItem.search ? navItem.search : '?';
-                  navItem.href = parentUrlWithoutParams + params + '&verticalUrl=' + verticalUrl;
-                  navItem.target = '_parent';
-                })
-              }
-            }
-            updateTabsLink();
-            ANSWERS.components
+          } else if (message.action === 'update-tab-navigation') {
+            ANSWERS?.components
               .getActiveComponents('Navigation')
-              .forEach(navigation => navigation.userOnMount = () => updateTabsLink())
+              .forEach(navigation => navigation.setParentUrl(message.parentUrl))
           }
         }
       };
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -152,7 +152,7 @@
           } else if (message.action === 'update-tab-navigation') {
             ANSWERS?.components
               .getActiveComponents('Navigation')
-              .forEach(navigation => navigation.setParentUrl(message.parentUrl))
+              .forEach(navigationComponent => navigationComponent.setParentUrl(message.parentUrl));
           }
         }
       };
--- a/layouts/html.hbs
+++ b/layouts/html.hbs
@@ -146,9 +146,31 @@
           } else if (window.isOverlay) {
             window.Overlay.onMessage(message);
           } else if (message.eventType && message.eventType === 'click') {
+            if (ANSWERS) {
+              ANSWERS.components
+                .getActiveComponents('AutoComplete')
+                .forEach(autocompleteComponent => autocompleteComponent.close());
+            }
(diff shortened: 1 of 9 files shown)


layouts/html.hbs
```diff
@@ -146,9 +146,31 @@
...
+            }
+            function updateTabsLink () {
+              const parentUrlWithoutParams = message.parentUrl.split('?')[0];
+              const navItems = document.getElementsByClassName('js-yxt-navItem');
+              if (navItems) {
+                Array.from(navItems).forEach(navItem => {
+                  const verticalUrl = navItem.pathname.replace(/^\//, '');
+                  const params = navItem.search ? navItem.search : '?';
+                  navItem.href = parentUrlWithoutParams + params + '&verticalUrl=' + verticalUrl;
+                  navItem.target = '_parent';
```
cea2aj, line 166: I would expect to find the code which defines the navItem hrefs inside the navigation component. If someone is reading the navigation component code, it might be confused to because the navigation component code doesn't include the parentURL, however in practice, it does. They probably wouldn't know to look here to find the code which updates the URLs.

Is it possible to add a function to the Navigation component which would allow you to set the parent URL? That way the Navigation component could take care of updating the URLs.

layouts/html.hbs
```diff
@@ -146,9 +146,31 @@
           } else if (window.isOverlay) {
             window.Overlay.onMessage(message);
           } else if (message.eventType && message.eventType === 'click') {
+            if (ANSWERS) {
+              ANSWERS.components
```
cea2aj, line 150: Since this code is wrapped in the #babel HBS helper, we can use the `ANSWERS?.` syntax since it will be transpiled in the bundle

oshi97:
I would probably remove the "fix" part from the title of the PR? This feels more like a feature than a bugfix - imo things were working as expected. Maybe just me though

layouts/html.hbs
oshi97: We could probably add it somewhere to the `addComponent` call for the Navigation component  (templates/common-partials/script/navigation.hbs), maybe in an `onMount`? I think adding new config to Navigation might also be a good choice if doing this purely in the theme gets too hacky, though.

script/iframe-messaging.js
```diff
@@ -13,4 +13,8 @@ ANSWERS.core.storage.registerListener({
       sessionId: ANSWERS.core.getOrSetupSessionId()
     }));
   }
-});
\ No newline at end of file
+});
+
+window.parentIFrame.sendMessage(JSON.stringify({
```
cea2aj, line 18: iI looks like this sends a message from the iframe to the parent, and then back to the iframe. Is this so that we know that Answers is initialized by the time we update the URLs? If so, I'm wondering if it makes sense to change this message to something like:
```
{ 
    body: 'answers-initialized'
}
```
That way the iframe-common code can send the update-tab-navigation back to the iframe once it gets the answers-initialized message.

yen-tt: yes this is to indicate that answers has been initialized so we can update the URL in iframeresizer. Sure, I can update the message action!

layouts/html.hbs
yen-tt: Having it in the component navigation is tricky when it comes to making sure iframe resizer are initialized so we can get the parent url before updating the tab links. I made changes to SDK and updated this pr to invoke a function to set up the parent url!

oshi97 (approved):

PR #1073, opened 2022-05-27
Fix Snyk vulnerabilities

nmanu1:
This PR fixes some of the Snyk vulnerabilities in the top-level package-lock.json. Specifically, it addresses some or all of the vulnerabilities from the following packages by upgrading their patch or minor versions: `ansi-regex`, `async`, `handlebars`, `urijs`, `ajv`, `underscore.string`, `glob-parent`, `postcss`, `node-fetch`, `tmpl`, `lodash`, `json-schema`, `moment`, `hosted-git-info`, `path-parse`, `prompts`, `ws`, `uglify-js`, `istanbul-reports`, and `minimist`.

J=none
TEST=manual

After this PR is merged into develop, will check to make sure the vulnerabilities are no longer in Snyk.


yen-tt (approved):

package.json
```diff
@@ -36,21 +36,21 @@
...
     "loader-utils": "^2.0.0",
     "locale-currency": "0.0.2",
     "lodash.clonedeep": "^4.5.0",
-    "postcss": "^8.2.1",
+    "postcss": "^8.4.14",
     "puppeteer": "^10.2.0",
     "serve": "^11.3.2",
     "simple-git": "^2.24.0",
-    "testcafe": "^1.18.4",
+    "testcafe": "^1.18.6",
```
oshi97, line 50: could we use 1.19 for testcafe?

oshi97 (approved):

package.json
nmanu1: it looks like v1.19 requires node >= 14. since we still want to support node 12 in the Theme, I'll leave this as is

PR #1088, opened 2022-08-16
automatically initialize when IS_DEVELOPMENT_PREVIEW is true

oshi97:
J=SLAP-2295
TEST=manual

see that if I have initializeManually true in the test-site's config,
without this code it will not initialize, with it stuff will be automatically
initialized.

### DIFF
--- a/hooks/templatedataformatter.js
+++ b/hooks/templatedataformatter.js
@@ -27,7 +27,7 @@ module.exports = function (pageMetadata, siteLevelAttributes, pageNameToConfig) 
     relativePath,
     env: {
       JAMBO_INJECTED_DATA: env.JAMBO_INJECTED_DATA,
-      IS_DEVELOPMENT_PREVIEW: !!env.IS_DEVELOPMENT_PREVIEW,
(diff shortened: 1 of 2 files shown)


hooks/templatedataformatter.js
```diff
@@ -27,6 +27,7 @@ module.exports = function (pageMetadata, siteLevelAttributes, pageNameToConfig)
     relativePath,
     env: {
       JAMBO_INJECTED_DATA: env.JAMBO_INJECTED_DATA,
+      IS_DEVELOPMENT_PREVIEW: !!env.IS_DEVELOPMENT_PREVIEW,
```
yen-tt, line 30: `!!` could be incorrect if somehow IS_DEVELOPMENT_PREVIEW set to "false" right? Maybe IS_DEVELOPMENT_PREVIEW === 'true' is better

oshi97: ah yeah that's true

oshi97: it's weird we used a string, I think looking back I would have asked it to be 0 or 1

yen-tt (approved):

PR #1101, opened 2022-10-03
Fix full page map keyboard on android, and set current center correctly

oshi97:
This PR fixes a bug where, on Android, it was not possible to run searches on the full page map, because
clicking into the keyboard would trigger a "searchOnMapMove", which would cause new search results to come in
and cause the keyboard to collapse itself.

The first issue is that, the android keyboard shifts all content on the page up, unlike the iOS keyboard, which does
not affect page content. This would cause the center of the map to move, which would cause searchOnMapMove
to be triggered. This was dealt with by fixing the height of the mapWrapper on page load for android devices.

The second issue was that the VerticalFullPageMapOrchestrator would ALWAYS have the incorrect map center.
This is because it would grab the center of the map before the map had actually loaded, resulting in it thinking the
center of the map was always the `this.defaultCenter` value, when it is virtually never that. This would make
it so that even with the first issue being patched, the keyboard would still always collapse itself the first time
you clicked into it, since the VerticalFullPageMapOrchestrator would not have the correct center until after that.
To patch this I added a new loadHandler to the map, and update the center in that handler.

J=TECHOPS-6738
TEST=manual

test on browserstack android chrome that I can click into the searchbar on the full page map, for both google and mapbox, and no search will be triggered and the keyboard will not hide itself and refuse to be typed into

test that locally on desktop the page looks fine on both google and mapbox


static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -193,9 +203,43 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
+    this.setFixedHeightsOnAndroid()
+  }
+
+  /**
+   * On Android browsers, opening up the keyboard will shift the contents of the entire page up,
+   * moving the map center, and thereby causing a searchOnMapMove to be triggered.
+   * The search response would then cause the page to update,
+   * and close the keyboard, making it impossible to actually type anything into the searchbar.
+   * 
+   * Setting a fixed height on elements like .Answers-mapWrapper prevents the keyboard from shifting the content
```
nmanu1, line 217: is this only a problem for the Theme's map?

static/js/theme-map/Maps/ProviderMap.js
```diff
@@ -44,6 +45,17 @@ class ProviderMapOptions {
...
   }
 
+  /**
+   * @typedef ProviderMap~loadHandler
+   * @function
+   */
+
+  withLoadHandler(handler) {
+    assertType(handler, Type.FUNCTION);
+    this.loadHandler = handler
```
nmanu1, line 55: nit: missing semicolon here and in other places

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -193,9 +203,43 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
+    }
+
+    setFixedHeight('.Answers-mapWrapper')
+
+    function getSingleElement(selector) {
+      const els = document.querySelectorAll(selector)
+      if (els.length === 0) {
+        console.error(`No ${selector} found, unable to set fixed height.`)
+      } else if (els.length > 1) {
+        console.error(`Multiple elements of for ${selector} found, expected only 1, not setting fixed height.`)
```
nmanu1, line 232: nit: "of for"

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -193,9 +203,43 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
+      } else {
+        const el = els[0]
```
nmanu1, line 234: can `els[0]` be returned directly here?

static/js/theme-map/Maps/Providers/Google.js
```diff
@@ -80,6 +80,9 @@ class GoogleMap extends ProviderMap {
     google.maps.event.addListener(this.map, 'click', () => {
       this._canvasClickHandler();
     });
+    google.maps.event.addListenerOnce(this.map, 'tilesloaded', () => {
```
yen-tt, line 83: is there a reason we only want to listen to the `tilesloaded` event only once? I assume every new search would re-load the tiles and trigger this event, do we not want to execute `loadHandler` function those times?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -193,9 +203,43 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
+    if (!this.isMobile() || !/Android/i.test(navigator.userAgent)) {
+      return
+    }
+
+    setFixedHeight('.Answers-mapWrapper')
+
+    function getSingleElement(selector) {
+      const els = document.querySelectorAll(selector)
+      if (els.length === 0) {
+        console.error(`No ${selector} found, unable to set fixed height.`)
```
yen-tt, line 230: nit: can we clarify a bit more in the console errors that it's map related?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
yen-tt, line 232: The issue still occurs with multiple map elements on page right? Can we not set fixed height for all even if there are multiple map wrapper elements found?

static/js/theme-map/ThemeMap.js
```diff
@@ -153,6 +153,9 @@ class ThemeMap extends ANSWERS.Component {
         this.config.zoomEndListener(this.map.getZoom(), zoomTrigger);
       });
       map.setCanvasClickHandler(() => this.config.canvasClickListener());
+      map.setLoadHandler(() => {
+        this.core.storage.set(StorageKeys.MAP_LOADED, true)
```
yen-tt, line 157: I guess there's no use to track loading event and set this back to false?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
oshi97: The full page map should never have multiple maps. If you're doing something custom with the map I didn't want to make assumptions

static/js/theme-map/VerticalFullPageMapOrchestrator.js
oshi97: Sure, you can always check the stack trace for the error but I don't think it hurts

static/js/theme-map/Maps/Providers/Google.js
oshi97: the load handler is only meant to run once when the map loads

static/js/theme-map/VerticalFullPageMapOrchestrator.js
oshi97: yes searchOnMapMove is only a thing in the full page map

static/js/theme-map/ThemeMap.js
oshi97: MAP_LOADED is only meant to trigger once if that's what you're asking

yen-tt: ah I see, I was thinking "load" event more of it loading tiles when the map is moved. Okay!

static/js/theme-map/Maps/Providers/Mapbox.js
```diff
@@ -51,6 +51,9 @@ class MapboxMap extends ProviderMap {
         this._canvasClickHandler();
       }
     });
+    this.map.on('load', () => {
```
yen-tt, line 54: is Google and Mapbox the only two map providers we want to fix?

oshi97: Yes we do not support other providers.

yen-tt (approved):

static/js/theme-map/Maps/Providers/Mapbox.js
oshi97: The other providers are facades and should have been ripped out but the creator of the full page map never had the time

nmanu1 (commented):
in the techops, there seemed to be a problem with a couple different types of phone. were they all android?

static/js/theme-map/Maps/Providers/Mapbox.js
oshi97: > in the techops, there seemed to be a problem with a couple different types of phone. were they all android?

Yes

oshi97:
> in the techops, there seemed to be a problem with a couple different types of phone. were they all android?

yes this is an android problem

PR #1112, opened 2022-11-30
Remove Browserstack acceptance testing on IE11

nmanu1:
Now that the Google Maps JS API has dropped support for IE11, Browserstack acceptance tests on IE11 fail. This PR updates the Browserstack acceptance workflow to only run the jobs on Safari and Firefox, and not on IE11 anymore. The `IE11NoCacheHook` is also removed as it's no longer needed.

J=SLAP-2502
TEST=auto

See that there are only Browserstack checks for Safari and Firefox on the `Test` commit in this PR and neither has Google Maps API errors.


oshi97:
do we also want to remove `registerIE11NoCacheHook` and it's usages?

nmanu1:
> Could we remove `registerIE11NoCacheHook` and it's usages? I think we could also remove it in the SDK but it's not really important either way just a small nicety.

Good call, removed! I'll make another PR in the SDK to remove it from there as well

oshi97 (approved):

PR #1115, opened 2022-12-09
fix iframe.js breaking links with hash URLs

oshi97:
Previously, iframe.js would not work with links contains hash URLs, i.e. links like

```html
<a id="skip-to-content-link" href="#main-content">Skip to content</a>
```

notice the URL starts with a hashtag. These links are different than normal HTML links,
in that instead of redirecting to a new page, they will scroll the browser down to the element
with the id matching the link. For the example above, it would scroll down to an element like

```html
<div id="main-content">main content!</div>
```

However, iframe.js has an issue where, when one of these links are clicked, the iframe.js will reload
and the parent frame's URL will be changed.
This issue was caused by the following process:
1. iframe.js's has a window.popstate listener, which reloads the iframe
whenever a popstate event was triggered (for instance by clicking on one of these links),
2. clicking a hash link will trigger this listener, reloading the iframe
3. reloading the iframe causes the inner iframe to send messages to the outer frame
4. these messages are received by iframe.js's onMessage listener, which then performs a history.replaceState (for certain types of messages)

To avoid this situation, this PR updates the popstate listener to be a no-op if the new URL of the page is the same as the old one, ignoring any hash URLs

J=TECHOPS-7401
TEST=manual

ran the theme's test site, created a basic iframe implementation
before this change, hash urls would have the broken behavior
after the change, they would scroll down the page without reloading the iframe or changing the parent's URL


test-site/.gitignore
```diff
@@ -13,6 +13,8 @@ webpack-config.js
 !config/global_config.json
 !config/locale_config.json
 !config/index.json
+!config/index.ar.json
```
oshi97, line 16: I deleted these pages to speed up the testing loop. When I added them back I noticed they were missing from the gitignore. They were probably being tracked in a similar way to `git add --force`

static/js/iframe-common.js
```diff
@@ -135,6 +131,26 @@ export function generateIFrame(domain, answersExperienceFrame) {
   }, '#answers-frame');
 }
 
+function registerPopStateListener() {
+  let previousLocationHref = window.location.href
+
+  /**
+   * Reloads the iframe with a recalculated src URL.
+   * Does not reload the iframe if the URL has only changed its hash param.
```
nmanu1, line 139: so we'd still want regular query params to trigger a reload?

static/js/iframe-common.js
```diff
@@ -135,6 +131,26 @@ export function generateIFrame(domain, answersExperienceFrame) {
...
+  let previousLocationHref = window.location.href
+
+  /**
+   * Reloads the iframe with a recalculated src URL.
+   * Does not reload the iframe if the URL has only changed its hash param. 
+   */
+  function popStateListener() {
+    /** @param {string} url  */
+    function getURLWithoutHash(url) {
+      return url.split('#')[0]
```
nmanu1, line 144: nit: missing semicolon here and below

static/js/iframe-common.js
oshi97: We don't want to touch the preexisting behavior because we're not sure how it's being used or for what purpose it's there. This code was copy and pasted all at once from the AEB (answers experience builder created by Mcgin) and it doesn't seem possible to track that info down.

nmanu1 (approved):

PR #1118, opened 2022-12-13
chore: update packages

tatimblin:
upgrade express (minor), serve (major), puppeteer (major), and loader-utils (patch) to resolve various non-critical sub-dependency vulnerabilities in snyk. Ran `npm audit fix` to to resolve various other vulnerabilities called out in npm.

TEST=manual

Served test-site without errors to confirm express and serve versions are compatible. Successfully ran tests to ensure new puppeteer version is compatible


package.json
```diff
@@ -31,31 +31,31 @@
   },
   "homepage": "https://github.com/yext/answers-hitchhiker-theme#readme",
   "devDependencies": {
```
nmanu1, line 33: do we also need to make any updates to the `static/package.json`?

package.json
```diff
@@ -31,31 +31,31 @@
...
     "libphonenumber-js": "^1.9.6",
-    "loader-utils": "^2.0.0",
+    "loader-utils": "^2.0.4",
     "locale-currency": "0.0.2",
     "lodash.clonedeep": "^4.5.0",
     "postcss": "^8.4.14",
-    "puppeteer": "^10.2.0",
-    "serve": "^11.3.2",
+    "puppeteer": "^13.7.0",
+    "serve": "^14.1.2",
```
nmanu1, line 58: I think this `serve` version drops support of node 12. I assume we're fine with that since we've stopped testing on node 12? and if so, do we need to update the package.json's required node engine to reflect the change?

package.json
nmanu1: I just meant that if you're addressing Snyk issues in the repo, maybe there were also Snyk issues in the `static/package-lock` that would be nice to fix too

package.json
tatimblin: That is a good idea. If it's agreed we should only support versions we test on then I can set the node version target to >14. If we're not sure I'll downgrade `serve`, to a version that has the same node support as originally.

package-lock.json
```diff
@@ -1,16551 +1,8 @@
 {
   "name": "answers-hitchhiker-theme",
   "version": "1.30.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
nmanu1, line 4: can we switch back to lockfileVersion 2?

package.json
nmanu1: I'm seeing @percy/puppeteer supports [>=14](https://github.com/percy/percy-puppeteer/blob/master/package.json#L20) for v2.0.2

package-lock.json
tatimblin: that comes from npm 7 and is backwards compatible with version 1. https://github.blog/2021-02-02-npm-7-is-now-generally-available/#changes-to-the-lockfile

package.json
tatimblin: hmm reverting this didn't work because the change was just a patch. I forced it

nmanu1: I still think it would be worth checking if Node 12 support is no longer necessary. It would be nice to get these fixes if we can

package.json
tatimblin: Ah I see, no the list I went through just had those packages as immediate dependencies with patch fixes

package-lock.json
tatimblin: Ah I had this backward, switched back

package.json
nmanu1: oh they're listed separately in Snyk. [here](https://app.snyk.io/org/answers/project/5fa047db-fb78-43cd-9ba1-6c56fbd123bc)'s the project for `static/package.json`

package-lock.json
```diff
@@ -3694,21 +3686,18 @@
...
+      "version": "1.16.0",
+      "resolved": "https://registry.npmjs.org/@percy/sdk-utils/-/sdk-utils-1.16.0.tgz",
+      "integrity": "sha512-/nPyK4NCjFGYNVQ7vOivfuEYveOJhA4gWzB7w2PjCkw/Y3kCtu+axRpUiDPEybTz2H6RTvr+I526DbtUYguqVw==",
       "dev": true,
-      "dependencies": {
-        "@percy/logger": "1.0.0-beta.63"
-      },
       "engines": {
-        "node": ">=12"
+        "node": ">=14"
```
nmanu1, line 3694: this also needs to be downgraded

static/package.json
```diff
@@ -36,7 +36,7 @@
     "locale-currency": "0.0.2",
     "mini-css-extract-plugin": "^1.6.0",
     "postcss": "^8.4.14",
-    "resolve-url-loader": "^3.1.4",
+    "resolve-url-loader": "^3.1.5",
```
nmanu1, line 39: can we run `npm install` in the `static` directory to get the package-lock updates? also, might be worth running `npm audit fix` there too

nmanu1 (commented):
looks like tests are failing

static/package.json
tatimblin: sure thing. I did try npm run audit, but there were no free fixes from that for the test-site

package.json
tatimblin: updated `resolve-url-loader` and tried `npm run audit`

package-lock.json
```diff
@@ -3686,12 +3686,12 @@
       }
     },
     "node_modules/@percy/sdk-utils": {
-      "version": "1.16.0",
-      "resolved": "https://registry.npmjs.org/@percy/sdk-utils/-/sdk-utils-1.16.0.tgz",
-      "integrity": "sha512-/nPyK4NCjFGYNVQ7vOivfuEYveOJhA4gWzB7w2PjCkw/Y3kCtu+axRpUiDPEybTz2H6RTvr+I526DbtUYguqVw==",
+      "version": "^1.0.0-beta.63",
```
nmanu1, line 3689: looks like this was a manual edit of the package-lock. the caret before the version number is causing tests to fail

PR #1121, opened 2022-12-15
v1.30.0

tmeyer2115:
### Changes
- If `initializeManually` is `true`, initialization will happen automatically in Live Preview. This is to improve local development experience. (#1088)
- Support for IE11 was officially dropped. We no longer guarantee that subsequent Theme versions will be IE11 compatible. (#1112)
- Assorted updates were made to our GH Actions. (#1113, #1111)


nmanu1 (commented):
do we want to wait for [this PR](https://github.com/yext/answers-hitchhiker-theme/pull/1118) to be merged in?

Edit: since we don't need to wait for it, lgtm!

nmanu1 (approved):

PR #1125, opened 2023-01-09
Correct bug in `isStaging` method.

tmeyer2115:
This PR corrects a bug in the `isStaging` method. If you were referred to a Production site from a Staging link, `isStaging` would incorrectly report `true`. This is because the `includes` check on the `href` would catch the Staging `referrerPageUrl`. Simply comparring `window.location.hostname` instead of `window.location.href` fixes this. I also did some code cleanup in this method and its test suite.

J=SLAP-2546
TEST=auto

### DIFF
--- a/static/js/is-staging.js
+++ b/static/js/is-staging.js
@@ -1,9 +1,19 @@
+import escapeStringRegexp from 'escape-string-regexp';
+
 /**
  * Determines whether the current url is staging or production.
  */
 export function isStaging(stagingDomains) {
   const defaultStagingDomains = ['127.0.0.1', 'localhost', 'office.yext.com'];
-  const _stagingDomains = defaultStagingDomains.concat(stagingDomains);
-  const currentUrl = window.location.href;
-  return _stagingDomains.some(domain => currentUrl.includes(domain));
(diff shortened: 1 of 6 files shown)


static/js/is-staging.js
```diff
@@ -1,9 +1,19 @@
...
-  return _stagingDomains.some(domain => currentUrl.includes(domain));
-}
+  const _stagingDomains = stagingDomains
+    ? defaultStagingDomains.concat(stagingDomains)
+    : defaultStagingDomains;
+  const currentDomain = window.location.hostname;
+
+  return _stagingDomains.some(domain => {
+    const regexEscapedDomain = escapeStringRegexp(domain);
+    const isSubdomainRegex = new RegExp(`[A-Za-z0-9]?[A-Za-z0-9\-]{0,61}[A-Za-z0-9]?\.${regexEscapedDomain}`);
```
oshi97, line 15: Why 0 to 61 instances?

tmeyer2115: The RFC for subdomains says they can contain a maximum of 63 characters. The leading and trailing characters must be alpha numeric.

oshi97 (approved):

PR #1128, opened 2023-04-17
Update Dependencies for Testing and Remove Unnecessary Actions.  

tmeyer2115:
This PR removes the GH Actions related to Browserstack and Percy, since we no longer have those services. It also updates a number of dependencies for testing to get rid of some Snyk vulnerabilities. 

TEST=auto

Made sure all unit tests and remaining acceptance tests passed. WCAG checks were previously failing.


package.json
```diff
@@ -38,7 +38,7 @@
     "@percy/puppeteer": "^2.0.0",
```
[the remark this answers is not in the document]
tmeyer2115: Removed the other Percy deps! I want to keep the code for generating the snapshots, however. That can serve as a guide for Backfire as they cut over from Percy to Playwright.

package.json
```diff
@@ -96,6 +94,7 @@
     ]
   },
   "dependencies": {
+    "decode-uri-component": "^0.2.2",
```
nmanu1, line 97: Why are we adding this as a new dependency? Can we just bump the version with `npm update` if it's a transitive dependency?

tmeyer2115: Unfortunately, that didn't work 😬 The only way I was able to get the updated version was making it a top-level dep.

tmeyer2115: Done!

nmanu1: I see. Now that it is updated in the package-lock, can we remove it from here?

PR #1134, opened 2023-05-04
Support EU Cloud Region

cea2aj:
Add a cloud region config option to the global_config to allow either 'us' (default), or 'eu' to be set as the cloud region.

The cloud region determines where the search assets and ytag are sourced from.

J=BACK-2277
TEST=manual, auto

Using the test site, set the cloudRegion to "eu" and confirmed the assets are loaded from the EU CDN and executed searches against the EU search endpoints

### DIFF
--- a/hbshelpers/sdkAssetUrl.js
+++ b/hbshelpers/sdkAssetUrl.js
@@ -2,6 +2,8 @@ const RELEASE_BRANCH_REGEX = /^release\/v[0-9.]+$/;
 const HOTFIX_BRANCH_REGEX = /^hotfix\/v[0-9.]+$/;
 const I18N_FEATURE_BRANCH_REGEX = /^feature\/.+-i18n$/;
 const SEM_VER_REGEX = /^[1-9]+$|^[1-9]+\.[0-9]+$|^[1-9]+\.[0-9]+\.[0-9]+$/;
+const US = 'us';
+const EU = 'eu';
 
 /**
  * Given a branch (or release) of the SDK and a locale, this helper provides the correct
@@ -17,7 +19,7 @@ module.exports = function sdkAssetUrl(
   branch,
   locale,
   assetName,
-  cloudRegion = 'us'
+  cloudRegion = US
 ) {
   const isReleasedBranch = SEM_VER_REGEX.test(branch);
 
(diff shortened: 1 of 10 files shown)


hbshelpers/sdkAssetUrl.js
```diff
@@ -24,17 +30,21 @@ module.exports = function sdkAssetUrl(branch, locale, assetName) {
...
+    (isReleasedBranch || isPreReleaseBranch || isI18nFeatureBranch || isDevelopBranch) &&
     !(locale.startsWith('en') || assetName === 'answers.css') ;
-  
+
   const parsedAssetName = isLocalizationSupported ?
     `${locale}-${assetName}` :
     assetName;
 
-  return `https://assets.sitescdn.net/answers/${parsedBranch}/${parsedAssetName}`;
+  const domain = cloudRegion === 'eu'
```
EmilyZhang777, line 45: can you create a const for 'eu' to improve readability? Maybe one for 'us' as well.

test-site/config/global_config.json
```diff
@@ -1,9 +1,10 @@
 {
-  "sdkVersion": "develop", // The version of the Answers SDK to use
+  "sdkVersion": "1.15", // The version of the Answers SDK to use
```
EmilyZhang777, line 2: is this change intended?

EmilyZhang777 (commented):
Only some nit.

layouts/html.hbs
```diff
@@ -24,8 +24,12 @@
     <link rel="dns-prefetch" href="//dynl.mktgcdn.com">
     <link rel="dns-prefetch" href="//dynm.mktgcdn.com">
```
benmcginnis, line 25: do we know if these domains will be different @EmilyZhang777 ?

benmcginnis (approved):
LGTM, just had one question but it can be handled later if there is a change we have to make.

hbshelpers/sdkAssetUrl.js
cea2aj: sure

test-site/config/global_config.json
cea2aj: nope, reverted

benmcginnis (approved):

layouts/html.hbs
EmilyZhang777: I'm not sure. Based on [this](https://yext.slack.com/archives/CNVUY09HS/p1679068294340899) conversation, seems like eu is going to another bucket?

EmilyZhang777 (approved):
Only a small comment that we can address later. Otherwise, lgtm!

PR #1138, opened 2023-06-15
Support direct rendering of HTML for richTextV2 and Markdown

saahithjanapati:
Support direct rendering of HTML (if richTextV2 or Markdown is used and conversion to HTML is enabled)

J=BACK-2241

T=manual

created a test site using the answers hitchikers theme and verified that rich text was rendered properly


cea2aj (approved):
LGTM!

PR #1142, opened 2023-08-14
Update FAQ card to support answerV2 instead of answer

likimmy:
J=BACK-2436
TEST=manual

Created test site using jambo template and verified that FAQ answer was rendered properly

### DIFF
--- a/cards/faq-accordion/component.js
+++ b/cards/faq-accordion/component.js
@@ -17,7 +17,9 @@ class faq_accordionCardComponent extends BaseCard['faq-accordion'] {
     if (profile.answerV2 && profile.answerV2.html) {
       profileAnswer = profile.answerV2.html;
     } else if (profile.answerV2 && profile.answerV2.json) {
-      console.warn('JSON is not supported, please switch to HTML.')
+      console.warn('JSON is not supported, please convert to HTML.')
+    } else if (profile.answer) {
+      profileAnswer = ANSWERS.formatRichText(profile.answer, "answer", linkTarget);
     }
 
     return {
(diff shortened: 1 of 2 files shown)


cards/faq-accordion/component.js
```diff
@@ -13,11 +13,18 @@ class faq_accordionCardComponent extends BaseCard['faq-accordion'] {
    */
   dataForRender(profile) {
     const linkTarget = AnswersExperience.runtimeConfig.get('linkTarget') || '_top';
+    let profileAnswer = '';
+    if (profile.answerV2 && profile.answerV2.html) {
+      profileAnswer = profile.answerV2.html;
+    } else if (profile.answerV2 && profile.answerV2.json) {
+      console.warn('JSON is not supported, please switch to HTML.')
```
cea2aj, line 20: nit: I think we should say use the word 'convert' rather than 'switch' because the feature on the backend uses the word 'convert' as well

cards/faq-accordion/component.js
```diff
@@ -13,11 +13,18 @@ class faq_accordionCardComponent extends BaseCard['faq-accordion'] {
...
+    }
 
     return {
       title: profile.question || profile.name, // The header text of the card
       // subtitle: '', // The sub-header text of the card
-      details: profile.answer ? ANSWERS.formatRichText(profile.answer, "answer", linkTarget) : null, // The text in the body of the card
```
cea2aj, line 20: I think it would be a good idea to add a 3rd if statement which maintains this functionality of calling ANSWERS.formatRichText if profile.answers exists. That way if someone upgrades to this version of the hitchhiker-theme, it won't break any existing sites which use the faq-accordion card

cards/faq-accordion/component.js
likimmy: updated in most recent commit

cards/faq-accordion/component.js
likimmy: that makes sense! i've added that check into the logic in the most recent commit

cea2aj (approved):
LGTM!

PR #1147, opened 2023-08-14
Fix failing translation tests

cea2aj:
The translation files were failing because the line numbers in the multilang-accordion card changed

I ran `npm run extract-translations` to update the files

J=BACK-2463
TEST=manual

Confirmed tests now pass


likimmy (approved):
lgtm!

PR #1148, opened 2023-09-25
Fix iframe Answers Map/List toggle button on Mobile Error

jesuyedavid:
Iframe needs height in order for map and toggle button to be initialized and to fill js-answersMap


benmcginnis:
@jesuyedavid what's the deal with this pull request?

vijay267:
@jesuyedavid was this change actually no longer needed?

PR #1150, opened 2023-09-25
Fix iframe Answers Map/List toggle button on Mobile Error

jesuyedavid:
Iframe needs height in order for map and toggle button to be initialized and to fill js-answersMap

### DIFF
--- a/static/js/iframe-common.js
+++ b/static/js/iframe-common.js
@@ -70,8 +70,6 @@ export function generateIFrame(domain, answersExperienceFrame) {
   iframe.frameBorder = 0;
 
    // For dynamic iFrame sizing
-  iframe.style.height = '100%';
-  iframe.style.minHeight = '100%';
   iframe.style.width = '1px';
   iframe.style.minWidth = '100%';
   iframe.id = 'answers-frame';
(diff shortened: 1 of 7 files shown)


static/js/iframe-common.js
```diff
@@ -70,6 +70,8 @@ export function generateIFrame(domain, answersExperienceFrame) {
   iframe.frameBorder = 0;
 
    // For dynamic iFrame sizing
+  iframe.style.height = '100%';
+  iframe.style.minHeight = '100%';
```
benmcginnis, line 74: I'm a little uncertain about this solution. There are a lot of potential problems with setting 100% height. 

Can we take a look at this with a test across a few different browsers?

static/js/theme-map/VerticalFullPageMapOrchestrator.js
```diff
@@ -205,42 +205,42 @@ class VerticalFullPageMapOrchestrator extends ANSWERS.Component {
...
-        return els[0];
-      }
-    }
-
-    function setFixedHeight(selector) {
-      const el = getSingleElement(selector)
-      el.style.height = `${el.scrollHeight}px`
-    }
-  }
+  //NOTE: set min-height for js-answersMap instead to prevent keyboard from shifting the page
```
cea2aj, line 210: If we don't need this code we can delete it instead of commenting it out

cea2aj (dismissed):

cea2aj (approved):
LGTM! We can ignore the percy and browserstack errors since we don't use those services anymore

PR #1152, opened 2023-09-29
fix(iframe-common.js): properly encode referrerPageUrl param

JRonkin:
Currently, referrerPageUrl is added as a query param value without being escaped. That can cause problems on sites with tighter security restrictions. We faced this problem on several sites such as this one: https://yexttest.atlassian.net/browse/PC-208916

It can also cause functional problems when the referrer URL has query parameters. For example, a URL of http://example.com?a=1&b=2 will lead to https://answers.example.com?query=TEST&referrerPageUrl=http://example.com?a=1&b=2

In that example, &b=2 will be treated as a query parameter of the  results page URL instead of being part of the referrerPageUrl value.

### DIFF
--- a/static/js/iframe-common.js
+++ b/static/js/iframe-common.js
@@ -29,7 +29,7 @@ export function generateIFrame(domain, answersExperienceFrame) {
     // Parse the params out of the URL
     var params = paramString.split('&'),
                  verticalUrl;
-    var referrerPageUrl = document.referrer.split('?')[0].split('#')[0];
+    var referrerPageUrl = encodeURIComponent(document.referrer.split('?')[0].split('#')[0]);
 
     if (pathToIndex) {
       verticalUrl = pathToIndex;


cea2aj:
Can you change the base of this PR to the hotfix/v1.33.2 branch? Otherwise LGTM!

cea2aj:
> > Can you change the base of this PR to the hotfix/v1.33.2 branch? Otherwise LGTM!
> 
> Done. There are still a couple automated tests failing. Is that expected?

We can ignore all of the browserstack and percy tests because we don't use those services anymore. The headless acceptance tests are still used, however those test seem to behave very unpredictably based on the last few commits in master.  I re-ran the headless acceptance tests to see if any more pass

JRonkin:
> We can ignore all of the browserstack and percy tests because we don't use those services anymore. The headless acceptance tests are still used, however those test seem to behave very unpredictably based on the last few commits in master. I re-ran the headless acceptance tests to see if any more pass

Still failed. Should we ignore that too?

cea2aj:
> Still failed. Should we ignore that too?

Only one test failed this time instead of eight, which is the same as the master branch which suggests that the issue isn't due to this change. So we can ignore that failing test

cea2aj (approved):

PR #1154, opened 2023-10-23
feat: iframe accessibility

benmcginnis:
This change udpates the iframe-common script to set several important attributes for screen reader's to properly interpret the presence of the iframe.

The attributes do the following things:
* ariaBusy (aria-busy): notifies the screen reader that the iframe is loading
* role=document: tells screen readers to treat the frame contents as a document
* aria-live=assertive: tells screen readers to immediately read the frame content
* tabindex=0: tabindex is required on document roles to ensure that it is treated as keyboard accessible

J=TECHOPS-10193
TEST=manual

Manually tested using both Mac VoiceOver and Windows Narrator and validated that the results were read by the screen reader.


EmilyZhang777 (approved):

nmanu1 (commented):
You'll also want to point the PR to `hotfix/v1.33.2` instead of `master` if it should be part of the new patch. Otherwise, it should point to `develop`

PR #1155, opened 2023-11-17
Implement playwright acceptance tests for vertical full page map

likimmy:
J=BACK-2714


.gitignore
```diff
@@ -2,4 +2,12 @@ static/node_modules/
...
\ No newline at end of file
+.idea/
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
```
cea2aj, line 13: it looks like these are duplicates

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+test.describe('full page map test suite', () => {
+  test.beforeEach(async ({page}) => {
+    await page.goto('http://localhost:5042/locations_full_page_map');
+    await page.getByPlaceholder('Search for locations').click();
+  });
+
+  test('can search and get results', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').fill('virginia');
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    await page.locator('#js-answersVerticalResults div').filter({ hasText: '1 Office Space 8.2 mi Close Card 7900 Westpark Drive Suite T200 McLean, VA 22102' }).nth(3).click();
```
cea2aj, line 12: I think it's best to have an `expect` statement at the end of the test. Can we check that an element with `.yxt-Results` exists like the old tests? I think that would be more robust than trying to match that entire string

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+  test('clicking on a pin focuses on a result card', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').fill('virginia');
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    await page.getByRole('button', { name: 'Result number 5' }).click();
+    const locator = page.locator('#js-answersVerticalResults div').nth(134);
```
cea2aj, line 19: what is the 134 value here? is there another way to target it that is makes easier to understand what is being selected?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+  // search when map moves works
+  test('search when map moves works', async ({ page }) => {
+    const searchLogo = '#js-yext-submit';
+    await page.mouse.move(600, 300);
+    await page.mouse.down();
+    await page.mouse.move(1200, 450, {steps: 5});
+    await page.mouse.up();
+    await page.waitForSelector(searchLogo, { state: 'detached' });
+    const detachedSearchLogo = await page.$(searchLogo);
+    expect(detachedSearchLogo).toBeFalsy();
```
cea2aj, line 32: is this attempting to determine that a search was ran from the search logo detaching?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+  test('search this area button works', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').fill('virginia');
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    // search bar yext logo remounts when search occurs
```
cea2aj, line 38: Have you seen the playwright [waitForResponse](https://playwright.dev/docs/api/class-page#page-wait-for-response) method? I think that checking for that request in the network tab is  more robust than checking if the logo remounts

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+    await expect(secondPage).toHaveText(/21/);
+  });
+
+  test('pagination scrolls the results to the top', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').fill('virginia');
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    await page.getByLabel('Go to the next page of results').click();
+    const isScrolledUp = await page.evaluate(() => {
+      return window.scrollY === 0;
+    });
```
cea2aj, line 67: we should check the scrollTop of the `.Answers-resultsWrapper`. I don't believe the scrollY of the window itself changes on pagination

tests-examples/demo-todo-app.spec.ts
```diff
@@ -0,0 +1,437 @@
+import { test, expect, type Page } from '@playwright/test';
+
+test.beforeEach(async ({ page }) => {
```
cea2aj, line 3: were these tests intended to be here?

e2e/vertical-full-page-map.spec.ts
likimmy: I was struggling getting the correct component because Answers.resultsWrapper didn't seem to work/have a scrollTop member. Instead I just checked if the first element of the results was visible, but let me know if this is not robust enough for what we are testing here!

tests-examples/demo-todo-app.spec.ts
likimmy: oops, these were auto-generated. deleted!

cea2aj: It looks like these were auto-generated again. I wonder if you need to disable something in order to prevent this from being created again

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  test('search when map moves works', async ({ page }) => {
+    const searchLogo = '#js-yext-submit';
+    await page.mouse.move(600, 300);
+    await page.mouse.down();
+    await page.mouse.move(1200, 450, {steps: 5});
+    await page.mouse.up();
+    await page.waitForSelector(searchLogo, { state: 'detached' });
+    const detachedSearchLogo = await page.$(searchLogo);
+    expect(detachedSearchLogo).toBeFalsy();
```
cea2aj, line 32: is it possible to check for a search request here as well, rather than checking the search logo?

e2e/vertical-full-page-map.spec.ts
cea2aj: I wonder if it's possible to write a query that checks for the `pinFocused` class on any of the children of `#js-answersVerticalResults div`.  That way we don't need to include the 134 number because that seems like it could break easily if the search results change

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    const count = await page.locator('#js-answersVerticalResults').count();
+    expect(count).toBeGreaterThan(0);
+  });
+
+  test('clicking on a pin focuses on a result card', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').fill('virginia');
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    await page.getByRole('button', { name: 'Result number 5' }).click();
+    const locator = page.locator('#js-answersVerticalResults div').nth(134);
```
cea2aj, line 20: Would `resultCard` be a more accurate name than `locator`? Or is it a pin that we are checking if it has the `pinFocused` class?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    const result = page.locator('#js-answersVerticalResults div').nth(0);
+    await expect(result).toBeAttached();
+  });
+
+  test('pagination works', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').fill('virginia');
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    await page.getByLabel('Go to the next page of results').click();
+    const secondPage = page.locator('#js-answersVerticalResultsCount');
+    await expect(secondPage).toHaveText(/21/);
```
cea2aj, line 54: I'm cautious about this check because this would break if an entity is added to the account because it would change the results count. Another idea is checking the query params of the web page to see if the offset changed. But if it's too much effort I think we can stick with this because the slapshot test account is pretty stable

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    await page.mouse.up();
+    await page.waitForSelector(searchLogo, { state: 'detached' });
+    const detachedSearchLogo = await page.$(searchLogo);
+    expect(detachedSearchLogo).toBeFalsy();
+  });
+
+  test('search this area button works', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').fill('virginia');
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    const responsePromise = await page.waitForResponse(/https:\/\/prod-cdn\.us\.yextapis\.com\/v2\/accounts\/me\/search\/vertical/i);
```
nmanu1, line 38: nit: looks like this is storing the actual response, not a promise

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    await expect(filterView).toBeVisible();
+    await page.getByRole('button', { name: 'Result number 1' }).click();
+    await expect(filterView).not.toBeVisible();
+  });
+
+  test('clicking on a cluster causes the map to zoom in', async ({ page }) => {
+    const mapboxPinCount = await page.locator('#js-answersMap div').count();
+    await page.getByRole('button', { name: 'Cluster of 2 results' }).click();
+    const mapboxPinCountAfterSelectingCluster = await page.locator('#js-answersMap div').count();
+    expect(mapboxPinCount).not.toBe(mapboxPinCountAfterSelectingCluster);
```
nmanu1, line 87: can we check that there are fewer pins, rather than an arbitrary different number?

playwright.config.ts
```diff
@@ -0,0 +1,48 @@
...
+    {
+      name: 'chromium',
+      use: { ...devices['Desktop Chrome'] },
+    },
+    {
+      name: 'webkit',
+      use: { ...devices['Desktop Safari'] },
+    },
+    // {
+    //   name: 'firefox',
```
nmanu1, line 37: rather than disabling firefox entirely, what do you think of only disabling it in CI if it works locally?

e2e/vertical-full-page-map.spec.ts
likimmy: I used 21 because the offset is set to 20 for each page, so i think as long as the results size is more than 20, even if the total results count changes the check for 21 should work, but let me know if i am misunderstanding your comment!

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    const detachedSearchLogo = await page.$(searchLogo);
+    expect(detachedSearchLogo).toBeFalsy();
+  });
+
+  test('search this area button works', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').fill('virginia');
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    const responsePromise = await page.waitForResponse(/https:\/\/prod-cdn\.us\.yextapis\.com\/v2\/accounts\/me\/search\/vertical/i);
+    await page.getByLabel('Map controls').getByText('Search When Map Moves').click();
+    const response = responsePromise;
```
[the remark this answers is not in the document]
likimmy: i've changed this so that it's used to check if the response code is 200!

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    const response = await page.waitForResponse(resp => resp.url().includes('https://prod-cdn.us.yextapis.com/v2/accounts/me/search/vertical/query')
+      && resp.url().includes('input=virginia') 
+      && !resp.url().includes('filters'));
+    await expect(response.status()).toBe(200);
+  });
+
+  test('clicking on a pin closes the filter view', async ({ page }) => {
+    await page.getByRole('button', { name: 'filter results' }).click();
+    await page.getByText('Cats (1)').click();
+    const filterView = page.getByLabel('Main location search').locator('div').filter({ hasText: 'Filters Services reset Cats (1) Dogs (1) Sleep (1)' }).first();
```
EmilyZhang777, line 95: Would this text always be exactly like this? I'm worried the order/number might change. Would you be able to select the element using something like classname/label or do a regex match of the text?

EmilyZhang777 (approved):
Looks good! Just have a few comments.

e2e/vertical-full-page-map.spec.ts
```diff
@@ -17,27 +17,41 @@ test.describe('full page map test suite', () => {
     await page.getByPlaceholder('Search for locations').fill('virginia');
     await page.getByPlaceholder('Search for locations').press('Enter');
     await page.getByRole('button', { name: 'Result number 5' }).click();
-    const locator = page.locator('#js-answersVerticalResults div').nth(134);
-    await expect(locator).toHaveClass(/pinFocused/);
+    const results = page.locator('#js-answersVerticalResults div');
+    const count = await results.count();
+    let hasPinFocused = false;
+
+    for (var i = 0; i < count; i++) {
```
nmanu1, line 24: can we use a stream function (probably `some`) here instead of a for loop? or could we even directly check for a match for a locator on the `yxt-Card--pinFocused` class?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -17,27 +17,41 @@ test.describe('full page map test suite', () => {
...
-    const searchLogo = '#js-yext-submit';
-    await page.mouse.move(600, 300);
-    await page.mouse.down();
-    await page.mouse.move(1200, 450, {steps: 5});
-    await page.mouse.up();
-    await page.waitForSelector(searchLogo, { state: 'detached' });
-    const detachedSearchLogo = await page.$(searchLogo);
-    expect(detachedSearchLogo).toBeFalsy();
+    await page.getByPlaceholder('Search for locations').fill('virginia');
+    await page.getByPlaceholder('Search for locations').press('Enter');
```
nmanu1, line 39: what's the reason for conducting a search at the start of this test? couldn't the `waitForResponse` below be matching this search instead of the one triggered from moving the map?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -17,27 +17,41 @@ test.describe('full page map test suite', () => {
...
+    await page.mouse.wheel(600, 300);
+    const response = await page.waitForResponse(resp =>
+      resp.url().includes('https:\/\/prod-cdn\.us\.yextapis\.com\/v2\/accounts\/me\/search\/vertical\/query'));
+    await expect(response.status()).toBe(200);
   });
 
   test('search this area button works', async ({ page }) => {
     await page.getByPlaceholder('Search for locations').fill('virginia');
     await page.getByPlaceholder('Search for locations').press('Enter');
```
nmanu1, line 48: same as above. what's the reason for making this search at the beginning?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -17,27 +17,41 @@ test.describe('full page map test suite', () => {
...
-    const responsePromise = await page.waitForResponse(/https:\/\/prod-cdn\.us\.yextapis\.com\/v2\/accounts\/me\/search\/vertical/i);
     await page.getByLabel('Map controls').getByText('Search When Map Moves').click();
-    const response = responsePromise;
+    await page.locator('div').filter({ hasText: /^Search This Area$/ }).nth(1).click();
+    await page.mouse.move(1200, 450, {steps: 5});
```
nmanu1, line 51: why do we move the map here? the search should be conducted just by clicking the `Search this area` button. is the idea that we want to check that moving the map no longer triggers a search since `Search when map moves` was toggled to false?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -81,16 +99,20 @@ test.describe('full page map with filters test suite', () => {
...
-    expect(mapboxPinCount).not.toBe(mapboxPinCountAfterSelectingCluster);
+     const response = await page.waitForResponse(resp => resp.url().includes('https://prod-cdn.us.yextapis.com/v2/accounts/me/search/vertical/query')
+      && resp.url().includes('input=virginia') 
+      && resp.url().includes('filters'));
+     await expect(response.status()).toBe(200);
+    const countAfterSelectingCluster = await page.locator('.yxt-Card').count();
+    expect(originalCount).toBeGreaterThan(countAfterSelectingCluster);
   });
 
   test('clicking on a cluster causes a new search to be run', async ({ page }) => {
```
nmanu1, line 112: it looks like we already test this in the test above? if we still want it as a separate test, can we update to check that the response below has a filter so we know it's not matching the original `virginia` query?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,118 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('full page map test suite', () => {
+  test.beforeEach(async ({page}) => {
+    await page.goto('http://localhost:5042/locations_full_page_map');
+    await page.getByPlaceholder('Search for locations').click();
```
nmanu1, line 6: does this just click into the search bar? if so, why is that needed?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  });
+
+  test('can search and get results', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').fill('virginia');
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    const count = await page.locator('#js-answersVerticalResults').count();
+    expect(count).toBeGreaterThan(0);
```
nmanu1, line 13: even if the `virginia` search didn't happen, the count would be greater than 0 because of the search conducted when first visiting the page. instead, can we check that the count is different from before the search vs. after?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  });
+
+  test('clicking on a pin focuses on a result card', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').fill('virginia');
```
nmanu1, line 17: can we skip doing this search? since all results are returned for the blank initial search, I think we could proceed with test without it

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  });
+
+  test('search this area button works', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').fill('virginia');
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    await page.getByLabel('Map controls').getByText('Search When Map Moves').click();
+    await page.locator('div').filter({ hasText: /^Search This Area$/ }).nth(1).click();
+    await page.mouse.move(1200, 450, {steps: 5});
+    const response = await page.waitForResponse(resp =>
+      resp.url().includes('https:\/\/prod-cdn\.us\.yextapis\.com\/v2\/accounts\/me\/search\/vertical\/query'));
```
nmanu1, line 53: can we update this check so it wouldn't match the default initial search?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -3,54 +3,41 @@ import { test, expect } from '@playwright/test';
...
-      const classes = await currResult.evaluate(node => Array.from(node.classList));
-
-      if (classes.includes('yxt-Card--pinFocused')) {
-        hasPinFocused = true;
-        break;
-      }
-    }
-
-    expect(hasPinFocused).toBe(true);
+    await expect(page.locator('.yxt-Card--pinFocused')).toBeAttached();
```
nmanu1, line 12: can we also check that this locator doesn't match anything before clicking the pin?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -91,28 +78,24 @@ test.describe('full page map with filters test suite', () => {
...
+    const filters = page.locator('#js-answersFacets');
+    await expect(filters).toBeVisible();
+    await page.getByRole('button', { name: 'Result number 7' }).click();
+    await expect(filters).not.toBeVisible();
   });
 
-  test('clicking on a cluster causes the map to zoom in', async ({ page }) => {
+  test('clicking on a cluster causes the map to zoom in and new search is ran', async ({ page }) => {
+    const result = page.locator('#js-answersVerticalResults div').nth(0);
+    await expect(result).toBeAttached();
```
nmanu1, line 89: is this check needed for timing purposes? if so, it might need to be wrapped in a `waitFor`. same with below (line 97)

e2e/vertical-full-page-map.spec.ts
```diff
@@ -3,54 +3,41 @@ import { test, expect } from '@playwright/test';
...
     await expect(response.status()).toBe(200);
   });
 
   test('search this area button works', async ({ page }) => {
+    await page.getByLabel('Map controls').getByText('Search When Map Moves').click();
     await page.getByPlaceholder('Search for locations').fill('virginia');
     await page.getByPlaceholder('Search for locations').press('Enter');
-    await page.getByLabel('Map controls').getByText('Search When Map Moves').click();
+    await page.waitForResponse(resp =>
+      resp.url().includes('https:\/\/prod-cdn\.us\.yextapis\.com\/v2\/accounts\/me\/search\/vertical\/query'));
```
nmanu1, line 36: can we check that this response doesn't have `radius` in the url?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -3,54 +3,41 @@ import { test, expect } from '@playwright/test';
 test.describe('full page map test suite', () => {
   test.beforeEach(async ({page}) => {
     await page.goto('http://localhost:5042/locations_full_page_map');
-    await page.getByPlaceholder('Search for locations').click();
-  });
-
-  test('can search and get results', async ({ page }) => {
```
nmanu1: how come this test is removed?

likimmy: oops i think this was an accident

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+  test('pagination works', async ({ page }) => {
+    await page.getByLabel('Go to the next page of results').click();
+    const secondPage = page.locator('#js-answersVerticalResultsCount');
+    await expect(secondPage).toHaveText(/21/);
+  });
+
+  test('pagination scrolls the results to the top', async ({ page }) => {
+    await page.getByLabel('Go to the next page of results').click();
+    const topOfResults = page.locator('#js-answersVerticalResults div').nth(0);
+    await expect(topOfResults).toBeVisible();
```
nmanu1, line 66: can we check that the topOfResults is not visible before the pagination click?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  test('clicking on a cluster causes the map to zoom in and new search is ran', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').fill('virginia');
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    await page.waitForResponse(resp => resp.url().includes('https://prod-cdn.us.yextapis.com/v2/accounts/me/search/vertical/query')
+      && resp.url().includes('input=virginia') 
+      && !resp.url().includes('filters'));
+
+    const result = page.locator('#js-answersVerticalResults div').nth(0);
+    await expect(result).toBeAttached();
```
nmanu1, line 92: since we do this check in a few places, can we pull this out into a separate function/check that we reuse in the different tests?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -1,15 +1,31 @@
 import { test, expect } from '@playwright/test';
 
+async function getAnswersVerticalResults(page) {
```
nmanu1, line 3: nit: maybe this could be renamed to better explain it's purpose, e.g. `waitForResultsLoad` or `waitForResultsToRender`

e2e/vertical-full-page-map.spec.ts
```diff
@@ -1,15 +1,31 @@
...
   test.beforeEach(async ({page}) => {
     await page.goto('http://localhost:5042/locations_full_page_map');
   });
 
   test('can search and get results', async ({ page }) => {
+    const prevResponse = await page.waitForResponse(resp =>
+      resp.url().includes('https:\/\/prod-cdn\.us\.yextapis\.com\/v2\/accounts\/me\/search\/vertical\/query')
+      && resp.url().includes('queryTrigger'));
+    const prevResponseJson = await prevResponse.json();
+    const prevResultsCount = prevResponseJson.response.resultsCount;
```
nmanu1, line 18: can we check the count from the ResultsCount in the UI, rather than the response, to make sure it's rendering properly?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -49,20 +62,32 @@ test.describe('full page map test suite', () => {
...
+  test('empty search works', async ({ page }) => {
+    await page.getByPlaceholder('Search for locations').press('Enter');
+    const response = await page.waitForResponse(resp =>
+      resp.url().includes('https:\/\/prod-cdn\.us\.yextapis\.com\/v2\/accounts\/me\/search\/vertical\/query')
+      && resp.url().includes('input=&'));
+    expect(response.status()).toBe(200);
+  })
+
   test('pagination works', async ({ page }) => {
+    const firstPage = page.locator('#js-answersVerticalResultsCount');
```
nmanu1, line 80: looks like the `firstPage` and `secondPage` locators are the same. can we just use one of them and call it `resultsCount` or something?

e2e/vertical-full-page-map.spec.ts
```diff
@@ -49,20 +62,32 @@ test.describe('full page map test suite', () => {
...
+    await expect(firstPage).toHaveText(/20/);
     await page.getByLabel('Go to the next page of results').click();
     const secondPage = page.locator('#js-answersVerticalResultsCount');
     await expect(secondPage).toHaveText(/21/);
   });
 
   test('pagination scrolls the results to the top', async ({ page }) => {
-    await page.getByLabel('Go to the next page of results').click();
     const topOfResults = page.locator('#js-answersVerticalResults div').nth(0);
+    await expect(topOfResults).not.toBeVisible();
```
nmanu1, line 89: out of curiosity, how come the top of the results isn't visible when we haven't scrolled down the page before this?

e2e/vertical-full-page-map.spec.ts
likimmy: it will only render 20 results since the ui supports pagination so i don't think there is a ui component i can locate to check the total results count

nmanu1: I thought the results count component should give you a total, like "1-20 of 72", and you could locate the "72"?

e2e/vertical-full-page-map.spec.ts
likimmy: on second thought in response to your previous comment about checking if the top of page is not visible before we go to the next page of results, i don't think that would be reasonable in this test until we added in actions for manually scrolling to the bottom of the page, since with playwright we can just use a locator to find the pagination button. i think it just makes sense for this test to check if the first result card is visible after clicking the button and not if the first result isn't visible before clicking it.

nmanu1: based on the name of this test, I thought the purpose was to check that the scrolling offset resets itself after pagination? if you're already at the top of the page and then click the pagination button, it won't actually test that the scroll offset has reset to 0, because it was already 0 to begin with. we likely would need an action for manually scrolling the page first

PR #1163, opened 2024-01-16
Handle EU-Served Images

EmilyZhang777:
This PR makes sure EU-served images are handled properly since they have a different format than US-served images.

J=TECHOPS-10895
TEST=auto,manual

Added and ran jest tests.
Manually checked out pages with EU and US-served images and made sure they all show up correctly.

### DIFF
--- a/static/js/formatters-internal.js
+++ b/static/js/formatters-internal.js
@@ -284,15 +284,15 @@ export function joinList(list, separator) {
 
 /**
  * Given an image object with a url, changes the url to use dynamic thumbnailer and https.
- * 
+ *
  * Note: A dynamic thumbnailer url generated with atLeastAsLarge = true returns an image that is
  * at least as large in one dimension of the desired size. In other words, the returned image will
  * be at least as large, and as close as possible to, the largest image that is contained within a
  * box of the desired size dimensions.
- * 
+ *
  * If atLeastAsLarge = false, the dynamic thumbnailer url will give the largest image that is
  * smaller than the desired size in both dimensions.
- * 
+ *
  * @param {Object} simpleOrComplexImage An image object with a url
  * @param {string} desiredSize The desired size of the image ('<width>x<height>')
  * @param {boolean} atLeastAsLarge Whether the image should be at least as large as the desired
@@ -321,12 +321,40 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
     throw new Error(`Object of type boolean expected. Got ${typeof atLeastAsLarge}.`);
   }
 
-  let desiredWidth, desiredHeight;
-  let desiredDims = desiredSize.split('x');
+  const isEuImage = image.url.includes('eu.mktgcdn.com');
 
-  const [urlWithoutExtension, extension] = _splitStringOnIndex(image.url, image.url.lastIndexOf('.'));
+  const dynamicUrl = isEuImage
+    ? _getEuImageDynamicUrl(image, desiredSize, atLeastAsLarge)
+    : _getUsImageDynamicUrl(image.url, desiredSize, atLeastAsLarge);
+
+  return Object.assign(
+    {},
+    image,
+    {
+      url: dynamicUrl.replace('http://', 'https://')
+    }
+  );
+}
+
+/**
+ * Given a US image url, returns the dynamic url.
+ *
+ * @param {string} imageUrl Image's url (e.g.
+ *   'https://dynl.mktgcdn.com/p/ldMLwj1JkN94-2pwh6CjR_OMy4KnexHJCfZhPAZCbi0/196x400.jpg',
+ *   'https://a.mktgcdn.com/p/ldMLwj1JkN94-2pwh6CjR_OMy4KnexHJCfZhPAZCbi0/196x400.jpg')
+ * @param {string} desiredSize The desired size of the image ('<width>x<height>')
+ * @param {boolean} atLeastAsLarge Whether the image should be at least as large as the desired
+ *                                 size in one dimension or smaller than the desired size in both
+ *                                 dimensions.
+ * @returns {string} A dynamic url (e.g. 'https://dynl.mktgcdn.com/p/ldMLwj1JkN94-2pwh6CjR_OMy4KnexHJCfZhPAZCbi0/200x1.jpg')
+ */
+function _getUsImageDynamicUrl(imageUrl, desiredSize, atLeastAsLarge) {
+  const [urlWithoutExtension, extension] = _splitStringOnIndex(imageUrl, imageUrl.lastIndexOf('.'));
   const [urlBeforeDimensions, dimensions] = _splitStringOnIndex(urlWithoutExtension, urlWithoutExtension.lastIndexOf('/') + 1);
   const fullSizeDims = dimensions.split('x');
+
+  let desiredWidth, desiredHeight;
+  let desiredDims = desiredSize.split('x');
 
   if (desiredDims[0] !== '') {
     desiredWidth = Number.parseInt(desiredDims[0]);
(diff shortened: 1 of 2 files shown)


static/js/formatters-internal.js
```diff
@@ -321,13 +321,41 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
+    {
+      url: dynamicUrl.replace('http://', 'https://')
+    }
+  );
+}
+
+/**
+ * Given a US image url, returns the dynamic url.
+ *
+ * @param {string} imageUrl Image's url (e.g.
```
benmcginnis, line 342: @EmilyZhang777 it's somewhat hard for me to tell from this diff, but are we repeating the logic to deal with the dimensions separately for the EU and US functions?  If we are, could we refactor to separate those concerns so that the dimension logic is properly encapsulated?

EmilyZhang777: I actually tried to refactor this part so they can be reused. But we'll need to handle atLeastAsLarge differently between EU vs US, and it made things unnecessarily complicated that way.

nmanu1: is the difference that the EU url has format options at the end as params rather than as `<width>x<height>` dimensions?

EmilyZhang777: Yes. EU also handles atLeastAsLarge differently as the dynl/dynm domain only supports US images

benmcginnis: Ah ok, that's unfortunate!  Do the US urls not let you use the EU syntax? that seems like an odd choice.

EmilyZhang777: US images can use dynl/dynm.mktgcdn.com to handle the different fit settings, which is not available for EU. More details here: https://hitchhikers.yext.com/docs/content/image-storing-and-serving/
I CAN switch US images off of them. But I think it's preferred to use them if available.

benmcginnis: Ok sounds good

benmcginnis (approved):

static/js/formatters-internal.js
```diff
@@ -348,22 +376,80 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
+/**
+ * Given an EU image url, returns the dynamic url.
+ *
+ * @param {Object} image The image object. (e.g.
+ * {
+ *   url: 'https://a.eu.mktgcdn.com/f/0/FLVfkpR1IwpWrWDuyNYCJWVYIDfPO6x1QSztXozMIzo.jpg',
+ *   sourceUrl: 'https://a.mktgcdn.com/p/UN9RPhz0V9D8bNZ3XfNpkGnAk6ikFhVmgvntlBjVyMA/1200x675.jpg',
+ *   width: 1200,
+ *   height: 675,
+ * })
```
nmanu1, line 393: could the image object ever have a different shape? I noticed the param for the formatter is called `simpleOrComplexImage`, but wasn't sure if the main difference between those is handled in line 304, or if there are other things we need to check for

static/js/formatters-internal.js
```diff
@@ -348,22 +376,80 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
+  if (image.height) {
+    fullSizeHeight = image.height;
+  }
+
+  if (image.sourceUrl && (!fullSizeWidth || !fullSizeHeight)) {
+    const [urlWithoutExtension, _] = _splitStringOnIndex(image.sourceUrl, image.sourceUrl.lastIndexOf('.'));
+    const [__, dimensions] = _splitStringOnIndex(urlWithoutExtension, urlWithoutExtension.lastIndexOf('/') + 1);
+    const fullSizeDims = dimensions.split('x');
+
+    fullSizeWidth = Number.parseInt(fullSizeDims[0]);
```
nmanu1, line 414: if `image.width` is present (but not `image.height`), would we still want to replace the `fullSizeWidth` value?

static/js/formatters-internal.js
EmilyZhang777: I didn't update the logic there. The only difference between a US and EU image is the format of the url so I don't think we'll need to anything extra. I've checked both a EU and US site, and made sure the images on both of them load correctly.

static/js/formatters-internal.js
```diff
@@ -348,22 +376,80 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
+
+  if (desiredDims[0] !== '') {
+    const desiredWidth = Number.parseInt(desiredDims[0]);
+    if (Number.isNaN(desiredWidth)) {
+      throw new Error("Invalid width specified");
     }
-  );
+
+    formatOptions.push(`width=${desiredWidth}`);
+  } else if (!atLeastAsLarge && fullSizeWidth) {
```
nmanu1, line 428: to double check, we don't want to add the `width` param unless the image should be contained by that width?

tests/static/js/formatters-internal/image.js
```diff
@@ -1,56 +1,106 @@
...
+        { url: euUrl,
+          width: 1024,
+          sourceUrl: 'https://a.mktgcdn.com/p/FOO/516x384.jpg'
+        }, 'x', false).url;
+      expect(euImageUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/width=516,height=384,fit=contain');
+    });
+
+    it('when choosing the smallest image over threshold, omit width/height if can\'t parse it from the image object', () => {
+      const euImageUrl = Formatters.image({url: euUrl}, 'x', true).url;
+      expect(euImageUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/fit=pad');
```
nmanu1, line 103: omitting the width/height dimensions with `fit=pad` returns the smallest image? I would have expected it to return the largest image because there are no bounds

static/js/formatters-internal.js
EmilyZhang777: Just double checked. A url would always be there for simple/complex images in US/EU and the format match what we expect in the code

static/js/formatters-internal.js
EmilyZhang777: yep. I was trying to match what we are doing for US images based on my understanding.

nmanu1: but I think the US images add a size of 1 for any missing dimension when `atLeastAsLarge=true`

static/js/formatters-internal.js
EmilyZhang777: Yep. I have a test case to cover that ("when choosing the biggest image under threshold, use dimensions from the sourceUrl if width/height does not exist")

nmanu1: that test case has a smaller width in the `sourceUrl` than in the `image.width`, so that makes sense to me. but, if the `image.width` were smaller than `sourceUrl`, would we want to use the `image.width` instead?

tests/static/js/formatters-internal/image.js
EmilyZhang777: Taking another look at this. I think `cover` would be a better option. Updated

static/js/formatters-internal.js
nmanu1: actually, I think the current behavior is correct for EU. looks like `dynm`/`dynl` end up ignoring the dimensions with size 1 for US images anyways ([thread](https://yext.slack.com/archives/C01AFQG2ASH/p1633797078031300?thread_ts=1633115673.019900&cid=C01AFQG2ASH))

static/js/formatters-internal.js
EmilyZhang777: My concern with using one dimension from image.width/image.height and the other dimension from sourceUrl is that the ratio might be off. We just use this as a fallback. So if the client wants to have a precise dimension, they can always specify it with the desiredDims

EmilyZhang777: I believe the dimensions in the `sourceUrl` would always match `image.width` and `image.height`. This is just to make sure we wouldn't get any errors from `image.width/height` not existing

PR #1171, opened 2024-02-20
Version 1.33.5

nmanu1:
### Fixes
- Resolve vulnerabilities (#1167)
- Remove `span` styling and bump answers-search-ui version to 1.17 so icon `div`s are changed to `span`s without affecting their styling (#1169, #1170)
  - Note, any custom styling that is applied by targeting `div`s or `span`s directly may be impacted by this change. This could result in previously applied styles no longer being applied to these icons, or other styles inadvertently being applied to the icons


static/js/theme-map/Util/Accessibility.js
benmcginnis, line 1: did we mean to delete this file?

nmanu1: yeah, it wasn't being used anywhere. it contained an insecure document method usage that Liah was having a hard time verifying, so I recommended she just delete it

benmcginnis (approved):

PR #1173, opened 2024-03-04
Vulnerabilities 2

EmilyZhang777:
Remove insecure document method and upgrade jambo to v1.12.5

J=VULN-37757

Spun up and verified test site. Although ElementRenderTarget doesn't seem to be used.


static/js/theme-map/Renderer/ElementRenderTarget.js
```diff
@@ -47,11 +47,11 @@ class ElementRenderTarget extends RenderTarget {
 
   /**
    * async render(data) => DOMNode
-   * Calls templateFunction(data) and sets value to element.innerHTML, then returns element
+   * Calls templateFunction(data) and sets value to element.textContnet, then returns element
```
vijay267, line 50: Seems like we have a typo here? I imagine it's supposed to be textContent right?

vijay267 (commented):
typo in "textContent"

static/js/theme-map/Renderer/ElementRenderTarget.js
```diff
@@ -47,11 +47,11 @@ class ElementRenderTarget extends RenderTarget {
...
    */
   async render(data) {
     if (this._element) {
-      this._element.innerHTML = this._templateFunction(data);
+      this._element.textContnet = this._templateFunction(data);
```
nmanu1, line 54: can you manually use this function if it's not used by the test site and verify the behavior is unchanged? I think right now it would fail because of the typo in `textContent`

static/js/theme-map/Renderer/ElementRenderTarget.js
EmilyZhang777: Nice catch! Updated

static/js/theme-map/Renderer/ElementRenderTarget.js
EmilyZhang777: fixed and tested it's working

static/js/theme-map/Renderer/ElementRenderTarget.js
nmanu1: looks like the type in this doc is still not updated

static/js/theme-map/Renderer/ElementRenderTarget.js
nmanu1: cool! out of curiosity, what did the function do when you called it? did it just run the function you provided and set the return value as the element content?

EmilyZhang777: yep, it's used to update the element when data changes

static/js/theme-map/Renderer/ElementRenderTarget.js
EmilyZhang777: lol, updated

nmanu1 (commented):
are we planning on releasing a new version for this fix?

EmilyZhang777:
> are we planning on releasing a new version for this fix?

I was thinking about putting it in a hotfix. What do you think?

nmanu1:
> I was thinking about putting it in a hotfix. What do you think?

yeah, that works. we'll want to merge into a hotfix branch instead of develop then. we can also bump the jambo version to get the most recent vulnerability fixes

EmilyZhang777:
> > I was thinking about putting it in a hotfix. What do you think?
> 
> yeah, that works. we'll want to merge into a hotfix branch instead of develop then. we can also bump the jambo version to get the most recent vulnerability fixes

updated!

nmanu1 (commented):
To bump jambo, you should update both the top-level package.json and static/package.json, as well as their respective package-locks

PR #1176, opened 2024-03-07
Version 1.33.6

EmilyZhang777:
### Fixes
- Remove insecure document method and upgrade jambo to v1.12.5 (#1173)


nmanu1 (approved):
:fire:

PR #1181, opened 2024-04-11
Version 1.33.7

nmanu1:
### Fixes
- Update the vertical full page map so it keeps location filters during pagination (#1180)


sahilvaidya (approved):
Makes sense to me!

PR #1184, opened 2024-04-19
Update Image Formatter To Work With All Image Url Formats

EmilyZhang777:
This change updates the image handling to accommodate all image url formats before, during and after the image url change.

Related thread: https://yext.slack.com/archives/C0DA5HJCT/p1713445795909829

J=WAT-3980
TEST=auto, manual

Ran `npm run test`.
Manually copied my change to two answers experience (one in US and one in EU). Confirmed the images in both experiences loads fine in Live Preview.

### DIFF
--- a/static/js/formatters-internal.js
+++ b/static/js/formatters-internal.js
@@ -311,110 +311,52 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
   if (!(Object.prototype.toString.call(image).indexOf('Object') > 0)) {
     throw new Error("Expected parameter of type Map");
   }
-  if ((typeof desiredSize !== 'string') || (desiredSize == null)) {
+  if (typeof desiredSize !== 'string') {
     throw new Error(`Object of type string expected. Got ${typeof desiredSize}.`);
   }
   if (desiredSize.indexOf('x') === -1) {
     throw new Error("Invalid desired size");
   }
-  if ((typeof atLeastAsLarge !== 'boolean') || (atLeastAsLarge == null)) {
+  if (typeof atLeastAsLarge !== 'boolean') {
     throw new Error(`Object of type boolean expected. Got ${typeof atLeastAsLarge}.`);
   }
 
-  const isEuImage = image.url.includes('eu.mktgcdn.com');
-
-  const dynamicUrl = isEuImage
-    ? _getEuImageDynamicUrl(image, desiredSize, atLeastAsLarge)
-    : _getUsImageDynamicUrl(image.url, desiredSize, atLeastAsLarge);
-
-  return Object.assign(
-    {},
-    image,
-    {
-      url: dynamicUrl.replace('http://', 'https://')
+  let url;
+  try {
+    url = new URL(image.url);
+    let urlPath = url.pathname;
+    if (url.pathname.match('^\/p')) {
+      urlPath = _removePhotoImageUrlExtension(urlPath);
     }
-  );
+
+    const hostname = url.hostname.replace(/^[a-z]+\./, 'dyn.');
+    const formatOptionsString = _getImageFormatOptions(desiredSize, atLeastAsLarge, image.width, image.height);
+
+    return Object.assign(
+      {},
+      image,
+      {
+        url: `https://${hostname}${urlPath}${formatOptionsString}`,
+      }
+    );
+
+  } catch (error) {
+    throw new Error(`Error processing image url ${image.url}: ${error}`);
+  }
 }
 
 /**
- * Given a US image url, returns the dynamic url.
+ * Construct the format options string with given parameters.
  *
- * @param {string} imageUrl Image's url (e.g.
- *   'https://dynl.mktgcdn.com/p/ldMLwj1JkN94-2pwh6CjR_OMy4KnexHJCfZhPAZCbi0/196x400.jpg',
- *   'https://a.mktgcdn.com/p/ldMLwj1JkN94-2pwh6CjR_OMy4KnexHJCfZhPAZCbi0/196x400.jpg')
  * @param {string} desiredSize The desired size of the image ('<width>x<height>')
  * @param {boolean} atLeastAsLarge Whether the image should be at least as large as the desired
  *                                 size in one dimension or smaller than the desired size in both
  *                                 dimensions.
- * @returns {string} A dynamic url (e.g. 'https://dynl.mktgcdn.com/p/ldMLwj1JkN94-2pwh6CjR_OMy4KnexHJCfZhPAZCbi0/200x1.jpg')
+ * @param {number?} fullSizeWidth The full size width of the original image
+ * @param {number?} fullSizeHeight The full size height of the original image
+ * @returns {string} A string representing the format options (e.g. 'height=200,width=100,fit=cover')
  */
-function _getUsImageDynamicUrl(imageUrl, desiredSize, atLeastAsLarge) {
-  const [urlWithoutExtension, extension] = _splitStringOnIndex(imageUrl, imageUrl.lastIndexOf('.'));
-  const [urlBeforeDimensions, dimensions] = _splitStringOnIndex(urlWithoutExtension, urlWithoutExtension.lastIndexOf('/') + 1);
-  const fullSizeDims = dimensions.split('x');
-
-  let desiredWidth, desiredHeight;
-  let desiredDims = desiredSize.split('x');
-
-  if (desiredDims[0] !== '') {
-    desiredWidth = Number.parseInt(desiredDims[0]);
-    if (Number.isNaN(desiredWidth)) {
-      throw new Error("Invalid width specified");
-    }
-  } else {
-    desiredWidth = atLeastAsLarge ? 1 : Number.parseInt(fullSizeDims[0]);
-  }
-
-  if (desiredDims[1] !== '') {
-    desiredHeight = Number.parseInt(desiredDims[1]);
-    if (Number.isNaN(desiredHeight)) {
-      throw new Error("Invalid height specified");
-    }
-  } else {
-    desiredHeight = atLeastAsLarge ? 1 : Number.parseInt(fullSizeDims[1]);
-  }
-
-  const urlWithDesiredDims = urlBeforeDimensions + desiredWidth + 'x' + desiredHeight + extension;
-
-  return atLeastAsLarge
-    ? _replaceUrlHost(urlWithDesiredDims, 'dynl.mktgcdn.com')
-    : _replaceUrlHost(urlWithDesiredDims, 'dynm.mktgcdn.com');
-}
-
-/**
- * Given an EU image url, returns the dynamic url.
- *
- * @param {Object} image The image object. (e.g.
- * {
- *   url: 'https://a.eu.mktgcdn.com/f/0/FLVfkpR1IwpWrWDuyNYCJWVYIDfPO6x1QSztXozMIzo.jpg',
- *   sourceUrl: 'https://a.mktgcdn.com/p/UN9RPhz0V9D8bNZ3XfNpkGnAk6ikFhVmgvntlBjVyMA/1200x675.jpg',
- *   width: 1200,
- *   height: 675,
- * })
- * @param {string} desiredSize The desired size of the image ('<width>x<height>')
- * @param {boolean} atLeastAsLarge Whether the image should be at least as large as the desired
- *                                 size in one dimension or smaller than the desired size in both
- *                                 dimensions.
- * @returns {string} A dynamic url (e.g. 'https://dyn.eu.mktgcdn.com/f/0/FLVfkpR1IwpWrWDuyNYCJWVYIDfPO6x1QSztXozMIzo.jpg/width=200,fit=contain')
- */
-function _getEuImageDynamicUrl(image, desiredSize, atLeastAsLarge) {
-  let fullSizeWidth, fullSizeHeight;
-  if (image.width) {
-    fullSizeWidth = image.width;
-  }
-  if (image.height) {
-    fullSizeHeight = image.height;
-  }
-
-  if (image.sourceUrl && (!fullSizeWidth || !fullSizeHeight)) {
-    const [urlWithoutExtension, _] = _splitStringOnIndex(image.sourceUrl, image.sourceUrl.lastIndexOf('.'));
-    const [__, dimensions] = _splitStringOnIndex(urlWithoutExtension, urlWithoutExtension.lastIndexOf('/') + 1);
-    const fullSizeDims = dimensions.split('x');
-
-    fullSizeWidth = Number.parseInt(fullSizeDims[0]);
-    fullSizeHeight = Number.parseInt(fullSizeDims[1]);
-  }
-
+function _getImageFormatOptions(desiredSize, atLeastAsLarge, fullSizeWidth, fullSizeHeight) {
   let desiredDims = desiredSize.split('x');
   let formatOptions = [];
 
@@ -442,33 +384,21 @@ function _getEuImageDynamicUrl(image, desiredSize, atLeastAsLarge) {
 
   formatOptions.push(`fit=${atLeastAsLarge ? 'cover' : 'contain'}`);
 
-  const urlWithOptions = image.url + `/${formatOptions.join(',')}`;
-
-  return _replaceUrlHost(urlWithOptions, 'dyn.eu.mktgcdn.com');
+  return `/${formatOptions.join(',')}`;
 }
 
 /**
- * Splits a string into two parts at the specified index.
+ * Given a photo image url path, remove the trailing extension.
  *
- * @param {string} str The string to be split
- * @param {number} index The index at which to split the string
- * @returns {Array<string>} The two parts of the string after splitting
+ * @param {string} imageUrlPath Image's url path (e.g.
+ *   '/p/mFsjqWGQEOMQGNoNIcnq61JtdSGiCs/225x225.jpg',
+ *   '/p-sandbox/mFsjqWGQEOMQGNoNIcnq61JtdSGiCs/225x225.jpg')
+ * @returns {string} A canonicalized image url path (e.g.
+ *   '/p/mFsjqWGQEOMQGNoNIcnq61JtdSGiCs',
+ *   '/p-sandbox/mFsjqWGQEOMQGNoNIcnq61JtdSGiCs')
  */
-function _splitStringOnIndex(str, index) {
-  return [str.slice(0, index), str.slice(index)];
-}
-
-/**
- * Replaces the current host of a url with the specified host.
- *
- * @param {string} url The url whose host is to be changed
- * @param {string} host The new host to change to
- * @returns {string} The url updated with the specified host
- */
-function _replaceUrlHost(url, host) {
-  const splitUrl = url.split('://');
-  const urlAfterHost = splitUrl[1].slice(splitUrl[1].indexOf('/'));
-  return splitUrl[0] + '://'  + host + urlAfterHost;
+function _removePhotoImageUrlExtension(imageUrlPath) {
+  return imageUrlPath.replace(/(\/[0-9]+x[0-9]+\.[a-z]+)|(\/)$/, '');
 }
 
 /**
--- a/tests/static/js/formatters-internal/image.js
+++ b/tests/static/js/formatters-internal/image.js
@@ -1,106 +1,159 @@
 import Formatters from 'static/js/formatters.js';
 
 describe('image formatter', () => {
-  const usUrl = 'https://a.mktgcdn.com/p/1024x768.jpg';
-  const euUrl = 'https://dyn.eu.mktgcdn.com/f/0/FOO.jpg';
-  const usImg = {url: usUrl};
-  const euImg = {
-    url: euUrl,
+  const photoUrl = 'https://a.mktgcdn.com/p/FOO/1024x768.jpg';
+  const oldFileUrl = 'https://a.mktgcdn.com/f/0/FOO.jpg';
+  const newFileUrl = 'https://a.mktgcdn.com/f/FOO.jpg';
+  const euFileUrl = 'https://a.eu.mktgcdn.com/f/0/FOO.jpg';
+
+  const photoImg = {
+    url: photoUrl,
     width: 1024,
     height: 768,
-    sourceUrl: 'https://a.mktgcdn.com/p/FOO/1024x768.jpg',
+  };
+
+  const oldFileImg = {
+    url: oldFileUrl,
+    width: 1024,
+    height: 768,
+  };
+
+  const newFileImg = {
+    url: newFileUrl,
+    width: 1024,
+    height: 768,
+  };
+
+  const euFileImg = {
+    url: euFileUrl,
+    width: 1024,
+    height: 768,
   }
 
   describe('when choosing the smallest image over threshold', () => {
     it('By default chooses the smallest image with width >= 200', () => {
-      const usImageUrl = Formatters.image(usImg).url;
-      expect(usImageUrl).toEqual('https://dynl.mktgcdn.com/p/200x1.jpg');
-      const euImageUrl = Formatters.image(euImg).url;
-      expect(euImageUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/width=200,fit=cover');
+      const photoImgUrl = Formatters.image(photoImg).url;
+      expect(photoImgUrl).toEqual('https://dyn.mktgcdn.com/p/FOO/width=200,fit=cover');
+      const oldFileImgUrl = Formatters.image(oldFileImg).url;
+      expect(oldFileImgUrl).toEqual('https://dyn.mktgcdn.com/f/0/FOO.jpg/width=200,fit=cover');
+      const newFileImgUrl = Formatters.image(newFileImg).url;
+      expect(newFileImgUrl).toEqual('https://dyn.mktgcdn.com/f/FOO.jpg/width=200,fit=cover');
+      const euFileImgUrl = Formatters.image(euFileImg).url;
+      expect(euFileImgUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/width=200,fit=cover');
     });
 
     it('Can restrict the dimensions by width', () => {
-      const usImageUrl = Formatters.image(usImg, '601x').url;
-      expect(usImageUrl).toEqual('https://dynl.mktgcdn.com/p/601x1.jpg');
-      const euImageUrl = Formatters.image(euImg, '601x').url;
-      expect(euImageUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/width=601,fit=cover');
+      const photoImgUrl = Formatters.image(photoImg, '601x').url;
+      expect(photoImgUrl).toEqual('https://dyn.mktgcdn.com/p/FOO/width=601,fit=cover');
+      const oldFileImgUrl = Formatters.image(oldFileImg, '601x').url;
+      expect(oldFileImgUrl).toEqual('https://dyn.mktgcdn.com/f/0/FOO.jpg/width=601,fit=cover');
+      const newFileImgUrl = Formatters.image(newFileImg, '601x').url;
+      expect(newFileImgUrl).toEqual('https://dyn.mktgcdn.com/f/FOO.jpg/width=601,fit=cover');
+      const euFileImgUrl = Formatters.image(euFileImg, '601x').url;
+      expect(euFileImgUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/width=601,fit=cover');
     });
 
     it('Can restrict the dimensions by height', () => {
-      const usImageUrl = Formatters.image(usImg, 'x338').url;
-      expect(usImageUrl).toEqual('https://dynl.mktgcdn.com/p/1x338.jpg');
-      const euImageUrl = Formatters.image(euImg, 'x338').url;
-      expect(euImageUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/height=338,fit=cover');
+      const photoImgUrl = Formatters.image(photoImg, 'x338').url;
+      expect(photoImgUrl).toEqual('https://dyn.mktgcdn.com/p/FOO/height=338,fit=cover');
+      const oldFileImgUrl = Formatters.image(oldFileImg, 'x338').url;
+      expect(oldFileImgUrl).toEqual('https://dyn.mktgcdn.com/f/0/FOO.jpg/height=338,fit=cover');
+      const newFileImgUrl = Formatters.image(newFileImg, 'x338').url;
+      expect(newFileImgUrl).toEqual('https://dyn.mktgcdn.com/f/FOO.jpg/height=338,fit=cover');
+      const euFileImgUrl = Formatters.image(euFileImg, 'x338').url;
+      expect(euFileImgUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/height=338,fit=cover');
     });
 
     it('Can restrict by both dimensions', () => {
-      const usImageUrl = Formatters.image(usImg, '601x338').url;
-      expect(usImageUrl).toEqual('https://dynl.mktgcdn.com/p/601x338.jpg');
-      const euImageUrl = Formatters.image(euImg, '601x338').url;
-      expect(euImageUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/width=601,height=338,fit=cover');
+      const photoImgUrl = Formatters.image(photoImg, '601x338').url;
+      expect(photoImgUrl).toEqual('https://dyn.mktgcdn.com/p/FOO/width=601,height=338,fit=cover');
+      const oldFileImgUrl = Formatters.image(oldFileImg, '601x338').url;
+      expect(oldFileImgUrl).toEqual('https://dyn.mktgcdn.com/f/0/FOO.jpg/width=601,height=338,fit=cover');
+      const newFileImgUrl = Formatters.image(newFileImg, '601x338').url;
+      expect(newFileImgUrl).toEqual('https://dyn.mktgcdn.com/f/FOO.jpg/width=601,height=338,fit=cover');
+      const euFileImgUrl = Formatters.image(euFileImg, '601x338').url;
+      expect(euFileImgUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/width=601,height=338,fit=cover');
     });
 
     it('returns the smallest image when no dimensions given', () => {
-      const usImageUrl = Formatters.image(usImg, 'x').url;
-      expect(usImageUrl).toEqual('https://dynl.mktgcdn.com/p/1x1.jpg');
-      const euImageUrl = Formatters.image(euImg, 'x').url;
-      expect(euImageUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/fit=cover');
+      const photoImgUrl = Formatters.image(photoImg, 'x').url;
+      expect(photoImgUrl).toEqual('https://dyn.mktgcdn.com/p/FOO/fit=cover');
+      const oldFileImgUrl = Formatters.image(oldFileImg, 'x').url;
+      expect(oldFileImgUrl).toEqual('https://dyn.mktgcdn.com/f/0/FOO.jpg/fit=cover');
+      const newFileImgUrl = Formatters.image(newFileImg, 'x').url;
+      expect(newFileImgUrl).toEqual('https://dyn.mktgcdn.com/f/FOO.jpg/fit=cover');
+      const euFileImgUrl = Formatters.image(euFileImg, 'x').url;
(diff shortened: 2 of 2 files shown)


static/js/formatters-internal.js
```diff
@@ -442,33 +384,21 @@ function _getEuImageDynamicUrl(image, desiredSize, atLeastAsLarge) {
 
   formatOptions.push(`fit=${atLeastAsLarge ? 'cover' : 'contain'}`);
```
jfancher-yext, line 385: I realize this is existing code, but is it right? There's no exact equivalent, but I think it might be better to use `scale-down` for both (which is the default). Maybe `contain` for least would be more correct, but dynl doesn't enlarge.

But the other options mess with the aspect ratio (`cover` crops it), which dynl/dynm never do, so that doesn't seem right. See docs:
https://developers.cloudflare.com/images/transform-images/transform-via-workers/#fit

EmilyZhang777: Talked to @nmanu1 offline. The expected behavior would be:
- if both dims (50*100 or 200*450) provided and original image size were 100x150
    - atLeastAsLarge=true, return 66x100 or 300x450
    - atLeastAsLarge=false, return 50x75 or 200x300
- if one dim (width=50 or 200) provided and if the original image size were 100x150, return 50x75 or 200x300 regardless of atLeastAsLarge
- if no dims provided, return original size

Updated the code to follow that.

static/js/formatters-internal.js
```diff
@@ -311,164 +311,117 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
+ * @returns {string} A canonicalized image url path (e.g.
+ *   '/p/mFsjqWGQEOMQGNoNIcnq61JtdSGiCs',
+ *   '/p-sandbox/mFsjqWGQEOMQGNoNIcnq61JtdSGiCs')
  */
-function _replaceUrlHost(url, host) {
-  const splitUrl = url.split('://');
-  const urlAfterHost = splitUrl[1].slice(splitUrl[1].indexOf('/'));
-  return splitUrl[0] + '://'  + host + urlAfterHost;
+function _removePhotoImageUrlExtension(imageUrlPath) {
+  return imageUrlPath.replace(/(\/[0-9]+x[0-9]+\.[a-z]+)|(\/)$/, '');
```
nmanu1, line 424: is there a reason we check for the image dimensions in this regex? could we just split on the last index of `.` instead?

static/js/formatters-internal.js
```diff
@@ -311,164 +311,117 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
- * @returns {string} A dynamic url (e.g. 'https://dyn.eu.mktgcdn.com/f/0/FLVfkpR1IwpWrWDuyNYCJWVYIDfPO6x1QSztXozMIzo.jpg/width=200,fit=contain')
- */
-function _getEuImageDynamicUrl(image, desiredSize, atLeastAsLarge) {
-  let fullSizeWidth, fullSizeHeight;
-  if (image.width) {
-    fullSizeWidth = image.width;
-  }
-  if (image.height) {
-    fullSizeHeight = image.height;
+    return `/${formatOptions.join(',')}`;
```
nmanu1, line 391: can we perform this step once at the end before returning? it seems like it's repeated for each case

static/js/formatters-internal.js
```diff
@@ -311,164 +311,117 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
+    return '';
+  }
+
+  const originalRatio =
+    (!!fullSizeWidth && !!fullSizeHeight) ? (fullSizeWidth / fullSizeHeight) : undefined;
+  let desiredWidth;
+  let desiredHeight;
+  let formatOptions = ['fit=contain'];
+  if (desiredWidthProvided) {
     desiredWidth = Number.parseInt(desiredDims[0]);
```
nmanu1, line 375: why do we need both a desiredWidthProvided and a desiredWidth?

static/js/formatters-internal.js
```diff
@@ -311,164 +311,117 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
  */
-function _getUsImageDynamicUrl(imageUrl, desiredSize, atLeastAsLarge) {
-  const [urlWithoutExtension, extension] = _splitStringOnIndex(imageUrl, imageUrl.lastIndexOf('.'));
-  const [urlBeforeDimensions, dimensions] = _splitStringOnIndex(urlWithoutExtension, urlWithoutExtension.lastIndexOf('/') + 1);
-  const fullSizeDims = dimensions.split('x');
-
-  let desiredWidth, desiredHeight;
+function _getImageFormatOptions(desiredSize, atLeastAsLarge, fullSizeWidth, fullSizeHeight) {
   let desiredDims = desiredSize.split('x');
+  const desiredWidthProvided = desiredDims[0] !== '' && desiredDims[0] !== '1';
```
nmanu1, line 361: why are we adding special case logic for a size of `1`? we didn't have that previously, right?

static/js/formatters-internal.js
```diff
@@ -311,164 +311,117 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
     throw new Error(`Object of type boolean expected. Got ${typeof atLeastAsLarge}.`);
   }
 
-  const isEuImage = image.url.includes('eu.mktgcdn.com');
+  let url;
+  try {
+    url = new URL(image.url);
+    let urlPath = url.pathname;
+    if (url.pathname.match('^\/p')) {
+      urlPath = _removePhotoImageUrlExtension(urlPath);
```
nmanu1, line 329: does URL have any built in way to remove the extension?

static/js/formatters-internal.js
```diff
@@ -311,164 +311,117 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
+    }
 
-  const dynamicUrl = isEuImage
-    ? _getEuImageDynamicUrl(image, desiredSize, atLeastAsLarge)
-    : _getUsImageDynamicUrl(image.url, desiredSize, atLeastAsLarge);
+    const hostname = url.hostname.replace(/^[a-z]+\./, 'dyn.');
```
nmanu1, line 332: can we split on the first `.` here? that seems easier to understand than a regex

static/js/formatters-internal.js
```diff
@@ -311,164 +311,117 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
+  const desiredHeightProvided = desiredDims[1] !== '' && desiredDims[1] !== '1';
 
-  if (desiredDims[0] !== '') {
+  // both dimensions are not provided, return original image
+  if (!desiredWidthProvided && !desiredHeightProvided) {
+    return '';
+  }
+
+  const originalRatio =
+    (!!fullSizeWidth && !!fullSizeHeight) ? (fullSizeWidth / fullSizeHeight) : undefined;
```
nmanu1, line 370: in what cases might the width and height of the image not be provided?

static/js/formatters-internal.js
```diff
@@ -311,164 +311,117 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
+  let desiredWidth;
+  let desiredHeight;
+  let formatOptions = ['fit=contain'];
```
nmanu1, line 373: don't we want this to be a cover if atLeastAsLarge=true?

test-site/scripts/build.sh
```diff
@@ -7,4 +7,6 @@ set_working_dir_to_test_site () {
 
 set_working_dir_to_test_site
 
-npx jambo build && npx grunt webpack
+npx jambo build
+chmod u+x ./public/static/node_modules/esbuild/bin/esbuild
```
nmanu1, line 11: I don't think we need this step for the script. ideally, there shouldn't be a node_modules folder present in the static dir

static/js/formatters-internal.js
EmilyZhang777: Per Jacob's comment [here](https://github.com/yext/answers-hitchhiker-theme/pull/1184/files#r1574716422), previously, if people pass in "200x1", we'd render the height as the original height. We wouldn't want any existing use cases to be rendered as 1 px. Ack

static/js/formatters-internal.js
jfancher-yext: Per my previous comment, `cover` crops the image rather than respecting the aspect ratio. I do not imagine that is the desired behavior in any case (and it doesn't match the old dynl).

static/js/formatters-internal.js
EmilyZhang777: I wanted to break early and avoid nested is/else. Ack

static/js/formatters-internal.js
jfancher-yext: Do you mean the last index of `/`?

I almost did comment that photo urls ending with `/original` do exist (no width/height and no extension), but you should never see one in entity data so it didn't seem that relevant. This should be safe in practice, but just getting the directory name would also work.

static/js/formatters-internal.js
nmanu1: By passing `contain` as the formatting option, the image will be shrunk until it fits within the size boundaries. When a user passes in atLeastAsLarge, they are looking for an image size that fully covers the desired dimensions. It's actually a pretty common use case (e.g. for a hero image or banner) to want the image to take up the whole space (or more) rather than be too small and show empty whitespace.

Because the user's desired size may not be the same as the aspect ratio, it's not always possible to both preserve the aspect ratio and keep the whole image while fully covering the area. In these scenarios, we'd return `cover` (because dynl doesn't change the aspect ratio of the image or crop it), and it would be preferable to have an image with the correct aspect ratio that gets cropped than the opposite.

jfancher-yext: I see, that makes sense. Well, as long as the difference in behavior from dynl is understood, that's fine.

static/js/formatters-internal.js
EmilyZhang777: Seems like using a regex is safer just in case we get a url without width/height and extension? Keeping it for now

nmanu1: ah yes, I meant the last `/`.

could a photo url ever end in `/`? and maybe I'm misunderstanding the last part of the regex, but why would we want to remove any arbitrary last `/` if it doesn't match the `/200x200.jpg` format? also, do we get the height/width information through some other means now if we remove it from the url before doing the dimension parsing? is it guaranteed to always be a part of the image object?

EmilyZhang777: I just wanted to be sure there's no trailing `/` before we append the format options

jfancher-yext: > is it guaranteed to always be a part of the image object?

Size should always be returned for image objects. There have been bugs over the years where the metadata didn't get recorded, but to the best of my knowledge they've been fixed. So, width/height is _expected_ to be on any image-type entity field-value at least.

(It's also possible to upload an image to a generic File-type field; the entity value wouldn't have image metadata in this case, but also the URL wouldn't necessarily adhere to the same format anyway)

static/js/formatters-internal.js
nmanu1: I think I got mixed up with the EU and US handling and thought we were already doing a cover/crop. It looks like we only do that for EU, and use dynl for US, so one would crop and one wouldn't. To be consistent, we should pick one. I'm leaning towards keeping what we do for US (i.e. don't crop) because that's where most of our current users are. But, it depends on the behavior of dyn. @jfancher-yext, what does dyn return if, via the URL, we pass in:
- a width and a height that don't match the aspect ratio, and no `fit` param
- one dimension (either width or height), and no `fit` param
- no width or height or fit

static/js/formatters-internal.js
nmanu1: would this regex remove `/200x200.jpg/` if there was a trailing slash after the dimensions and extension?

and @jfancher-yext, is it common to upload an image to a generic file-type field? and do you know what format the URL would have? trying to determine if this is something we need support for, and if so, how much work it would be

static/js/formatters-internal.js
nmanu1: if this is a boolean, can the name start with `is` or `has` (e.g. `hasDesiredWidth` or `isDesiredWidthProvided`)?

static/js/formatters-internal.js
jfancher-yext: The default `fit` is `scale-down` (not by us, we don't set anything, but it's the apparent default of Cloudflare).

* If width and height are provided, it will choose the largest AR-conforming size that is <= the requested size in both dimensions and <= the base image size. (Or equivalently, the smallest size that is >= the requested size in at least one dimension, but not larger than the base image). Basically the same as dynl except it's continuous instead of a set of fixed sizes.
* If one dimension is given, the largest AR-confirming size that is <= the requested size in that dimension and <= the base image size. Again, like dynl if the other dimension is `1`.
* It is an error to provide no arguments. If you provide other arguments but not width/height/fit then it'll just use the base image size, but if you provide nothing it'll fail at the parsing step. This is probably an oversight, not sure.

static/js/formatters-internal.js
nmanu1: In that comment, Jacob says that only dynl has the special behavior, not dynm. And it makes sense that 1 would be treated as no dim for dynl because the difference is just saying "any size" vs. "any size >= 1 px", which is pretty much the same. But, if a user passed in 1 with atLeastAsLarge=false, then I think they'd expect to have an image returned with size of 1px (although that seems like a rare case). Even for atLeastAsLarge=true, I'd rather not add in special case logic like this, since it doesn't seem needed in my opinion. But, let me know if I'm missing something!

nmanu1: And another question @jfancher-yext, what happened if you passed `1x1` to dynl? Would it return the smallest image or the original size image?

test-site/scripts/build.sh
```diff
@@ -8,3 +8,4 @@ set_working_dir_to_test_site () {
 set_working_dir_to_test_site
 
 npx jambo build && npx grunt webpack
+npx grunt webpack
```
nmanu1, line 11: looks like there's an extra step here still

static/js/formatters-internal.js
jfancher-yext: It is not common in my experience, though the [docs](https://hitchhikers.yext.com/docs/content/image-storing-and-serving/) do mention it as a possibility (n.b. these docs are slightly incorrect otherwise, will be looking to get them corrected as part of this project).

Images as Files would be the same as all images are in the EU except with the US domain -- `a.mktgcdn.com/f[-env]/id/hash.ext`. If this is supported, ideally the id segment would be optional, i.e., also allow `a.mktgcdn.com/f[-env]/hash.ext`. We would like to remove that segment; due to hard-to-change dependencies it's not going to happen any time soon, so it's not urgent, but in any case to make the dynamic urls for `/f/` inputs it's a simple algorithm: replace `a.` with `dyn.` and add `/width=x,height=y,etc` to the end; no need to worry about the exact structure beyond that.

(As we consolidate the two systems we're going to be changing this so that images are stored as a `/p/...` URL in the US even if they comes in via the generic File API, but existing file images won't move)

static/js/formatters-internal.js
jfancher-yext: I don't think that's specifically-defined or documented behavior, but yes, it'd return the smallest available size in that case. There have been no such requests in the past month at least (well, except for images that are actually 1x1 base size) so I don't think it is worth worrying about.

static/js/formatters-internal.js
nmanu1: > - If width and height are provided, it will choose the largest AR-conforming size that is <= the requested size in both dimensions and <= the base image size. (Or equivalently, the smallest size that is >= the requested size in at least one dimension, but not larger than the base image). Basically the same as dynl except it's continuous instead of a set of fixed sizes.

Isn't the first description (largest size smaller than requested) describing dynm and the second (smallest size larger than requested) a description for dynl? And similarly, isn't the second bullet describing dynm, not dynl?

And just to confirm, dyn returns exact pixel match (fully dynamic) rather than sizes chosen from certain thumbnails like dynl and dynm do?

static/js/formatters-internal.js
nmanu1: Thanks! In that case, there doesn't really seem to be anything special about `1` as a value for dynl, so I'm in favor of removing the special case logic

static/js/formatters-internal.js
nmanu1: Got it. So it sounds like the URL formatting should already be handled by the changes in this PR. But, if the entity has no image metadata and the URL doesn't include the dimensions, is there any way to get the original image size?

jfancher-yext: If it doesn't have metadata (which again would be a bug if it happened for an Image-type field, but expected for a File-type field) then no, it's not possible short of actually requesting the image.

nmanu1: Ah ok, so sounds like it wouldn't really be possible (or at least pretty difficult) for us to support this fully. Since it's not that common, I think we can continue without supporting it unless we get a feature request for it

static/js/formatters-internal.js
nmanu1: Thanks for the clarification! In that case, we can stick with `fit=contain` or switch between `contain` and `cover`. Functionally, they'll be the same. We just have to be careful to always specify the correct dimensions when the user passes in both a desired width and height with atLeastAsLarge=true

static/js/formatters-internal.js
```diff
@@ -311,164 +317,122 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
-
-  let desiredWidth, desiredHeight;
+function _getImageFormatOptions(desiredSize, atLeastAsLarge, fullSizeWidth, fullSizeHeight) {
   let desiredDims = desiredSize.split('x');
+  const hasDesiredWidth = desiredDims[0] !== '';
+  const hasDesiredHeight = desiredDims[1] !== '';
 
-  if (desiredDims[0] !== '') {
+  // both dimensions are not provided, return original image
+  if (!hasDesiredWidth && !hasDesiredHeight) {
```
nmanu1, line 371: can we update the logic in this function to follow a simplified flow like this? (apologies for the rough pseudocode)
```
const formatOptions = ['fit=contain'];
if desiredWidth && desiredHeight && originalRatio
    const mathFunc = atLeastAsLarge ? Math.max : Math.min
    use mathFunc and originalRatio to push width and height to formatOptions
    return joined formatOptions
if desiredWidth
    push width to formatOptions
if desiredHeight
    push height to formatOptions
return joined formatOptions
```
right now, having all the different cases is a little confusing to follow, not to mention that returning early like this without specifying any param would actually cause dyn to throw an error per Jacob's comment [here](https://github.com/yext/answers-hitchhiker-theme/pull/1184#discussion_r1577876311)

EmilyZhang777: I kept parsing the desiredWidth/Height number and short circuiting to be clearer. Also, I tested that without any options, the original image would be returned. Done

static/js/formatters-internal.js
```diff
@@ -311,164 +317,122 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
-function _getUsImageDynamicUrl(imageUrl, desiredSize, atLeastAsLarge) {
-  const [urlWithoutExtension, extension] = _splitStringOnIndex(imageUrl, imageUrl.lastIndexOf('.'));
-  const [urlBeforeDimensions, dimensions] = _splitStringOnIndex(urlWithoutExtension, urlWithoutExtension.lastIndexOf('/') + 1);
-  const fullSizeDims = dimensions.split('x');
-
-  let desiredWidth, desiredHeight;
+function _getImageFormatOptions(desiredSize, atLeastAsLarge, fullSizeWidth, fullSizeHeight) {
   let desiredDims = desiredSize.split('x');
+  const hasDesiredWidth = desiredDims[0] !== '';
+  const hasDesiredHeight = desiredDims[1] !== '';
```
nmanu1, line 368: I know you had these earlier from when we had the special logic for `1`. But, now that we removed that, can we removed these boolean variables? They seem unnecessary and potentially problematic with regards to edge cases like `desiredSize=''` and `desiredSize='0x0'` where they would return true. Instead we can just calculate desiredWidth and desiredHeight as
```js
const desiredWidth = desiredDims[0] ? Number.parseInt(desiredDims[0]) : 0;
if (Number.isNaN(desiredWidth) {
  throw ...
}
// same for height
...
if (desiredWidth && desiredHeight && originalRatio) {
  ...
}
```
This way, all falsy values will automatically be treated as desiredWidth=0 and by checking `if (desiredWidth)` these values would be ignored.

static/js/formatters-internal.js
```diff
@@ -311,164 +317,122 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
+  let formatOptions = ['fit=contain'];
```
nmanu1, line 369: this can be const since we don't directly update formatOptions again

static/js/formatters-internal.js
```diff
@@ -311,164 +317,122 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
     if (Number.isNaN(desiredHeight)) {
       throw new Error("Invalid height specified");
     }
-  } else {
-    desiredHeight = atLeastAsLarge ? 1 : Number.parseInt(fullSizeDims[1]);
   }
 
-  const urlWithDesiredDims = urlBeforeDimensions + desiredWidth + 'x' + desiredHeight + extension;
+  const originalRatio =
+    (!!fullSizeWidth && !!fullSizeHeight) ? (fullSizeWidth / fullSizeHeight) : undefined;
```
nmanu1, line 392: do we need the double negations here? they already get treated as boolean values because of the && operator, so having the !! just makes it a little harder to read imo

static/js/formatters-internal.js
```diff
@@ -311,164 +317,122 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
 
-  return atLeastAsLarge
-    ? _replaceUrlHost(urlWithDesiredDims, 'dynl.mktgcdn.com')
-    : _replaceUrlHost(urlWithDesiredDims, 'dynm.mktgcdn.com');
-}
+  // both dimensions are provided
+  if (hasDesiredWidth && hasDesiredHeight && originalRatio) {
+    const width = atLeastAsLarge
+      ? Math.max(desiredWidth, Math.round(desiredHeight * originalRatio))
```
nmanu1, line 397: nit: can we condense this logic?
```js
// please feel free to change this naming. didn't spend much time on it
const mathFunc = atLeastAsLarge ? Math.max : Math.min;
const width = mathFunc(desiredWidth, Math.round(...));
const height = mathFunc(desiredHeight, Math.round(...));
```
this way, we can avoid some of the long/repeated ternaries

static/js/formatters-internal.js
```diff
@@ -311,164 +317,122 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
-function _getEuImageDynamicUrl(image, desiredSize, atLeastAsLarge) {
-  let fullSizeWidth, fullSizeHeight;
-  if (image.width) {
-    fullSizeWidth = image.width;
-  }
-  if (image.height) {
-    fullSizeHeight = image.height;
-  }
+    formatOptions.push(`width=${width}`);
+    formatOptions.push(`height=${height}`);
```
nmanu1, line 404: nit: we can do these together, `formatOptions.push('width=...', 'height=...');`

static/js/formatters-internal.js
```diff
@@ -311,164 +317,122 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
+ *   '/p-sandbox/mFsjqWGQEOMQGNoNIcnq61JtdSGiCs',
+ *   '/p-sandbox/mFsjqWGQEOMQGNoNIcnq61JtdSGiCs',
+ *   '/p-sandbox/mFsjqWGQEOMQGNoNIcnq61JtdSGiCs/1.0000' respectively)
  */
-function _replaceUrlHost(url, host) {
-  const splitUrl = url.split('://');
-  const urlAfterHost = splitUrl[1].slice(splitUrl[1].indexOf('/'));
-  return splitUrl[0] + '://'  + host + urlAfterHost;
+function _removePhotoImageUrlExtension(imageUrlPath) {
+  return imageUrlPath.replace(/(\/[0-9]+x[0-9]+\.[a-z]+(\/)*)|(\/)$/, '');
```
nmanu1, line 435: should this regex be `/(\/[0-9]+x[0-9]+\.[a-z]+(\/)?)|(\/)$/`? (the difference is `(\/)?` instead of `(\/)*` after the dimension match)

tests/static/js/formatters-internal/image.js
```diff
@@ -1,106 +1,197 @@
...
-  const euUrl = 'https://dyn.eu.mktgcdn.com/f/0/FOO.jpg';
-  const usImg = {url: usUrl};
-  const euImg = {
-    url: euUrl,
-    width: 1024,
-    height: 768,
-    sourceUrl: 'https://a.mktgcdn.com/p/FOO/1024x768.jpg',
+  const photoUrl = 'https://dynm.mktgcdn.com/p/FOO/2000x1000.jpg/';
+  const photoWithPaddingUrl = 'https://dynm.mktgcdn.com/p/FOO/1.0000/2000x1000.jpg/';
+  const oldFileUrl = 'https://a.mktgcdn.com/f/0/FOO.jpg';
```
nmanu1, line 6: I thought only EU URLs were in the file format with the `/f/`, not US?

tests/static/js/formatters-internal/image.js
```diff
@@ -1,106 +1,197 @@
...
+  const newFileUrl = 'https://a.mktgcdn.com/f/FOO.jpg';
+  const euFileUrl = 'https://a.eu.mktgcdn.com/f/0/FOO.jpg';
+
+  const photoImg = {
+    url: photoUrl,
+    width: 2000,
+    height: 1000,
```
nmanu1, line 13: nit: can we pull the width and height properties out into a separate object and spread them for each type of img so they're not repeated?

tests/static/js/formatters-internal/image.js
```diff
@@ -1,106 +1,197 @@
...
 
-    it('when choosing the biggest image under threshold, use dimensions from the sourceUrl if width/height does not exist', () => {
-      const euImageUrl = Formatters.image(
-        { url: euUrl,
-          width: 1024,
-          sourceUrl: 'https://a.mktgcdn.com/p/FOO/516x384.jpg'
-        }, 'x', false).url;
-      expect(euImageUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/width=516,height=384,fit=contain');
-    });
+    it('do not transform image regardless of atLeastAsLarge\'s', () => {
```
nmanu1, line 173: can you update this test name to indicate that non-parseable or NaN dimensions are passed in? and can you also add tests for `desiredSize=''` and if `0` is passed for one or both desired dims?

static/js/formatters-internal.js
```diff
@@ -283,22 +283,28 @@ export function joinList(list, separator) {
...
+ *     preserving the original image ratio (e.g. if original image is 100x100, returns an image of
+ *     100x100).
  *
- * If atLeastAsLarge = false, the dynamic thumbnailer url will give the largest image that is
- * smaller than the desired size in both dimensions.
+ * If only one dimension is provided in desiredSize (e.g. "200x", which is also the default):
+ *   - returns an image that matches this dimension while preserving ratio of the original image
+ *     (e.g. if original image is 100x100, returned image would be 200x200).
+ *
+ * If "" is provided as the desiredSize, returns an image in the original size.
```
nmanu1, line 301: can this be `If "", "x", "0x0", or a string with unexpected format is provided as the desiredSize ....`? We may want to call out that 0 values are disregarded

tests/static/js/formatters-internal/image.js
EmilyZhang777: It doesn't hurt to have extra test coverage, going to keep. Ack

tests/static/js/formatters-internal/image.js
EmilyZhang777: Done

tests/static/js/formatters-internal/image.js
EmilyZhang777: Done. desiredSize='' is forbidden from the caller

static/js/formatters-internal.js
```diff
@@ -311,164 +319,113 @@ export function image(simpleOrComplexImage = {}, desiredSize = '200x', atLeastAs
...
+ * @param {number?} fullSizeHeight The full size height of the original image if exists
+ * @returns {string} A string representing the format options
+ *                   (e.g. 'height=200,width=100,fit=contain')
  */
-function _getEuImageDynamicUrl(image, desiredSize, atLeastAsLarge) {
-  let fullSizeWidth, fullSizeHeight;
-  if (image.width) {
-    fullSizeWidth = image.width;
+function _getImageFormatOptions(desiredSize, atLeastAsLarge, fullSizeWidth, fullSizeHeight) {
+  let desiredDims = desiredSize.split('x');
```
nmanu1, line 368: nit: const

tests/static/js/formatters-internal/image.js
```diff
@@ -93,15 +71,15 @@ describe('image formatter', () => {
...
 
     it('if can\'t get original ratio, use desired dimensions', () => {
-      const photoImgUrl = Formatters.image({url: photoUrl}, '1000x200', false).url;
+      const photoImgUrl = Formatters.image({url: photoUrl, width: 100}, '1000x200', false).url;
       expect(photoImgUrl).toEqual('https://dyn.mktgcdn.com/p/FOO/fit=contain,width=1000,height=200');
-      const photoWithPaddingImgUrl = Formatters.image({url: photoWithPaddingUrl}, '1000x200', false).url;
+      const photoWithPaddingImgUrl = Formatters.image({url: photoWithPaddingUrl, width: 0}, '1000x200', false).url;
       expect(photoWithPaddingImgUrl).toEqual('https://dyn.mktgcdn.com/p/FOO/1.0000/fit=contain,width=1000,height=200');
-      const oldFileImgUrl = Formatters.image({url: oldFileUrl}, '200x500', false).url;
+      const oldFileImgUrl = Formatters.image({url: oldFileUrl, height: 100}, '200x500', false).url;
```
nmanu1, line 78: nit: can we update one of these to test both height and width being 0?

tests/static/js/formatters-internal/image.js
```diff
@@ -170,14 +148,14 @@ describe('image formatter', () => {
       expect(euFileImgUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/fit=contain,width=200');
     });
 
-    it('do not transform image regardless of atLeastAsLarge\'s', () => {
+    it('do not transform image regardless of atLeastAsLarge', () => {
       let photoImgUrl = Formatters.image(photoImg, 'x', false).url;
       expect(photoImgUrl).toEqual('https://dyn.mktgcdn.com/p/FOO/fit=contain');
-      let photoWithPaddingImgUrl = Formatters.image(photoWithPaddingImg, 'x', false).url;
+      let photoWithPaddingImgUrl = Formatters.image(photoWithPaddingImg, '0x0', false).url;
```
nmanu1, line 154: can we also update one of these tests to check `''`?

static/js/formatters-internal.js
```diff
@@ -283,22 +283,30 @@ export function joinList(list, separator) {
...
+ *     preserving the aspect ratio. In other words, both dimensions will either match or be smaller
+ *     than the desired dimensions (e.g. if original image is 100x100, returns an image of 100x100).
  *
- * If atLeastAsLarge = false, the dynamic thumbnailer url will give the largest image that is
- * smaller than the desired size in both dimensions.
+ * If only one dimension is provided in desiredSize (e.g. "200x", which is also the default):
+ *   - returns an image that matches this dimension while preserving ratio of the original image
+ *     (e.g. if original image is 100x100, returned image would be 200x200).
+ *
+ * If "", "x", "0x0", or a string with unexpected format is provided as the desiredSize, returns an
```
nmanu1, line 301: can you also mention that `100x0` would be treated as `100x` because 0 values are disregarded?

static/js/formatters-internal.js
nmanu1: !! actually doesn't change how 0 values are handled, which is why I think it would be clearer without the double negation. but also don't feel super strongly about it

tests/static/js/formatters-internal/image.js
```diff
@@ -1,106 +1,181 @@
...
+      expect(euFileImgUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/fit=contain');
     });
+  });
 
-    it('when choosing the smallest image over threshold, omit width/height if can\'t parse it from the image object', () => {
-      const euImageUrl = Formatters.image({url: euUrl}, 'x', true).url;
-      expect(euImageUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/fit=cover');
+  describe('when desiredSize is not parseable', () => {
+    it('throws an error', () => {
+      expect(() => {Formatters.image(photoImg, 'ax')}).toThrowError();
```
nmanu1, line 178: can you also add a case for the height throwing an error (e.g. `100x-1`?

static/js/formatters-internal.js
EmilyZhang777: done

tests/static/js/formatters-internal/image.js
nmanu1: actually, just saw your previous response to this. what do you mean by it being "forbidden from the caller"?

static/js/formatters-internal.js
EmilyZhang777: done

static/js/formatters-internal.js
EmilyZhang777: Ack
https://stackoverflow.com/questions/29312123/how-does-the-double-exclamation-work-in-javascript

tests/static/js/formatters-internal/image.js
nmanu1: I think we're missing a case though. The current euFileUrl is actually the old format, and we also need to test for the new EU file url format without the account id, right?

tests/static/js/formatters-internal/image.js
EmilyZhang777: done

tests/static/js/formatters-internal/image.js
EmilyZhang777: Like I commented on the other thread, '' would throw an error. Added a test to cover that

tests/static/js/formatters-internal/image.js
EmilyZhang777: done

tests/static/js/formatters-internal/image.js
EmilyZhang777: I don't think it's worth/possible to add a test case for every single possible case. The "with account id" case is already covered by newFileUrl. euFileUrl is mainly to test for the eu domain. Ack

nmanu1: I think that's the opposite. Based on [this](https://yext.slack.com/archives/C0DA5HJCT/p1713445795909829) thread, the new behavior is **without** the account id, specifically for EU file URLs, which is why it's important to test this case. Going forward, the two types of URLs we'll get is US photo URLs and EU file URLs without account id, so I think we should cover those cases, even if we don't cover others like file URLs with account id in US domain

static/js/formatters-internal.js
nmanu1: yeah, that does a good job explaining how 0 is a falsy value in JS. and [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND) explains how the && evaluates values to truthy/falsy when making comparisons

tests/static/js/formatters-internal/image.js
```diff
@@ -1,106 +1,189 @@
...
-      expect(euImageUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/width=516,height=384,fit=contain');
+    it('throws an error', () => {
+      expect(() => {Formatters.image(photoImg, 'xa')}).toThrowError();
     });
 
-    it('when choosing the smallest image over threshold, omit width/height if can\'t parse it from the image object', () => {
-      const euImageUrl = Formatters.image({url: euUrl}, 'x', true).url;
-      expect(euImageUrl).toEqual('https://dyn.eu.mktgcdn.com/f/0/FOO.jpg/fit=cover');
+    it('throws an error', () => {
+      expect(() => {Formatters.image(photoImg, '')}).toThrowError();
```
nmanu1, line 186: can these tests be more specific? it would help if the name explained what error/why it's being thrown. and you should be able to check the error message being thrown with `toThrowError('Invalid desired size')`

EmilyZhang777: done

tests/static/js/formatters-internal/image.js
EmilyZhang777: done

nmanu1 (approved):

PR #1187, opened 2024-05-09
Version 1.34.0

jesuyedavid:
### Changes
- Updated the image formatter to return a fully dynamically-sized image instead of the closest thumbnail out of a fixed set (#1184)


nmanu1 (commented):
can you put the release notes in the pr description?

nmanu1 (commented):
Can you include the PR number in the change description?

Also, the release notes are intended for customers and should be phrased accordingly. For example, a customer wouldn't understand what "the image url change" is and would probably not care much about the format of the URL. Something like this would be more relevant for them:

`- Updated the image formatter to return a fully dynamically-sized image instead of the closest thumbnail out of a fixed set (#1184)`

nmanu1 (approved):

PR #1193, opened 2024-06-27
Upgrade testcafe to remove dependency vuln

Fondryext:
Upgrade testcafe from 1.18.6 to 3.6.1. In the process, this removes the dependency on jsonwebtoken, which had a vuln in the version we were using. 
https://nvd.nist.gov/vuln/detail/CVE-2022-23539#vulnCurrentDescriptionTitle 
J=VULN-39102
TEST=auto

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,19 +36,26 @@
         "puppeteer": "^13.7.0",
         "serve": "^13.0.4",
         "simple-git": "^2.24.0",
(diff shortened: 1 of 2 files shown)


Fondryext:
Note: WCAG failures occur on master as well.

vijay267 (approved):
LGTM

PR #1200, opened 2024-08-12
Add a document vertical standard card

Fondryext:
To help with development of the document vertical, this change adds a default document vertical card and styling for users to build off of.
J=WAT-4183
TEST=manual

I made these changes in a test site, and confirmed that dcoument results displayed correctly

### DIFF
--- a/cards/document-vertical-standard/component.js
+++ b/cards/document-vertical-standard/component.js
@@ -1,0 +1,55 @@
+{{> cards/card_component componentName='document-vertical-standard' }}
+
+class document_vertical_standardCardComponent extends BaseCard['document-vertical-standard'] {
+    constructor(config = {}, systemConfig = {}) {
+        super(config, systemConfig);
+    }
+
+    /**
+     * This returns an object that will be called `card`
+     * in the template. Put all mapping logic here.
+     *
+     * @param profile profile of the entity in the card
+     */
+    dataForRender(profile) {
+        const linkTarget = AnswersExperience.runtimeConfig.get('linkTarget') || '_top';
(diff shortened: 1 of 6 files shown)


cards/document-vertical-standard/component.js
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param profile profile of the entity in the card
+     */
+    dataForRender(profile) {
+        const linkTarget = AnswersExperience.runtimeConfig.get('linkTarget') || '_top';
+
+        return {
+            title: profile.name, // The header text of the card
+            // subtitle: '', // The sub-header text of the card
+            url: profile.landingPageUrl, // If the card title is a clickable link, set URL here
```
vijay267, line 20: Would these ever be populated? Plus is there a field like "c_department" always present on a profile? If they're not present what happens?

vijay267 (commented):
Generally looks fine, but had some questions about fields that I wouldn't think would be present for a document vertical?

cards/document-vertical-standard/component.js
Fondryext: subtitle was a miss. landingPageURL is a universal field, so I left it in since I could see clients wanting to use it. linkTarget is configurable and all cards reference it, and should be populated. That said, if we don't want any url fields maybe linkTarget isn't getting us much done here.
If a value isn't present then in the handlebars file we don't render it.
I'm going to remove subtitle now (as a populated field, but keep it commented out), but let me know about the other two.

vijay267 (approved):

PR #1215, opened 2025-02-12
GDA Card Generator

k-gerner:
Add card generation script for GDA. This should work in the same way as the existing direct answer card generation script.

J=[WAT-4596](https://yexttest.atlassian.net/browse/WAT-4596?atlOrigin=eyJpIjoiNTU5MzYxMWJjYmY4NGE2NTg5NDNjNmQxZDk0NzA4NjEiLCJwIjoiaiJ9)
TEST=manual
Ran `npx jambo init --theme answers-hitchhiker-theme` then `npx jambo upgrade --branch gda-card-generator`

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,3 +78,13 @@ npx jambo directanswercard --name custom-directanswer --templateCardFolder direc
 ```
 
 See `jambo directanswercard --help` for more info.
+
+### Generative Direct Answer Card
+Creates a new, custom generative direct answer (GDA) card.
+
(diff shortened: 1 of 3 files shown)


Fondryext (commented):
Looks good but I think the chrome test failure is real:

UserError: The folder ../../answers-hitchhiker-theme/generativedirectanswercards/generative-custom does not exist at the root or in the theme.

Fondryext (approved):

PR #1220, opened 2025-03-03
Add GDA to all vertical templates

k-gerner:
Previously GDA was only on the standard vertical/universal templates. It is now present on all vertical templates.

### DIFF
--- a/test-site/pages-patches/faqs.html.hbs.patch
+++ b/test-site/pages-patches/faqs.html.hbs.patch
@@ -1,6 +1,6 @@
 --- faqs.html.hbs	2022-04-05 11:53:14.000000000 -0400
 +++ faqs.updated.html.hbs	2022-04-05 11:54:18.000000000 -0400
-@@ -15,7 +15,7 @@
+@@ -17,7 +17,7 @@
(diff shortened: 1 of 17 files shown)


test-site/pages-patches/faqs.html.hbs.patch
```diff
@@ -1,6 +1,6 @@
 --- faqs.html.hbs	2022-04-05 11:53:14.000000000 -0400
 +++ faqs.updated.html.hbs	2022-04-05 11:54:18.000000000 -0400
```
vijay267, line 2: Sorry what's the purpose of these updated .hbs files for the test site?

vijay267: Also does this file also need GDA?

k-gerner: The `npm run setup-test-site` command will fail because it attempts to apply the patches in these files. Because the files that the patches are applying to have changed due to the template changes, we need to upgrade the patches too. 

This patch file doesn't need the GDA specifically because it's outside the ranges of the lines that we're changing. That's why I changed the line number ranges still

Fondryext (approved):
LGTM, assuming Deepak is ok with the other conversation

PR #1222, opened 2025-03-05
Use sourceUrlFields from config to set citation links

k-gerner:
Allow custom fields for GDA citation links

The entity field used to populate the `href` (click through link) on GDA citations is now configurable. 
This is configurable in the `sourceUrlFields` property of the GDA component, on a per-vertical basis. 
This value can be a single string, or a list of strings. If multiple are listed, then they will be tried in order, with the secondary fields being used if the primary field is not valid. 

J=[WAT-4680](https://yexttest.atlassian.net/browse/WAT-4680?atlOrigin=eyJpIjoiMmE2Y2M2MzU5MTY1NDcxZThkYWIyY2I1YzVmMjdkMzMiLCJwIjoiaiJ9)

Related change in `answers-search-ui`: https://github.com/yext/answers-search-ui/pull/1923

### DIFF
--- a/test-site/config/index.json
+++ b/test-site/config/index.json
@@ -7,7 +7,7 @@
     "search": {
       "defaultInitialSearch": "virginia" // Enter a default search term
     }
-  },  
+  },
   "componentSettings": {
     /**
     "QASubmission": {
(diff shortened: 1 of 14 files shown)


test-site/config/index.json
```diff
@@ -16,7 +16,8 @@
     },
     **/
     "GenerativeDirectAnswer": {
-      "cardType": "generative-standard" // The card type to use for the generated direct answer
+      "cardType": "generative-standard", // The card type to use for the generated direct answer
+      "sourceUrlFields": ["website", "featuredMessage.url"] // The fields to use as the redirect URL for the sources
```
Fondryext, line 20: I don't feel strongly + I know this is test data, but it might be nice to have this be in line with the other template files

Fondryext (approved):

PR #1229, opened 2025-07-16
chore: change phone number library to awesome-phonenumber

k-gerner:
Needed this change due to an APV. 
Also updated the `es` targets for the build step so that we can build the test site. This shouldn't affect any browsers except maybe IE11. 

J=[APV-796](https://yexttest.atlassian.net/browse/APV-796?atlOrigin=eyJpIjoiZmE5NGMzZDg2YWY1NDY2YWJlZWFhNjI3YTMxMzIzOWIiLCJwIjoiaiJ9)
TEST=manual
Created entity in UK, spun up test site, and saw phone number get formatted correctly

### DIFF
--- a/static/js/formatters-internal.js
+++ b/static/js/formatters-internal.js
@@ -1,4 +1,4 @@
-import { parsePhoneNumberFromString } from 'libphonenumber-js'
+import { parsePhoneNumber } from 'awesome-phonenumber';
 import { components__address__i18n__addressForCountry } from './address-i18n.js'
 import { getDistanceUnit } from './units-i18n';
 import OpenStatusMessageFactory from './hours/open-status/messagefactory.js';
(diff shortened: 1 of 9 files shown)


static/js/formatters-internal.js
```diff
@@ -39,11 +39,12 @@ export function phoneDisplay(profile, key = 'mainPhone') {
 }
 
 export function nationalizedPhoneDisplay(profile, key = 'mainPhone') {
-  if (!profile[key]) {
+  if (!profile[key] || !profile?.address?.countryCode) {
```
vijay267, line 42: So do we expect the country code to always be provided? If so, this is fine, but otherwise maybe it makes sense to have a fallback country code of US.

vijay267 (approved):

PR #1246, opened 2026-02-09
fix: percy workflow actions

k-gerner:
Fixes the Percy Snapshots checks for this repository.

Various fixes including:
- update workflow version to point to `main` branch
- moving build command into a `.sh` file, so that the workflow can correctly parse the instructions (previously the workflow was only parsing the first command, not the command after the `&&`)
- make sure `esbuild` modules are accessible for CI build
- remove GDA from config used for snapshots, so that it doesn't mark a visual regression


Fondryext:
Any idea what's up with that last percy check?

Fondryext (approved):

PR #1249, opened 2026-04-30
v2.0.0

Fondryext:
Adds support for SDK v2.0, which itself adds support for the Events API. This change is not too large, but does add a new global config field (eventsApiKey), and removes a deprectaed one (conversionTrackingEnabled). It also updates the old event names to the new ones in our premade code - mostly TITLE_CLICK -> TITLE.
J=WAT-4646
TEST=manual

Spun up the test site and saw new events send as expected.


k-gerner (dismissed):
LGTM. I think we will need to work with documentation team / David to get the HH docs updated though, so that it's clear how to enable/disable analytics with the new flow

vijay267 (dismissed):

anguyen-yext2 (approved):

